PR #33, opened 2020-10-23
Strawman: How would projects map to working groups?

loewenstein:
This is meant as a strawman to get something out. It definitely needs input from the community and projects, if the grouping would make sense for them.

### DIFF
--- a/working-groups/WORKING-GROUPS.md
+++ b/working-groups/WORKING-GROUPS.md
@@ -27,7 +27,7 @@ other system (Apple Calendar or Outlook, for example),
 
 # Working Groups
 
-The current working groups are:
+Initial straw proposal for the working groups, to be modified and discussed by the first TOC:
 
 - [API](#api)
 - [Application Runtime](#application-runtime)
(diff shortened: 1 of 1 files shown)


gcapizzi:
What do people think about having different WGs for CF-on-VMs vs CF-on-K8s?

I'm on the Eirini team, and while we have context con both worlds for historical reasons (we are 100% made of people who worked on Garden and we still maintain Garden to this day), it doesn't look like the Diego/Garden stack could share a Roadmap with Eirini.

chipchilders:
> What do people think about having different WGs for CF-on-VMs vs CF-on-K8s?
> 
> I'm on the Eirini team, and while we have context con both worlds for historical reasons (we are 100% made of people who worked on Garden and we still maintain Garden to this day), it doesn't look like the Diego/Garden stack could share a Roadmap with Eirini.

This comment makes a ton of sense. At this point, this PR is sitting here as a starting point for the TOC to use *once the TOC exists*. All of the current projects will need to be appropriately structured. Keep comments / ideas like this coming! :)

emalm:
Adding a link to the [public Miro board](https://miro.com/app/board/o9J_lJ8wLOE=/) where we played around with mapping the existing projects to new working groups, plus a snapshot of the frame for posterity: [[Public] CFF Working Groups scratch space - 2021-04-16.pdf](https://github.com/cloudfoundry/community/files/6326389/Public.CFF.Working.Groups.scratch.space.-.2021-04-16.pdf).

emalm:
Added the password `workingonit` to the public [Miro board](workingonit) as a screen-door level of security (FYI @loewenstein @bkrannich @norman-abramovitz @jenspinney @chipchilders).

loewenstein:
obsolete

Issue #46, opened 2021-01-08
Remove TOC estimate of hours of committment


chipchilders:
We simply don't know how much work it is. Should remove that sentence.

chipchilders:
Done

PR #48, opened 2021-01-08
Update TOC.md

troytop:
As per #46, we don't know yet how much of a time commitment this will be yet.


chipchilders (approved):
Agreed to in meeting

PR #60, opened 2021-01-29
Reflow TOC charter section

emalm:
Content now flows as prose, with specific activities still structured as
bullet points.

Fixes #56 by linking to the community principles document.

PR #61, opened 2021-01-29
Remove language referring to "subprojects"

emalm:
The notion of a "subproject" is not clearly defined in the governance
rules, and based on the existing minimal usage in the ROLES document
seems to correspond to a Working Group. This change removes language
referring to subprojects in favor of consolidating on only Working
Groups as the governance structure below the TOC.

Sadly, this change will conflict with the excellent PR #60.

### DIFF
--- a/TOC.md
+++ b/TOC.md
@@ -35,10 +35,6 @@ product and design decisions.
   - Establish and maintain the overall technical governance guidelines for the
     project.
 
-  - Decide which sub-projects are part of the Cloud Foundry project, including
-    accepting new sub-projects and pruning existing sub-projects to maintain
-    community focus
-
   - Ensure the team adheres to our
     [code of conduct](./CONTRIBUTING.md#code-of-conduct) and respects our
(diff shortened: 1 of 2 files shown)


TOC.md
```diff
@@ -35,10 +35,6 @@ product and design decisions.
   - Establish and maintain the overall technical governance guidelines for the
     project.
 
-  - Decide which sub-projects are part of the Cloud Foundry project, including
-    accepting new sub-projects and pruning existing sub-projects to maintain
-    community focus
-
```
loewenstein, line 41: I am all for removing undefined and duplicative terms. The only question I would raise is if the whole thing belongs into the technical or community governance section.
The former already has (and had in addition of what you propose to remove) the following:
>Approve the creation and dissolution of working groups and approve leadership changes of working groups.

loewenstein (approved):
I am fine with the removal. Just wanted to take the chance that the topic of WG creation/removal is present twice to raise the question of where it belongs.

PR #67, opened 2021-01-29
Ensure time for community input when no consensus

### LINKED ISSUE #55: If a TOC vote is required to reach a decision, this should usually be taken offline to allow participation and time for seeking community input
Didn't we plan to include something like this. in https://github.com/cloudfoundry/community/blob/main/TOC.md#committee-decision-making?

chipchilders:
Proposed language to try to ensure that the TOC gives decisions that are not easy consensus decisions enough time to gather community input.

Fixes #55


loewenstein (approved):
LGTM

PR #70, opened 2021-02-03
Remove repetition of the section title

loewenstein:


emalm:
I'd be in support of this smoother flow to the wording, but I think Chip intended that definition to be called out there to provide a clear referent for the "Employer Diversity Rule" references later. There's probably a way to rework both of those references to be less formal, though.

chipchilders:
@emalm is correct. This was a "formal definition" of the rule, which then is used elsewhere in the doc.

chipchilders:
Resolved via #81

PR #71, opened 2021-02-03
Use the term Board of DIrectors instead of Governing Board

loewenstein:


chipchilders:
As of 1 Jan 2021, the correct name for the board is "Governing Board".

PR #81, opened 2021-02-04
Revise TOC employer diversity without formal rule

emalm:
This is an attempt to express the employer-diversity limits without reference to a named rule.


loewenstein (approved):
lgtm

chipchilders (approved):
LGTM - nice flow @emalm

PR #84, opened 2021-02-04
Proposed change to who decides on voter eligibility

### LINKED ISSUE #74: Remove use of the term 'steering committee'
https://github.com/cloudfoundry/community/blob/main/TOC.md#voter-eligibility mentions this as the resolving body for voter eligibility exceptions, but we do not have a steering committee.

Should this be handled by the BoD or the TOC? I don't see how the BoD would, but giving the TOC too much influence over who is eligible to vote them might be seen as a conflict of interest. Thought?

chipchilders:
Fixes #74 but is open for discussion obviously :)

### DIFF
--- a/TOC.md
+++ b/TOC.md
@@ -177,7 +177,7 @@ shows only GitHub based contributions and does not capture all the contributions
 we value. **We expect this metric not to capture everyone who should be eligible
 to vote.** If a community member has had significant contributions over the past
 year but is not captured in the lfanalytics.io dashboard, they will be able
-to submit an exception form to the steering committee who will then review and
+to submit an exception form to the current TOC who will then review and
 determine whether this member should be marked as an exception.
 
 All eligible voters will be captured at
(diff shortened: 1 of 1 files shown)


TOC.md
```diff
@@ -177,7 +177,7 @@ shows only GitHub based contributions and does not capture all the contributions
 we value. **We expect this metric not to capture everyone who should be eligible
 to vote.** If a community member has had significant contributions over the past
 year but is not captured in the lfanalytics.io dashboard, they will be able
-to submit an exception form to the steering committee who will then review and
+to submit an exception form to the current TOC who will then review and
```
loewenstein, line 180: Should we explicitly allow an escalation to the Governing Board?

chipchilders:
I like that idea.

TOC.md
chipchilders: Added

PR #93, opened 2021-03-12
TOC: clean up election document locations

emalm:


chipchilders (approved):
LGTM - This is a simple change that I'm going to merge in today, so that I don't have a conflict when I move things around to support GB approvals.

PR #99, opened 2021-04-30
Add member nominations for first TOC election

chipchilders:
In discussion about our first election, it was identified that it might be useful for there to be an additional way for candidates to be added to the pool. One brainstorm was to allow all CFF member organizations to have the opportunity to nominate candidates. Again, first election only.

### DIFF
--- a/toc/CHANGEPLAN.md
+++ b/toc/CHANGEPLAN.md
@@ -32,7 +32,9 @@ eligibility to be nominated for the TOC during the first election will
 be defined as:
 
 * Community members who are PMC Leads, Project Leads or committers within 
-  any PMC or project.
+  any PMC or project. Additionally, each member organization of the CFF
+  will have an opportunity, if they desire, to nominate a candidate for the
+  first TOC election.
   
 ### Initial TOC Election Voter Eligibility
 
--- a/toc/CHANGEPLAN.md
+++ b/toc/CHANGEPLAN.md
@@ -32,8 +32,8 @@ eligibility to be nominated for the TOC during the first election will
 be defined as:
 
 * Community members who are PMC Leads, Project Leads or committers within 
-  any PMC or project. Additionally, each member organization of the CFF
-  will have an opportunity, if they desire, to nominate a candidate for the
+  any PMC or project. Each member organization of the CFF can also nominate 
+  one additional person not meeting these criteria as a candidate for the 
   first TOC election.
   
 ### Initial TOC Election Voter Eligibility


emalm:
Thanks, Chip, generally looks good to me, based on the group discussion last Friday.

I have a couple of clarifying questions:
- Would this nomination provision extend to any CFF member company, regardless of tier, or would there be a cut-off that excludes some tiers (say, Associate)?
- Should this provision state that each group of Related Companies (or whatever wording matches the CFF charter) gets to nominate one additional candidate for this first TOC election?

chipchilders:
> Thanks, Chip, generally looks good to me, based on the group discussion last Friday.
> 
> I have a couple of clarifying questions:
> 
> * Would this nomination provision extend to any CFF member company, regardless of tier, or would there be a cut-off that excludes some tiers (say, Associate)?

IMO, yes. I don't see this as a negative, since we aren't trying to tie TOC (or any technical community) participation to paid memberships.

> * Should this provision state that each group of Related Companies (or whatever wording matches the CFF charter) gets to nominate one additional candidate for this first TOC election?

Similar to the above, I don't think we have an issue here in a practical sense. If you'd like, we can add that limit though.

chipchilders:
> Would the member organization of the CFF be able to propose anybody or they would need to follow the same rules as community members? I am not sure this adds anything since there is no restriction on community members being nominated unless member organizations do not need to follow the same rules. Three community members from the same member organization can nominate 3+ people.

The problem we are looking to address is that there are qualified individuals that have done technical work that is not recognized through a formalized role in the *current* technical structure. As it stands now, the plan (without this proposed change) requires that TOC candidates are "Community members who are PMC Leads, Project Leads or committers within any PMC or project." By allowing 1 nomination from any member that feels they have someone that should also qualify, we can improve the situation.

emalm:
> Would the member organization of the CFF be able to propose anybody

My interpretation of the intent is that each member organization would be able to nominate one additional person without restriction -- they wouldn't even have to be an employee of that organization.

@chipchilders @norman-abramovitz @loewenstein: How does the following wording sound to make that clearer?

> Each member organization of the CFF can also nominate one additional person not meeting these criteria as a candidate for the first TOC election.

emalm:
> > Should this provision state that each group of Related Companies (or whatever wording matches the CFF charter) gets to nominate one additional candidate for this first TOC election?
> 
> Similar to the above, I don't think we have an issue here in a practical sense. If you'd like, we can add that limit though.

Oh, right, this is just for the first TOC election. In that case, I think it's fine not to worry about member companies being related for these initial nominations.

norman-abramovitz:
Hi Chip,  That works for me about member organizations!  🙆

chipchilders:
> @chipchilders @norman-abramovitz @loewenstein: How does the following wording sound to make that clearer?
> 
> > Each member organization of the CFF can also nominate one additional person not meeting these criteria as a candidate for the first TOC election.

@emalm I like it. Anyone object to this language?

norman-abramovitz:
> > @chipchilders @norman-abramovitz @loewenstein: How does the following wording sound to make that clearer?
> > > Each member organization of the CFF can also nominate one additional person not meeting these criteria as a candidate for the first TOC election.
> 
> @emalm I like it. Anyone object to this language?
No objects!

PR #121, opened 2021-05-25
Remove some duplicate names

emalm:
These duplicates seemed apparent to me, but if we have a more canonical identifier of someone's identity that backs the name we may want to check against that to make sure before pruning these entries.


chipchilders (approved):
Appreciate the review and corrections. Thanks!

PR #142, opened 2021-07-19
Add stub template for App Runtime Platform WG

emalm:
This PR proposes a charter document for the App Runtime Platform working group.

PR #141, opened 2021-07-19
Add stub template for App Runtime Interfaces WG

emalm:
TODO before merge:
* Identify leads for WG
* Identify initial goals and scope for WG
* List main technical assets for WG

PR #143, opened 2021-07-19
Add stub template for Paketo WG

emalm:
TODO before merge:
* Identify leads for WG
* Identify initial goals and scope for WG
* List main technical assets for WG


dmikusa-pivotal:
> Identify leads for WG

I will volunteer to be one if you need volunteers. For those that don't know me, I've been working with CF since 2013 & wrote my first buildpack shortly after that. I'm working at VMware & I am on the team that maintains the Java buildpack & related assets, as well as all things Java under Paketo.

>Identify initial goals and scope for WG

IMO, the scope seems to split between CF buildpacks (v2) and Paketo buildpacks (v3). Paketo already has existing governance around process, changes & direction, but the CF buildpacks do not really. It would be nice to continue using the existing governance for Paketo & at the same time institute something similar for CF buildpacks, so that community members have input into the process and direction of those buildpacks.

My thoughts for initial goals:
- increase community feedback & participation with CF buildpacks
- increase the transparency of the direction of CF buildpacks (perhaps a roadmap)
- increase awareness of Paketo community & increase participation in it

>List main technical assets for WG

I don't know if this is an exhaustive list, but here are the buildpack related repos I could come up with:

- https://github.com/cloudfoundry/java-buildpack
- https://github.com/cloudfoundry/java-buildpack-memory-calculator
- https://github.com/cloudfoundry/java-buildpack-release
- https://github.com/cloudfoundry/java-buildpack-system-test
- https://github.com/cloudfoundry/java-buildpack-dependency-builder
- https://github.com/cloudfoundry/java-buildpack-container-customizer
- https://github.com/cloudfoundry/java-test-applications
- https://github.com/cloudfoundry/java-buildpack-support
- https://github.com/cloudfoundry/java-buildpack-security-provider
- https://github.com/cloudfoundry/java-buildpack-metric-writer
- https://github.com/cloudfoundry/java-buildpack-client-certificate-mapper
- https://github.com/cloudfoundry/java-buildpack-auto-reconfiguration
- https://github.com/cloudfoundry/ibm-websphere-liberty-buildpack
- https://github.com/cloudfoundry/nodejs-buildpack
- https://github.com/cloudfoundry/nodejs-buildpack-release
- https://github.com/cloudfoundry/php-buildpack
- https://github.com/cloudfoundry/php-buildpack-release
- https://github.com/cloudfoundry/ruby-buildpack
- https://github.com/cloudfoundry/ruby-buildpack-release
- https://github.com/cloudfoundry/python-buildpack
- https://github.com/cloudfoundry/python-buildpack-release
- https://github.com/cloudfoundry/go-buildpack
- https://github.com/cloudfoundry/go-buildpack-release
- https://github.com/cloudfoundry/binary-buildpack
- https://github.com/cloudfoundry/binary-buildpack-release
- https://github.com/cloudfoundry/nodejs-buildpack
- https://github.com/cloudfoundry/nodejs-buildpack-release
- https://github.com/cloudfoundry/dotnet-core-buildpack
- https://github.com/cloudfoundry/dotnet-core-buildpack-release
- https://github.com/cloudfoundry/hwc-buildpack
- https://github.com/cloudfoundry/hwc-buildpack-release
- https://github.com/cloudfoundry/nginx-buildpack
- https://github.com/cloudfoundry/nginx-buildpack-release
- https://github.com/cloudfoundry/staticfile-buildpack
- https://github.com/cloudfoundry/staticfile-buildpack-release
- https://github.com/cloudfoundry/r-buildpack
- https://github.com/cloudfoundry/r-buildpack-release
- https://github.com/cloudfoundry/apt-buildpack
- https://github.com/cloudfoundry/docs-buildpacks
- https://github.com/cloudfoundry/brats
- https://github.com/paketo-buildpacks/*
- https://github.com/paketo-community/*

loewenstein:
I talked to @ekcasey and @ryanmoran recently and they repeated their preference **not** to mix Paketo with the Cloud Foundry buildpacks.
The main reason being an interest to keep/make the Paketo community appealing for participants from outside of the Cloud Foundry project. We briefly touched that achieving this might in the longer term even require to move Paketo to e.g. the CNCF.

Further, they stated a preference to handle the Cloud Foundry buildpacks in the [Application Runtime Interfaces WG](https://github.com/cloudfoundry/community/pull/141) instead of a separate working group.
The main reason being an interest to more closely align and collaborate with the other parts of the Cloud Foundry project.

The main theme for Cloud Foundry buildpacks, besides maintenance, seemed to be integrating Paketo buildpacks with Cloud Foundry, deprecating and eventually removing the v2 buildpacks.
I can see that this theme
- feels extraneous for working group participants interested in Paketo outside of the Cloud Foundry ecosystem
- would benefit from a close alignment with the App Runtime working groups

@ekcasey and @ryanmoran assured the importance of and their caring for the Cloud Foundry buildpacks. As long as we agree - which we do - on the objective to converge Cloud Foundry buildpacks to v3 I don't think that we need to install gravity by enforcing a single working group.

I am not entirely sure yet, with which of the two App Runtime working groups the alignment should be stronger. Interfaces, because buildpacks are kind of an interface towards developers or Platform, because assets like [cloudfoundry/buildpackapplifecycle](https://github.com/cloudfoundry/buildpackapplifecycle) are tied to Diego and likely belong into the [App Runtime Platform WG](https://github.com/cloudfoundry/community/pull/142).

Either way, or if we decide for two buildpacks related working groups (with or without an umbrella working group like @chipchilders proposed), I advised that the close ties of current Paketo contributors also to Cloud Foundry should be reflected by a strong approver and/or leadership involvement in whatever working group will host the Cloud Foundry buildpacks.

@dmikusa-pivotal It would be interesting to get yet another opinion from the buildpacks' community.

@cloudfoundry/toc Thoughts?

dmikusa-pivotal:
>The main reason being an interest to keep/make the Paketo community appealing for participants from outside of the Cloud Foundry project.

+1 This is important to the success of Paketo.

 >Further, they stated a preference to handle the Cloud Foundry buildpacks in the Application Runtime Interfaces WG instead of a separate working group.

That seems reasonable. The #1 thing I would like to see is first-class support for v3 buildpacks in CloudFoundry. A lot of that work would happen outside of buildpacks, in CAPI & Diego. If it helps to include buildpacks in a different WG to drive this forward then that is fine by me. I could see this WG having a focus on how buildpacks are integrated into the platform, and using the existing Paketo WG for discussions on v3 buildpacks.

>The main theme for Cloud Foundry buildpacks, besides maintenance, seemed to be integrating Paketo buildpacks with Cloud Foundry, deprecating and eventually removing the v2 buildpacks.

This is one area where a buildpacks WG might be helpful (although I guess any WG where folks could discuss this would be helpful). I don't feel like we get a lot of feedback from the community on the direction for buildpacks. It logically makes sense to deprecate v2 and move to v3, but that will require people to migrate, the transition is not always a seamless one between v2 and v3. It would be nice to have feedback from users on what they want, what sort of timelines they need to migrate, how we can help ease the migrations, etc... We would want the transition to be as smooth as possible and that's going to require feedback to understand the concerns of users.

>I am not entirely sure yet, with which of the two App Runtime working groups the alignment should be stronger. Interfaces, because buildpacks are kind of an interface towards developers or Platform, because assets like cloudfoundry/buildpackapplifecycle are tied to Diego and likely belong into the App Runtime Platform WG.

Same. I think either would be fine. If I had to pick, I'd probably go with Interfaces cause I think that would align more with CAPI & I think that's where most of the work for getting to v3 buildpacks exists.

LeePorte:
> > The main theme for Cloud Foundry buildpacks, besides maintenance, seemed to be integrating Paketo buildpacks with Cloud Foundry, deprecating and eventually removing the v2 buildpacks.

> This is one area where a buildpacks WG might be helpful (although I guess any WG where folks could discuss this would be helpful). I don't feel like we get a lot of feedback from the community on the direction for buildpacks. It logically makes sense to deprecate v2 and move to v3, but that will require people to migrate, the transition is not always a seamless one between v2 and v3. It would be nice to have feedback from users on what they want, what sort of timelines they need to migrate, how we can help ease the migrations, etc... We would want the transition to be as smooth as possible and that's going to require feedback to understand the concerns of users.

This is my concern with having different leads for V2 and V3 buildpacks. With the suggestion of placing V2 buildpacks in Application Runtime WG are V3 buildpacks going to remain a second class citizen? 

In order to help us get to the position of having V3 as a first-class support within Cloudfoundry, could a route be to have V2 buildpacks as an umbrella project under V3 with the goal of deprecating and archiving V2 buildpacks? What do you think @ekcasey and @ryanmoran?

stephanme:
I share the same concerns as @LeePorte and would actually prefer to have one Buildpack WG that feels responsible for the buildpack roadmap, their integration into CF and a migration path from v2 to v3.

I also understand the concerns of binding Paketo too close to CF (technically and wrt. governance). Maybe we can state in the charter that the existing Paketo governance structure shall remain and that the goal of Paketo is to address communities outside of CF. Sort of dealing with it as if Paketo would be part of CNCF.

Buildpacks is a pretty big and important topic for CF. I would say it deserves an own WG. For sure the WGs have to work together, this is the same for Interfaces and Runtime.

A final thought: the App Runtime Interfaces WG is already pretty big in terms of topics and github projects ...

ryanmoran:
I think my initial thought was not to create a hierarchy, but to establish shared leads that could ensure we create alignment on the goals we want to support. Today, we already have a lot of that shared alignment naturally due to the fact that the same sets of engineers maintain both the CF (v2) and Paketo (v3) buildpacks. What we don't have is a forum for those leads to engage in discussion with the broader CF community.

What we had proposed to @loewenstein in our earlier conversation was that the CF and Paketo buildpacks need different forums, but that they should share leadership, at least initially. Having different forums would allow each effort to prioritize its own immediate needs, and having shared leadership would ensure we maintain reasonable alignment. As a straw-dog proposal, we had suggested that at least one of the leads from the Paketo group should also act as a lead in whatever WG the CF buildpacks landed in. I don't think we had any preference where that was, but it did make a lot of sense to align it with the App Runtime Interfaces WG.

If CF buildpacks did land in the App Runtime Interfaces WG, then I agree with @stephanme that this group appears to be quite large, but my understanding was that this was somewhat intentional in the original design. It strikes me that there appear to be 2 possible "tiers" of concerns in that WG already: 1) the primary interfaces for users (CLI, CAPI, Buildpacks), and 2) extensions or optional interfaces (Autoscaler, Java Tools, MultiApps, Notifications, Statos). If we did add CF Buildpacks to this WG, then does it make some sense to split it up a bit?

emalm:
I agree with the Buildpacks representatives that we should structure the organization under the TOC so that Paketo can be as independent as possible, and without being bound closely to the needs of the CF App Runtime project. The CFF established Paketo as an independent project last year, and we should respect that intent while the TOC provides them with technical resources and guidance, helps them establish project goals and structure, and holds them accountable to those goals and to the overall principles of community governance articulated in this repository. Paketo represents an excellent opportunity to increase the engagement of the broader cloud-native community with CFF-sponsored projects, which in turn provides a major opportunity to grow the CFF community itself, and I am concerned that tying it too closely to its CF App Runtime origins will hamper investments in it to encourage that growth. 

I also agree that we should include the CF Buildpacks in the scope of one of the App Runtime-focused working groups. If our primary objective for the App Runtime is to preserve compatibility for existing apps while consolidate technically on cloud-native (v3) buildpacks, then it seems most appropriate to place that consolidation effort in the App Runtime Interfaces working group, as it has the most responsibility to ensure continuity across developer-facing interfaces (of which buildpack behavior is one). Conversely, I see substantial risk that placing both domains in a separate Buildpacks working group will prioritize technical convergence over behavioral continuity, particularly if that working group prioritizes Paketo development and growth above the CF Buildpacks.

LeePorte:
Hey @ekcasey and @ryanmoran, Would love your thoughts on this updated PR.

toc/working-groups/paketo.md
```diff
@@ -0,0 +1,40 @@
+# Paketo: Working Group Charter
+
+## Mission
+
+Provides Paketo buildpacks for servers, languages, and frameworks popular with application developers.
+
+
+## Goals
+
+- Provides end users with stable, reliable and performant buildpacks in which to host their application.
```
ryanmoran, line 10:
```suggestion
- Provides end users with stable, reliable and performant buildpacks with which to build their application.
```

toc/working-groups/paketo.md
```diff
@@ -0,0 +1,40 @@
...
+- Provides end users with buildpacks that are compatable with multiple platforms.
```
ryanmoran, line 11:
```suggestion
- Provides end users with buildpacks that are compatable with multiple platforms (i.e. Cloud Foundry, Docker, Kubernetes).
```

toc/working-groups/paketo.md
```diff
@@ -0,0 +1,40 @@
...
+- Provides documentation to enable platforms to adopt Paketo buildpacks and maintain compatability with Paketo buildpacks.
```
ryanmoran, line 12:
```suggestion
- Provides documentation to enable platforms to adopt Paketo buildpacks and maintain compatibility with Paketo buildpacks.
```

ryanmoran (commented):
This looks good to me. I made some very minor copy edits for clarity.

toc/working-groups/paketo.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Mission
+
+Provides Paketo buildpacks for servers, languages, and frameworks popular with application developers.
+
+
+## Goals
+
+- Provides end users with stable, reliable and performant buildpacks with which to build their application.
+- Provides end users with buildpacks that are compatable with multiple platforms (i.e. Cloud Foundry, Docker, Kubernetes).
```
emalm, line 11:
```suggestion
- Provides end users with buildpacks that are compatible with multiple platforms (such as Cloud Foundry, Docker, and Kubernetes).
```

toc/working-groups/paketo.md
```diff
@@ -0,0 +1,40 @@
...
+## Proposed Membership
+
+- Technical Lead(s): TBD
+- Execution Lead(s): TBD
+- Approvers: TBD
+
+
+## Technical Assets
+
+Technical assets for the Paketo working group are contained in the following Github Organisations;
```
emalm, line 36:
```suggestion
Technical assets for the Paketo working group are contained in the following Github Organizations:
```

toc/working-groups/paketo.md
```diff
@@ -0,0 +1,40 @@
...
+- Provide the community with technical and end user documentation.
+- Collaborate with App Runtime Interfaces WG to enable Paketo buildpacks to have first-class support in Cloud Foundry.
+
+## Non-Goals
+
+- Maintaining V2 (classic) buildpacks
+
+## Working Group Governance Structure
+
+[Working group governance structure](https://github.com/paketo-buildpacks/community/blob/main/GOVERNANCE.md)
```
loewenstein, line 28: Should we check for potential convergence of this document with the [working groups](https://github.com/cloudfoundry/community/blob/main/toc/working-groups/WORKING-GROUPS.md) and [roles](https://github.com/cloudfoundry/community/blob/main/toc/ROLES.md) documents?

If the answer is 'no', in favor of Paketo's independence as a separate project, should we at least check for potential incompatibilities?

At the very least we should make sure it is compatible (and probably mentions) the [principles](https://github.com/cloudfoundry/community/blob/main/toc/ROLES.md).

loewenstein: I do not see an immediate conflict. 

I still see potential for alignment. E.g. the *steering commitee* members seems to be roughly what the *Paketo working-group leads* would be. Subteam *maintainers* seem to be *approvers* and *contributors* members.

Having said that, in order to let people just work, I'll approve this PR and we'll just figure out what the relationship the Cloud Foundry working groups and the Paketo working group will develop and what the role of the TOC will be in that relationship.

LeePorte (approved):

emalm:
For the record, I approve this PR as worded. (I can't be a reviewer because I started it.)

dsboulder (approved):

PR #171, opened 2021-11-09
Working Agreement: Moving to a PR Only Workflow

ameowlia:
- Everyone is welcome to review and comment.
- We will discuss this in the TOC meeting today (Nov 9th).
- In the TOC meeting next week the TOC members will vote on whether to accept this working agreement.


loewenstein (approved):

totherme:
I like this! I also have a question about automated changes:

- Do we want to insist that all automated-changes get reviewed by a human being?

For example, see [this recent commit history](https://github.com/cloudfoundry-incubator/bosh-backup-and-restore/commits/master?after=fdc9f66f1ceae458381058178ebec9742b776b64+34&branch=master) in the BBR CLI. There are a large number of automated dependency-bump PRs opened by dependabot. These are then tested by a comprehensive CI, and then merged by a CI bot.

I think it's also common practice to have CI bots create final bosh releases as in [this commit in the BBR SDK](https://github.com/cloudfoundry/backup-and-restore-sdk-release/commit/d84431a4b2b267a3e183615a4130a3c58dab8afe).

I would like to advocate for retaining the ability to do these sorts of things. I'm afraid I'm not familiar enough with github branch protection rules to know if the proposed solution will break these workflows or not. It's possible that all I want is an FAQ to reassure me that automated changes of this kind are still welcome, and won't be broken by these branch protections.

stephanme (approved):

pivotal-marcela-campo:
Maybe adding a FAQ section with a note highlighting something in the lines of `predefined approver workflows can bypass this requirement` or alike, that tells maintainers that they can have bots that care for their code on a regular basis but that the approvals are still needed for community contributions?
Also, if we want to allow bots, ideally we could outline a few ways to set up permissions and bots to allow them while also complying with the foundation requirements.

ameowlia:
Hi @totherme and @pivotal-marcela-campo 👋 , 

Thanks for bringing this up. Apparently this is a common questions, especially in the comments of [this github doc about best branch protection practices](https://github.community/t/best-practices-for-protected-branches/10204/6). Here are some options I see: 

1️⃣   **Make the bots admin**
* Allow admins to bypass the review requirements by _not_ checking the "Include Administrators" checkbox on the branch protection rule.
* Maybe not great for security reasons.

2️⃣  **Use github actions**
* Apparently github actions can get around reviews.
* https://github.com/marketplace/actions/merge-me


3️⃣  **Let the bots dismiss pull request reviews**
* I didn't see this online, but it's an idea I had.
* There is a place to add a list of users who can "dismiss pull request reviews"
* Maybe you could update the bot to dismiss reviews before it merges its own PR?

❓ How do these options sound to you?

loewenstein:
I agree that allowing automation makes sense. I also agree that option 1 (allowing admins to merge anyway) is not ideal.

One option could be to have actually two technical users for bots, one that proposes changes and one used for approve and merge.

Istio and Kubernetes are - if I am not mistaken - always merging via Bot based on lgtm labels, approvals, automatic checks and CODEOWNERS files.

ameowlia:
I updated my comment above to add the codeowners option. Thanks @loewenstein !

loewenstein:
> 4️⃣ **Make the bot a codeowner**
> 
> * Add the bot to a codeowner file. [See here for docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-file-location).
> * Allow the codeowners to skip reviews by _not_ checking "Require review from Code Owners"

I wouldn't give anyone the permission to bypass branch protection. Although, it seems fair to give either the WG leads or even WG approvers the possibility to temporarily change that in case of emergency.

What I wanted to say was that we would need two distinct bot identities, one for doing proposals and one for doing approvals and merges (that's at least what we ended up doing).

totherme:
I think having two separate bots, and making one of them an approver/merger, will work great for dependency-bump bots, and many other possible automations -- thanks!

I think it might end up looking _very slightly_ silly in the case of creating bosh final releases. In that case we might be fine with the slight silliness, or we might want to re-visit @ameowlia's idea of using the `CODEOWNERS` file to give some extra power to a particular bot.

**The possible silliness:**

I think the CI flow for a dependency bot will look like:
1. dependencybot creates a bump-PR
2. CI runs some tests
3. mergebot sees the green tests, and so approves and merges the PR

However, I think the flow for a bosh final release may end up looking like:
1. a human creates a PR
2. CI runs some tests
3. a human or a bot merges the PR
4. CI optionally runs more tests
5. boshreleasebot creates a final release PR
6. mergebot immediately merges that PR without any additional tests being run

I think having the mergebot blindly accept the change from the boshrelease bot might feel slightly silly.

I think CI will end up looking like this because bosh really hates it when you create a "final" bosh release, and then delete it. For this reason, I think it's common practice for pipelines to create "dev" releases at the beginning of the pipeline, and "final" releases right at the end -- often as part of a "publish" or "ship" step.

I think the pipeline flow I describe above will work, even if it does look a bit silly having the two bots side-by-side like that. That silly-lookingness may be a small price to pay for what we gain.

**How CODEOWNERS files might help:**
Another option is to do as @ameowlia suggested above with the CODEOWNERS file. If I understand the docs correctly, we can use that file to give a particular user (or bot) extra powers over **a specific directory** in the repo. So perhaps we could give the boshreleasebot the power to merge its own PR, but only if that PR only affects the bosh release metadata directories.

LeePorte (approved):

LeePorte:
@ameowlia There are a few thing on this PR that need resolving, can you have a look when you get a chance?

ameowlia:
I added the bot question as an FAQ. I made the two bot workflow the _suggested_ workflow. This seems like the most secure option. Additionally, there are many merge bots out there that we can grab and "just use". As we enact this IRL if we learn if better alternatives we can update the suggestion.

ameowlia:
Updated to the [RFC format](https://github.com/cloudfoundry/community/tree/main/toc/rfc).

emalm (approved):

dsboulder (approved):

emalm:
Merged, as this has full TOC approval. Will try out https://github.com/cloudfoundry/community/blob/main/toc/rfc/assign-rfc-number.sh to number it automatically. 🤞

PR #186, opened 2022-01-09
Add the repo for CredHub BOSH Release to the scope of the infra WG

bgandon:
Hi CFF fellows,

I've just stumbled upon this repo, which in my opinion should be within the scope of our _Foundational Infrastructure_ Working Group, and should be a good candidate for being donated by VMware to the CFF foundation.

Who should we get in touch with for pushing forward that idea of this repo joining the `cloudfoundry` GitHub org?

Benjamin


loewenstein (changes requested):
As you propose, I think the repository would indeed need to be transferred to the CFF and into the Cloud Foundry Github organization for this change to be merged.

christopherclark:
While this is certainly within the scope of the CFF and F.I. WG, there's a proprietary license attached to this repository that isn't compatible with the Apache 2 license. 

After review from LF Legal, we cannot move this over to the cloudfoundry org as it currently stands.

PR #189, opened 2022-01-18
Remove cf-log-cache-ci user from members list

ctlong:
That user no longer exists / has been deleted


ctlong:
cc @mkocher

ctlong:
Someone removed the user separately to this PR

PR #192, opened 2022-01-24
No dry-run + switch to github app auth

rkoster:
Github org management has been running in dry run mode over the weekend. So far there have been no issues.
This PR changes the value for `--confirm` to `true`. In addition, it switches to using github app auth, so we can use the [cf-foundation github add](https://github.com/organizations/cloudfoundry/settings/apps/cf-foundation-org-manager). This add can be managed by all cloudfoundry org admins.

### DIFF
--- a/.github/workflows/org-management.yml
+++ b/.github/workflows/org-management.yml
@@ -20,4 +20,4 @@ jobs:
         uses: docker://gcr.io/k8s-prow/peribolos
         with:
           entrypoint: /app/prow/cmd/peribolos/app.binary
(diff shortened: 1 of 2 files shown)


halliwilljustin:
@rkoster This workflow will manage the GitHub org membership (addition and removal of members) as well as the team membership? Who will be responsible for merging PRs to the YAML file?

rkoster:
@halliwilljustin discussion around this has moved to: https://github.com/cloudfoundry/community/pull/194

LeePorte (approved):
As #194 has been approved We should be good to approve this PR

PR #195, opened 2022-02-01
update service-mananagement approvers list

pivotal-marcela-campo:


LeePorte:
Hi @pivotal-marcela-campo, are there lilely to be any side effects that you can foresee in removing @saud89? e.g. api tokens that need to be reassigned

LeePorte (approved):

PR #205, opened 2022-02-17
cf-performance-tests repos location

stephanme:
The repos moved from cloudfoundry-incubator to cloudfoundry org.


stephanme:
I re-used the existing teams [App Runtime Interfaces WG CAPI](https://github.com/orgs/cloudfoundry/teams/app-runtime-interfaces-wg-capi) and [go-cf-api](https://github.com/orgs/cloudfoundry/teams/go-cf-api) for write access. This allows the same access rights that we had in cloudfoundry-incubator.

When the automation for managing github teams is in place, I suggest to create a new area like "CAPI incubator".

LeePorte:
superceded by https://github.com/cloudfoundry/community/pull/206

PR #173, opened 2021-11-09
RFC: Project Roadmap + Visibility

geofffranks:
- Everyone is welcome to review and comment.
- We will discuss this in the TOC meeting today (Nov 9th).
- In the TOC meeting next week the TOC members will vote on whether to accept this working agreement.

PR #170, opened 2021-11-09
Working Agreement: GitHub Teams and Access

ameowlia:
* Everyone is welcome to review and comment.
* We will discuss this in the TOC meeting today (Nov 9th). 
* In the TOC meeting next week the TOC members will vote on whether to accept this working agreement.

PR #207, opened 2022-02-18
Acceptance script

LeePorte:
As we've had a couple of RFCs merged in that have required some post acceptance tweaks I thought it may be a good idea to script this process.

It can be tested by commenting out the push and providing a dummy PR number (line75) or commenting out `set -euo pipefail` on a local branch.


LeePorte:
If someone would like to try this out https://github.com/cloudfoundry/community/pull/173 is ready to be merged in

emalm:
At this point, I think #170 and #171 are also candidates to accept soon, along with #173. (We should also make sure that those RFCs all go through the Final Comment Period process, although if we already have unanimous approval from the TOC then we could start and resolve the FCP immediately.)

emalm (approved):
Let's try it on the next RFC that is approved to be merged.

stephanme (approved):

emalm:
Merging, since we have quorum from the TOC with over a week of commentary.

PR #208, opened 2022-02-18
Remove duplicated members

LeePorte:
There are some duplicated members that are causing the current pipeline to fail

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -2780,7 +2780,6 @@ orgs:
         - ddonchev
         - a-b
         - philippthun
-        - Gerg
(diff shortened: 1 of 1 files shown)


rkoster (approved):
LGTM

rkoster:
Have verified the changes by running:
```
spruce json org/cloudfoundry.yml | jq '.orgs.cloudfoundry.teams = (.orgs.cloudfoundry.teams | to_entries | map(.value.members = (try(.value.members - .value.maintainers) // .value.members)) | from_entries)' | bosh int - | grep -v 'members: null' > config.yml
```

And looking at the diff.

PR #212, opened 2022-02-23
Teams Structure for App Runtime Deployments WG

jochenehret:


rkoster (changes requested):
The changes to the cloudfoundry.yml file LTGM, but could you remove the `.idea` directory?

jochenehret:
Oops... removing accidental check-in.
Are the github team names correct? Or should they follow the naming convention "wg-app-runtime-deployments" / "wg-app-runtime-deployments-leads"?

rkoster:
The naming you used is in line with the groups that have already been created: https://github.com/orgs/cloudfoundry/teams?query=WG

On the other hand, naming with dashes is what is being proposed here: https://github.com/cloudfoundry/community/pull/170

To me it does not really matter, because the plan is to eventually generate definitions like the one you are making now based on the working group definition files. Once we do that we can have another discussion about the proper naming of things.

LeePorte:
There was a change to the location of the binary in the docker container that @rkoster fixed yesterday https://github.com/cloudfoundry/community/commit/2634422df561dac60c8e478d1e58d053123a6009. We have also adjusted this to only run against main as it was also running against PRs. The next step it to add a dry run in for use in PRs

LeePorte (approved):

stephanme (approved):

rkoster (approved):

PR #213, opened 2022-02-23
Only run on changes to main

rkoster:
Currently, when people create a PR the proposed changes are already applied.
A further improvement would be to use dry run mode on PR's


LeePorte (approved):
It would be great to have a dry run on PRs so that we can be sure before merging.

PR #215, opened 2022-02-28
Add Alex Rocha to Cloudfoundry members

sethboyles:
Hello,

@xandroc will be working with the App Runtime Interfaces Working Group (on CAPI) so we would like to add him to the CF org.  Thanks!

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -629,6 +629,7 @@ orgs:
     - weymanf
     - will-gant
     - willmadison
+    - xandroc
     - xiongli
     - xtreme-andrei-dinin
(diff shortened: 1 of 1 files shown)


LeePorte:
Hi @sethboyles :wave: 

I'm guessing that you'll need to add @xandroc to the appropriate teams too https://github.com/cloudfoundry/community/pull/215/files#diff-81cd47b9171f8de6dd9563548aa8dd398148ce2d4c63de03942db47a355eacc4R2834

emalm:
@LeePorte I don't think they're requesting adding @xandroc as an approver, so it wouldn't make sense to add them to any WG teams yet.

Gerg (approved):

PR #232, opened 2022-03-23
Add ifindlay-cci to org and Cryogenics team

christopherclark:


emalm (approved):
Approving based on discussion with Chris about membership of the Cryogenics team.

PR #237, opened 2022-03-30
Add David Alvarado as a CAPI team member

tjvman:
Co-authored-by: David Alvarado <alvaradoda@vmware.com>

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -2870,6 +2870,7 @@ orgs:
         - ikasarov
         - fejnartal
         - vkalapov
+        - dalvarado
         privacy: closed
         teams:
           App Runtime Interfaces WG Autoscaler:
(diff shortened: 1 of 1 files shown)


dsboulder:
We don't need to approve the request, WG leads can decide who plays what role in their WGs. We'd also like you to add David to the other WG definition file (sorry for the duplication for now!).

Gerg:
Here is the RFC I was mentioning: https://github.com/cloudfoundry/community/pull/248

dsboulder:
@Gerg You have to edit the charter definition as well. Then I can help get it merged, since it doesn't need TOC approval

Gerg:
See https://github.com/cloudfoundry/community/pull/262

org/cloudfoundry.yml
```diff
@@ -2925,6 +2926,7 @@ orgs:
             - MarcPaquette
             - sweinstein22
             - MerricdeLauney
+            - dalvarado
```
Gerg, line 2929: Does David meet the criteria described in https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0006-approver-requirements.md?

PR #252, opened 2022-04-19
Add korifi-bot to the cloudfoundry org

kieron-dev:
This account is used in korifi CI and needs to list members of the cloudfoundry/cf-k8s team. That requires membership of the cloudfoundry org.


emalm (approved):

emalm:
Merging as per [RFC 0002](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0002-github-members.md), since @kieron-dev is an approver in the CF on K8s WG.

PR #279, opened 2022-05-19
Add ARI RFC for Moving Buildpack Dependency Repositories to CFF AWS Accounts

Gerg:
Previous draft version in [this google doc](https://docs.google.com/document/d/1ElGwXudHUxzUuCBEkG8UXzI1YxqIq_eYJzrHw8PAPI4/edit#).

Rendered version of the RFC: https://github.com/cloudfoundry/community/blob/rfc-move-buildpacks-dependencies-to-cff/toc/rfc/app-runtime-interfaces/rfc-draft-move-buildpacks-dependencies-to-cff.md

### DIFF
--- a/toc/rfc/app-runtime-interfaces/rfc-draft-move-buildpacks-dependencies-to-cff.md
+++ b/toc/rfc/app-runtime-interfaces/rfc-draft-move-buildpacks-dependencies-to-cff.md
@@ -1,0 +1,82 @@
+# Meta
+[meta]: #meta
+- Name: Move Buildpack Dependencies Repository to CFF
+- Start Date: 2022-05-18
+- Author(s): @gerg
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request:
+  [https://github.com/cloudfoundry/community/pull/279](https://github.com/cloudfoundry/community/pull/279)
+
+
+## Summary
+
+Move ownership of the buildpack dependency repositories from VMware to the Cloud
+Found Foundation (CFF).
+
+## Problem
+
+As part of staging, Cloud Foundry buildpacks download their dependencies from a
+VMware-managed central dependency repository. For example, here is an excerpt
+from staging an app using the ruby buildpack:
+```
+   -----> Ruby Buildpack version 1.8.53
+   -----> Supplying Ruby
+   -----> Installing bundler 2.3.11
+   Download [https://buildpacks.cloudfoundry.org/dependencies/bundler/bundler_2.3.11_linux_noarch_any-stack_cab63d95.tgz]
+   -----> Installing ruby 2.7.5
+   Download [https://buildpacks.cloudfoundry.org/dependencies/ruby/ruby_2.7.5_linux_x64_cflinuxfs3_2c25fe7d.tgz]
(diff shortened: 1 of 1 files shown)


Gerg:
FYI @dmikusa-pivotal @pivotal-david-osullivan @arjun024 @brayanhenao

LeePorte:
@christopherclark is is possible to contact our AWS TAM to ask about transferring large buckets between accounts?

Gerg:
@LeePorte From their support site:
> Warning: AWS Support can't transfer bucket ownership or copy Amazon S3 objects or manipulate any configuration options in AWS accounts. Also, you can't separate an AWS account from an Amazon.com account.

Source: https://aws.amazon.com/premiumsupport/knowledge-center/account-transfer-s3/

Gerg:
Other ideas we discussed in this weeks TOC meeting:

1. Moving over the s3 buckets in addition to the CloudFront distributions
    - Probably too expensive with current bucket sizes
    - Maybe we can consider as a follow-up project if we can reduce bucket sizes
1. Deleting old dependencies to reduce bucket sizes
    - Buildpacks don't have a deprecation policy for this currently
    - If buildpacks WG had more direct access to the download statistics, making decisions like this might be easier
1. Re-architect buildpacks to remove centralized dependency cache
1. Move dependency cache to github asset storage instead of AWS
1. Worst case, VMware can fall back on working with the Linux Foundation to manually renew the certificates 
    - This is a bit risky long-term, since it involves a repeated manual process every year

Gerg:
@cloudfoundry/toc I motion to commence the Final Comment Period

emalm:
@Gerg Do we have an estimate on what the CloudFront costs are per month?

You also mention in the "cons" section that there could be some unforeseen issues in using CloudFront and S3 across 2 different AWS accounts: is there a plan to discover and mitigate those potential issues?

emalm:
Note from TOC meeting discussion on 2022-06-07: let's document that we'll try setting up CloudFront for this in the CFF account and monitor the cost when we switch over, and if it's more expensive than expected we'll switch it back to the VMware account until we figure something out.

Gerg:
> an estimate on what the CloudFront costs are per month?

I wasn't able to get a detailed breakdown from the team currently operating those CloudFront distributions, but they said the current cost for all CloudFront distributions on that account is ~$10k/mo.

Gerg:
@cloudfoundry/toc This issue has kind of stalled out. What are the next steps to getting it approved?

emalm:
@Gerg We had already said in https://github.com/cloudfoundry/community/pull/279#issuecomment-1154554921 that we're ok trying this out, but we'd like to document in the RFC text that we'll monitor the cost over some period of time after the CloudFront switchover and if it is more expensive than expected we'll switch it back to evaluate other options. Could you please update the proposal to that effect? Thanks!

Gerg:
@cloudfoundry/toc  I updated the RFC to incorporate your feedback. Please let me know if you would like any additional changes. Otherwise, I motion to commence the Final Comment Period.

AP-Hunt:
@Gerg TOC agrees. We start the final comment period. Ends 26 July.

stephanme (approved):

emalm (approved):

PR #282, opened 2022-05-20
Propose using Elekto as the system for the 2022 TOC election

emalm:
* Amend the CFF charter to permit a CFF-managed Elekto instance as an election system

PR #288, opened 2022-05-26
Update Readme.md

christopherclark:
Updates voting method to Elekto


LeePorte (approved):

emalm:
@christopherclark there's some overlap with changes proposed in https://github.com/cloudfoundry/community/pull/282/files#diff-cc10a2b6016c0e7531dade79086f2073501a70fe182a7b3c298ee46fd3c9bc0b, so should we just consolidate on that other pull request instead? I had overlooked referring to Elekto in the schedule table.

PR #296, opened 2022-06-06
Add candidates for TOC election to election README

christopherclark:


emalm (approved):
LGTM. Will merge since we're just making the readme reflect reality.

PR #295, opened 2022-06-03
Update ROLES.md to match RFC 0006 and RFC 0008

moleske:
- https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0006-approver-requirements.md and https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0008-role-change-process.md clarify and change some requirements and promotion/revocation processes for the different roles.  This was hard to find as it was buried in the RFC section of this repository.  Since synchronizes ROLES.md since this is where I think people would look first for how these roles and processes currently work

I was trying to figure the where I was qualification wise to become an approver and saw contradictions here from what other people pointed out to me in the RFC directory.  This is mostly a copy/paste from the RFC with minimal markdown styling changes

PR #305, opened 2022-06-10
update approver requirements per RFC 0006

ameowlia:
In [RFC 0006](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0006-approver-requirements.md) we agreed to update the approver requirements. This PR actually updates them.


Gerg:
I think this overlaps some with https://github.com/cloudfoundry/community/pull/295

ameowlia:
I will leave it with that PR then.

PR #308, opened 2022-06-14
[RFC] Cleanup Stemcells

rkoster:
Proposal to start cleanup up stemcells after x amount of year(s).


rkoster:
cc @cloudfoundry/foundational-infrastructure-wg

toc/rfc/rfc-draft-stemcell-cleanup.md
```diff
@@ -0,0 +1,25 @@
...
+
+
+## Summary
+
+In an effort to reduce ongoing IaaS spend of the Cloud Foundry Foundation this RFC proposes cleanup schedules for light and heavy stemcells.
+
+## Problem
+
+Currently heavy stemcells (full stemcell images stored in s3/gcs) are stored indefinitly. This is also true for the images backing light stemcells.
+This becomes a costly endevour, especially on AWS where we need to keep a EBS snapshot per region per light stemcell.
```
pburkholder, line 17: `keep an EBS snapshot` instead of `keep a EBS snapshot`.

toc/rfc/rfc-draft-stemcell-cleanup.md
```diff
@@ -0,0 +1,25 @@
...
+
+It has been estimated that cleaning up old stemcells could save tens of thousands of dollars.
+
+## Proposal
+
+Cleanup light stemcells older then 1 year.
+Cleanup heavy stemcells older then 3 years.
```
pburkholder, line 24: `older _than_ 1 year`, etc.

pburkholder:
Can you link to, or provide, the rough estimate that this would save 10s of thousands of dollars (per year)? 

Seems worthwhile even if it only saves thousands of dollars per year.

If someone _really_ wants to recreate CF from 2012, they'd still be able to build their own stemcell, right?

norman-abramovitz:
For the light stemcells it is dependent on the cloud provider.   I would suggest following the cloud provider policies.  If that is not reasonable, I do not understand why the light stemcells would be retired any sooner than the heavy stemcells?

norman-abramovitz:
If the light stemcell is somehow available from the Cloud Provider, maybe point to where it is stored by Cloud Provider.

nouseforaname:
That's exactly the problem for AWS case, light stemcells are hosted in "our" accounts. One copy per region served and unlike heavy stemcells (served from S3 buckets and global) they do not support cheaper storage classes. They're basically EBS snapshots lying around in some account that people can access. 

> If the light stemcell is somehow available from the Cloud Provider, maybe point to where it is stored by Cloud Provider.

LeePorte:
> If someone _really_ wants to recreate CF from 2012, they'd still be able to build their own stemcell, right?

Yep someone could build their own old stemcell using https://github.com/cloudfoundry/bosh-linux-stemcell-builder

jimconner:
I think that removing old stemcells is an excellent idea. Older stemcells have known security vulnerabilities and weaknesses, so getting rid of them reduces the chances of someone accidentally using them (and also reduces hosting costs). Would be good to do a similar exercise with the java-buildpack assets.. getting rid of the known-vulnerable and obsolete versions of tomcat, jdk etc.

rkoster:
> 

An ex-Pivotal account that hosts public facing AMIs and has 30541 EBS snapshots aged 6 months or more, with the oldest snapshot created in 2016. Even though storing a snapshot costs about 15 cents per month, the aggregate spend is currently about $3.3K per month, or $40K per year.

ramonskie:
in total its about 20cent per stemcell per region as the ami itself also cost money.
as we push to 19 regions including 2 gov regions.
it will cost roughly 3.80 per stemcell

rkoster:
Cost example for AWS lite stemcells:
15 cents per month per snapshot * 19 regions (including 2 gov regions) = 2,85 dollar per month per stemcell
Lets assume a stemcell every 3 weeks (more if high CVE's are released) 52/3 * annual cost per stemcell (2,85*12) = $600 dollar per year.

If we would keep 3 years of light stemcells that would cost around $1800 per year. Not accounting for bandwidth usage, which should not be affected by cleaning up stemcells or not.

Given the above calculations, I would advocate for simplifying the policy to 3 years for all stemcells (heavy, light, windows).

1,8k per year compared to the current 40k a year would be a great improvement.

beyhan:
I think that it makes sense to announce this change in the community and define a date when it will be implemented so that users requiring older versions of the  stamecells have also as option to download them before they get removed.

rkoster:
Have updated the proposal with 3 years retention for all stemcells + cleanup of unpublished candidate stemcells after 2 months.

rkoster:
https://github.com/orgs/cloudfoundry/teams/toc called for and approved the Final Comment Period for this RFC, so it is now open for final commentary until next Tuesday, July 5th.

toc/rfc/rfc-draft-stemcell-cleanup.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Problem
+
+Currently heavy stemcells (full stemcell images stored in s3/gcs) are stored indefinitly. This is also true for the images backing light stemcells.
+This becomes a costly endevour, especially on AWS where we need to keep an EBS snapshot per region per light stemcell.
+
+It has been estimated that cleaning up old stemcells could save tens of thousands of dollars.
+
+In addition to published stemcells we also store candiate stemcells which never have been published indefinitly.
+These should be save to cleanup after we are sure we are not gonna publish them.
```
emalm, line 22:
```suggestion
These should be safe to cleanup after we are sure we are not going to publish them.
```

toc/rfc/rfc-draft-stemcell-cleanup.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Proposal
+
+- Cleanup stemcells (heavy, light, windows, ect) older than 3 years.
+- Cleanup unpublished candidate stemcells after 2 months.
```
emalm, line 27:
```suggestion
- Clean up stemcells (heavy, light, windows, etc.) older than 3 years.
- Clean up unpublished candidate stemcells after 2 months.
```

emalm (changes requested):
Generally looks fine to me, but I requested some minor edits for style.

emalm (approved):

stephanme (approved):

PR #325, opened 2022-06-28
Transfer BOSH Bootloader to App Runtime deployments

rkoster:
This project was initially given to FI, probably because it has BOSH in its name, and the deployment working group did not exist yet, however, this project actually requires intimate knowledge of cf-deployment to be able to properly maintain it.

This PR proposes to move BOSH Bootloader and the related concourse bbl state resource to the App Runtime Deployments working group.

cc @jochenehret

### DIFF
--- a/toc/working-groups/app-runtime-deployments.md
+++ b/toc/working-groups/app-runtime-deployments.md
@@ -77,4 +77,6 @@ areas:
   - cloudfoundry/cf-test-helpers
   - cloudfoundry/uptimer
   - cloudfoundry/runtime-ci
+  - cloudfoundry/bosh-bootloader
+  - cloudfoundry/bbl-state-resource
 ```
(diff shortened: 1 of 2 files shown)


jochenehret (approved):
Ok, if those projects don't have a home yet...

emalm (approved):

davewalter:
The App Runtime Deployments working group is currently engaged in transferring testing of its assets (CF-Deployment, Smoke tests and CATs) to our new CI. Can we please defer this decision until that work is complete?

jochenehret:
Discusses this PR in the ARD WG meeting on July 14th: We should try to find a better home for this project, possibly the Foundational Infrastructure WG. ARD WG knows the project, but will likely have no capacity for maintenance. TOC will discuss this.

rkoster:
We had an offline meeting about this, there is currently no appetite to move this project over. This is because the App Runtime Deployments team does not feel comfortable with the BOSH context. The path forward will require us to split BBL into separate terraform modules maintained by the working groups with the most context (BOSH-related stuff by FI and CF load balancers by ARD).

This work however requires a not insignificant amount of work and is currently going through prioritization internally at VMware. At this point in time it is uncertain when this work will be picked up.

PR #336, opened 2022-07-06
Remove non existing github users

stephanme:
... that lead to failing github org automation.

- cf-pub-tools
- senatorsupes


emalm (approved):

emalm:
Merged, as this is uncontroversial and about the only thing we could do here.

PR #337, opened 2022-07-07
Request bruce as approver

strehle:
https://github.com/bruce-ricard should be in addition to Peter and myself


strehle:
@rkoster FYI

PR #341, opened 2022-07-12
[ARI] Remove App Autoscaler dev team access to `runtime-og-ci-private`

silvestre:
We (the current team) have never used `runtime-og-ci-private` and have no need to have access to it.


Gerg (approved):
Since `cf-autoscaler` is a legacy team, I'm going to merge this without going through the process described in [RFC-0007](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0007-repository-ownership.md).

PR #343, opened 2022-07-12
archive the event log release

Benjamintf1:


stephanme:
Looks like archiving by peribolos is not allowed by default:

https://github.com/cloudfoundry/community/runs/7309279143?check_suite_focus=true
```
{"component":"peribolos","file":"k8s.io/test-infra/prow/cmd/peribolos/main.go:1036","func":"main.configureRepos","level":"info","msg":"repo exists, considering an update","repo":"event-log-release","severity":"info","time":"2022-07-12T22:25:19Z"}
{"component":"peribolos","error":"asked to archive a repo but this is not allowed by default (see --allow-repo-archival)","file":"k8s.io/test-infra/prow/cmd/peribolos/main.go:1040","func":"main.configureRepos","level":"error","msg":"requested repo change is not allowed, removing from delta","repo":"event-log-release","severity":"error","time":"2022-07-12T22:25:19Z"}
```

Could add the flag --allow-repo-archival.

stephanme:
@rkoster: Maybe you archive event-log-release via UI for now to get the automation working again.

If archiving has no side effects, we can enable it in the automation. Alternative is to treat archiving as a special case that needs to be done manually via UI (in addition to the PR on cloudfoundry.yml).

We can then discuss in next TOC meeting.

Benjamintf1:
thanks!!!

Issue #342, opened 2022-07-12
'Check Github Organization Settings PRs' check fails for PRs from forks


stephanme:
PRs on WG charters, contributors.yml and cloudfoundry.yml trigger a github action workflow 'Check Github Organization Settings PRs'.

The workflow generate the resulting cloudfoundry org configuration and triggers a dry-run of peribolos.

The validation workflow succeeds for PRs created from a branch of the community repo (by users with write access to the community repo). E.g. #336, #334, #322

The validation workflow fails for PRs created from forks of the community repo (by users that don't have write access to the community repo). E.g. #338, #340, #329

It fails with am 'Unauthorized" error when reading org invitations: 
```
[...]

PR #345, opened 2022-07-13
#342: remove peribolos from PR validation check

stephanme:
org_management.py validates yaml of WG charters, cloudfoundry.zml
and contributors.yml and checks that org configuration can be generated.


stephanme:
There are many PRs on WG charters for which the validation fails due to peribolos and github token access rights.

I interpret the thumbs-up on https://github.com/cloudfoundry/community/issues/342#issuecomment-1181935515 as an approval for this PR and merge it.

Issue #349, opened 2022-07-13
Newly generated area approver team names are not RFC 0005 compliant


silvestre:
According to [RFC 0005 Standardizing Github Teams and Access](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0005-github-teams-and-access.md#proposal) the team name for approvers is supposed to have an `-aprovers` suffix:

| Name of Team  | Team Membership  | Permissions  |
--- | --- | ---
| wg-[WORKING-GROUP-NAME]-[AREA-NAME]-approvers | Approvers & tech leads for an area within a WG | Write access for all repos in the area |

However the org management code is omitting that suffix:

https://github.com/cloudfoundry/community/blob/eae328f4c0d4a79b9a91abb74e2f13859608e1c6/org/org_management.py#L241

(E.g. https://github.com/orgs/cloudfoundry/teams/wg-app-runtime-interfaces-autoscaler)

With https://github.com/cloudfoundry/community/pull/333/files#r916844866 proposing to add a `-reviewers` team omitting the suffix here could cause confusion down the line.

stephanme:
You are right. And nobody noticed so far. I'll prepare a fix.
This also means that any usage of the generated teams e.g. in Concourse or branch protection rules need to be adapted.

PR #262, opened 2022-04-28
Add RFC to remove non-standard Github teams

Gerg:
🌶

https://cloudfoundry.slack.com/archives/C026XJGNC2U/p1651186805030849

PR #350, opened 2022-07-13
#349: fix github team name for WG area approvers

stephanme:
- add missing '-approvers' suffix as required by RFC-0005

### DIFF
--- a/org/org_management.py
+++ b/org/org_management.py
@@ -238,7 +238,7 @@ class OrgGenerator:
             "maintainers": sorted(maintainers),
             "members": sorted(approvers - maintainers),
             "teams": {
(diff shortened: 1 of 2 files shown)


stephanme:
This change as side effects: any usage of the exiting teams e.g. in Concourse or branch protection rules needs to be adapted.

But better now than later as [the spicy PR](https://github.com/cloudfoundry/community/pull/262) is knocking on the door.

emalm (approved):

stephanme:
Rebased to test fix for #342.

emalm:
@cloudfoundry/wg-leads FYI, if you've set up branch-protection rules or concourse pipelines using the previous area-scoped teams, those team names will get changed once the automation runs with this fix.

PR #353, opened 2022-07-15
Add --allow-repo-archival to org automation

stephanme:
Archiving is triggered by adding "archived: true" to the repository entry
in cloudfoundry.yml.

Org automation fails without --allow-repo-archival.

### DIFF
--- a/.github/workflows/org-management.yml
+++ b/.github/workflows/org-management.yml
@@ -77,6 +77,7 @@ jobs:
             --fix-teams
             --fix-team-members
             --fix-team-repos
+            --allow-repo-archival
       - name: debug
         run: |
           echo "steps.peribolos.outcome = ${{ steps.peribolos.outcome }}"


stephanme:
Org automation currently fails because peribolos is not allowed to archive repos. See also #343.

https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories mentions that contributors and teams can't be changed anymore for archived repos. This may impact the removal of non-standard teams.

Alternative is to archive the repos manually (in addition to the PR that adds the archived flag in cloudfoundry.yml). This manual archiving could also include the removal of all teams from the repo.

Example of a repo to be archived: https://github.com/cloudfoundry/event-log-release/settings/access

stephanme:
Team deletion with peribolos works even if the team is referenced by an archived repository. Tested with https://github.com/cloudfoundry/event-log-release which I manually archived. It referenced a WG area team that was deleted by #350 (missing -approvers suffix).

rkoster (approved):

PR #358, opened 2022-07-15
Add new CF for K8s area to CF on K8s WG

davewalter:
Add missing CF for K8s repositories to the CF on K8s working group.

### DIFF
--- a/toc/working-groups/cf-on-k8s.md
+++ b/toc/working-groups/cf-on-k8s.md
@@ -35,7 +35,7 @@ execution_leads:
 technical_leads:
 - name: Giuseppe Capizzi
   github: gcapizzi
-- name: George
+- name: Georgi Sabev
   github: georgethebeatle
 bots:
 - name: eirinici
@@ -43,7 +43,7 @@ bots:
 - name: korifi-bot
   github: korifi-bot
 areas:
-- name: CF on k8s
+- name: Korifi
   approvers:
   - name: Andrew Wittrock
     github: Birdrock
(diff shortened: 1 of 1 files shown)


emalm:
If we're pulling in the cf-for-k8s repos for the time being until we retire them completely, should we have a new technical area for them that's separate from the existing "CF on K8s" one (which maybe we should also rename "Korifi" to match the new official project name)?

davewalter:
@emalm We updated the PR to incorporate your suggestions.

On a separate topic, is there a process for removing approvers who are no longer full-time contributors to a project? We noticed that Akira and Mario are still listed as approvers for Korifi.

emalm:
Thanks!

> is there a process for removing approvers who are no longer full-time contributors to a project?

Yep, https://github.com/cloudfoundry/community/blob/main/toc/ROLES.md#revoking-approver-role now represents the process that was approved in [RFC 0008](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0008-role-change-process.md).

Birdrock (approved):

toc/working-groups/cf-on-k8s.md
```diff
@@ -68,12 +68,29 @@ areas:
   - name: Clint Yoshimura
     github: clintyoshimura
   repositories:
+  - cloudfoundry/cf-crd-explorations
```
emalm, line 71: This repository is archived, so it shouldn't be under the ownership of any WG.
```suggestion
```

emalm (changes requested):
Minor change requested to exclude the archived CRD repository, but otherwise LGTM!

emalm (approved):

stephanme (approved):

PR #362, opened 2022-07-18
Update app-runtime-platform.md

mariash:


ameowlia:
@emalm Updated here: https://github.com/cloudfoundry/community/pull/366/files

PR #317, opened 2022-06-23
Enable generation of github teams for all WGs

stephanme:

PR #367, opened 2022-07-20
[ARP] Add tas-runtime-bot as a bot

ctlong:
🔑 Most of the current ARP CI uses this bot's credentials to interact with ARP's GitHub repos.

👎 The bot currently gets its access from the `runtime` team, which will likely be deleted when #262 is implemented.

➡️ Bring the bot account into the fold as a working group bot.

### DIFF
--- a/toc/working-groups/app-runtime-platform.md
+++ b/toc/working-groups/app-runtime-platform.md
@@ -40,7 +40,9 @@ execution_leads:
 technical_leads:
 - name: Amelia Downs
   github: ameowlia
-bots: []
+bots:
+- name: CI bot
+  github: tas-runtime-bot
 areas:
 - name: Diego
   approvers:


ctlong:
cc @ameowlia

ameowlia (approved):
LGTM

mkocher:
@emalm are you able to merge this for us or tell us what we need to do to get it merged? We'll have the same question for #377 as well.

emalm:
@mkocher Hmm, I don't think we actually defined the approval process for WG bots when #317 introduced the role and the automation supporting it. In terms of permissions, they're equivalent to an approver across all the WG areas, so from [RFC 0008](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0008-role-change-process.md) admitting one seems like the decision of the WG leads, with the TOC overriding at its discretion. I'm adding this topic to the TOC agenda for tomorrow for discussion.

Should we have a TAS-associated bot with that level of access to a CFF WG and all its repos, though?

stephanme:
#333 + #375 + #378 introduce WG area reviewers and WG area bots. Still WIP.
If accepted, RFC-0008 needs a replacement similar to #375 for RFC-0005.

ctlong:
> admitting one seems like the decision of the WG leads, with the TOC overriding at its discretion

Amelia has already approved this PR, but is currently OOO. If this decision is under her purview then it seems like the decision has been made in the positive, right? Ordinarily we could just wait for her to return, but given the upcoming non-standard GH teams removal we want to get this in sooner rather than later.

emalm (approved):
I approve these changes in the interest of keeping teams unblocked when #387 gets merged soon. Once we resolve #386 to establish better guidelines about bot account management, though, we should review all the bots with permissions within WGs or areas.

PR #377, opened 2022-07-28
Add runtime bot to foundational infrastructure for syslog/etc

Benjamintf1:
Most of the current CI uses this bot's credentials to interact with System Logging and Metrics (rsyslog / event-log) GitHub repos.

The bot currently gets its access from the runtime team, which will likely be deleted when https://github.com/cloudfoundry/community/pull/262 is implemented.

Bring the bot account into the fold as a working group bot.


mkocher:
@rkoster does this look good to you? /cc @ctlong

rkoster (approved):

PR #379, opened 2022-07-29
Add stefanlay to App Runtime Platform Networking Approvers/Main group

geofffranks:
Per [the App Runtime Platform WG page](https://github.com/cloudfoundry/community/blob/main/toc/working-groups/app-runtime-platform.md),
@stefanlay is an authorized approver. They are in the
`wg-app-runtime-platform-networking-approvers` group already, but absent from
`app-runtime-platform-wg-networking` and
`app-runtime-platform-wg-networking-reviewers`.

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -2957,6 +2957,7 @@ orgs:
             - reneighbor
             - selzoc
             - MarcPaquette
+            - stefanlay
           App Runtime Platform WG Networking:
             members:
               - aminjam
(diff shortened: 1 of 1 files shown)


stephanme:
This doesn't make much sense anymore.

#262 will remove all non-standard github teams soon, i.e. all teams that don't comply with https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0005-github-teams-and-access.md

Standard teams for reviewers per WG area is WIP (#333).

PR #376, opened 2022-07-28
Move all org admins into TOC charter

stephanme:
- execution leads = TOC = toc github team
- technical leads = CFF staff
- both get cloudfoundry org admins

PR #383, opened 2022-08-03
Remove mjear from all roles and teams

stephanme:
Was removed from org admins by #376 and now automation fails because there
are still references in non-standard teams.

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -4620,7 +4620,6 @@ orgs:
         maintainers:
         - dsboulder
         - emalm
-        - mjear
         members:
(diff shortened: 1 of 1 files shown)


stephanme:
Alternative is to remove @mjear from the team 'Directors-PMs-Anchors' and to remove the complete team 'License Finder'.
Both teams are non-standard teams that are going to be deleted by #262.

emalm:
I think it'd be fine to remove @mjear from that legacy team and to delete the License Finder team entirely.

emalm (approved):

AP-Hunt (approved):

PR #385, opened 2022-08-03
Add Seth Boyles to back to CAPI approvers

sethboyles:
I was erroneously removed in this commit: https://github.com/cloudfoundry/community/commit/72f17ed4719eb513e2aac3f2430b7ea385127a7b


Gerg:
Since you are a returning approver, I think we can bypass the ordinary promotion process.

Gerg (approved):

PR #387, opened 2022-08-09
Remove non-standard github teams

stephanme:
🌶

As decided in https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0013-remove-nonstandard-github-teams.md


davewalter (approved):
Looks good. I _think_ the App Runtime Deployments working group is ready. 🤞 We still have some decisions to make around where `bosh-bootloader` and `jumpbox-deployment`, but at least they are currently included in a working group, so we can adjust as needed.

org/cloudfoundry.yml
```diff
@@ -2287,4357 +2287,4 @@ orgs:
...
-          volman: admin
-          winc: admin
-          winc-release: admin
-          windows-syslog-release: admin
-          windows-tools-release: admin
-          windows2016fs: admin
-          windows2019fs-release: admin
-          windowsfs-online-release: admin
-          workpool: admin
-      wg-service-management-service-fabrik-approvers:
```
pivotal-marcela-campo, line 6331: Will this be removed and then recreated based on the WG charter? Or deletion will be skipped because it is a group included in a wg charter? I am just concerned there will be a gap where the team will lose access to the repo.

stephanme: wg-service-management-service-fabrik-approvers is a standard github team. It will not be removed and there won't be a gap.
The [org automation](https://github.com/cloudfoundry/community/tree/main/org) generates the standard teams from the WG charters and merges the generated teams into the static config of cloudfoundry.yml. The result is then applied by [peribolos](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/peribolos).

pivotal-marcela-campo: Thanks for the clarification Stephan!

pivotal-marcela-campo (approved):

rkoster (approved):

emalm:
@Benjamintf1 @ctlong Does the ARP WG or any of its areas need to keep around any of these non-standard teams on an interim basis? If so, please suggest changes within the next week (by 2022-08-22) on which ones you would need to keep. Thanks!

Benjamintf1:
I think our most major concerns are keeping our ability to use bots(https://github.com/cloudfoundry/community/pull/367, https://github.com/cloudfoundry/community/pull/377) for pipelines to keep golang updating and keep being able to easily create releases/etc.

AP-Hunt:
I've rebased this PR on `main` to resolve a conflict in `org/cloudfoundry.yml`

emalm (approved):
Let's plan to merge this on Tuesday, 2022-09-06, after the US gets back from the Labor Day holiday.

AP-Hunt (approved):

PR #393, opened 2022-08-11
wg-service-management: add area specific bots

pivotal-marcela-campo:
Adding area specific bots to the wg charter

### DIFF
--- a/org/contributors.yml
+++ b/org/contributors.yml
@@ -538,7 +538,6 @@ contributors:
 - sebGilR
 - selzoc
 - services-api-ci
-- servicesenablement
 - sethboyles
 - shamus
 - Shanfan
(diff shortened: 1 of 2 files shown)


rkoster (approved):

org/contributors.yml
```diff
@@ -538,6 +538,7 @@ contributors:
 - sebGilR
 - selzoc
 - services-api-ci
+- servicesenablement
```
stephanme, line 541: A bot should not be listed as contributor. Org automation is supposed to add the bots as github org member.

stephanme (changes requested):

org/contributors.yml
pivotal-marcela-campo: Done!

stephanme (approved):

PR #395, opened 2022-08-18
Reviewers and area bots shall be org members

stephanme:
... even if not contained in contributors.yml
(especially the bots)


stephanme:
This should fix the disappearing area bots from #393

rkoster (approved):
LGTM

PR #413, opened 2022-09-22
Reject duplicate keys in WG charters etc.

stephanme:
Duplicate keys can lead to missing WG areas and teams. pyyaml silently ignores duplicate keys instead of rejecting them.


stephanme:
This should prevent problems like https://cloudfoundry.slack.com/archives/C026XJGNC2U/p1663842442887559

rkoster (approved):

emalm (approved):

PR #421, opened 2022-09-22
remove identity-ci

bruce-ricard:
We moved this repo in the pivotal org here: https://github.com/pivotal/uaa-ci as it is a private repo containing private information.

The whole repo can be deleted.

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -1328,10 +1328,6 @@ orgs:
       ibm-websphere-liberty-buildpack:
         description: IBM WebSphere Application Server Liberty Buildpack
         has_projects: true
-      identity-ci:
-        description: where everything else lives
-        has_projects: true
-        private: true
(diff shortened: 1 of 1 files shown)


rkoster (approved):

rkoster (changes requested):
@bruce-ricard could you update the PR to add `archive: true` instead of removing the repo from the list?

bruce-ricard:
I don't really want to archive it though, because we are not archiving it, we are moving it.

It feels odd to copy it to another org, and archive it here.

But I'll do it if that's what you want. Let me know.

rkoster:
It's about intellectual property being moved out of a foundation-owned GitHub org, which is why it's better to archive the repo and update the Readme with a link to the new location.

beyhan:
Closing in favour of https://github.com/cloudfoundry/community/pull/697

PR #435, opened 2022-09-29
Add garden bot to ARP bots list

ameowlia:
We are quickly learning which bots we missed :D


emalm (approved):
Approving before the gnome disappears into its burrow.

Issue #438, opened 2022-09-30
Branch protection rules should become a part of the automation


Benjamintf1:
I'd suggest that we have standardized rules for repositories for branch protection rules. I suggest the following for those rules.

Which branches?:

I'd suggest that branch protection rules should be applied to: `main` and to `v[0-9]` branches. This should give some protection to release line branches, while allowing for non approvers to still be able to create branches in cloudfoundry rather then having to keep and maintain forks. This should cover golang libraries well as well. 

While certain repos have a `develop` or `release-elect` pattern, I suggest these patterns be considered deprecated as per discussion about better following open source patterns of development

Rules:
Require a pull request before merging (this encourages good practice with regards to making sure changes that are submitted are good, but also fostering open discussion and tracking of meaningful changes). 
    Required Approvals: I set this to 1 before the permissions changes. I think this is a good idea. Arguably, it only prevents accepting your own pr if you are a approver, and then merging. You don't need two approvers, you just need 2 people of which one is an approver. For some areas we only *have* one approver or one very active approver. It doesn't do everything. I added it mostly to indicate ci requirements, but it's not technecially necessary for it.
    Dismiss stale pull request approvals when new commits are pushed: I think this one is important and good for ci requirements. Prevents ddos and other such attacks against ci infrastructure, especially ci infrastructure that provisions iaas resources. We right now run against forks in logging-and-metric things maintained by our ci, and allowing running against forks is condusive to accepting contributions, but having this rule protects against attacks.
    Require review from Code Owners: I don't think we need this. Would require keeping owners up to date, sounds like a lot of toil for not a lot of value.
    Restrict who can dismiss pull request reviews: No opinions. Hasn't been a problem so far.
    Allow specified actors to bypass required pull requests: I'd like this to include bots. Specifically, it'd be nice to continue to use ci to cut releases and bump submodules/golang(a not great situation, but still nice), and creating releases requires creating a commit and adding it to a branch for bosh. 
    
    Require status checks to pass before merging: Seems unnecessary to me. Obviously checks can be important, but I think it discourages adding new tests, and should be left to approver discression.
    
Require conversation resolution before merging: Seems a bit like busy work to me? Could be left to discression?

Require signed commits: Seems a bit onerious to me for little benefit. 
Require linear history: I don't see why this should be a requirement
Require deployments to succeed before merging: I htink this should be approver specific, that said I don't think any repos use deployments.
Do not allow bypassing the above settings: I think this should be false
[...]

PR #449, opened 2022-10-07
Add draft RFC for default git branch protection rules

AP-Hunt:
Following from discussion at the TOC meeting on 4th October 2022, I've written up a draft RFC to formalise the discussion around branch protection rules which has happened in the TOC meeting and in #438.

I've added @Benjamintf1 and @rkoster as authors to acknowledge that while I put the words together, the ideas are theirs.

### DIFF
--- a/toc/rfc/rfc-draft-branch-protection.md
+++ b/toc/rfc/rfc-draft-branch-protection.md
@@ -29,6 +29,9 @@ The CFF TOC should implement basic branch protection rules for all repositories 
 We propose that protection be applied to all branches matching the expressions `main` or `v[0-9]*`.
 
 On protected branches, we propose the following rules:
-* working group leads, approvers, and bots will be able to push directly to the branch
-* contributors outside of leads, approvers, and bots must raise a pull request before the change can be merged
-* pull requests must have at least 2 approvals, at least 1 of which must come from a working group approver or lead.
+* only bots may push directly to the branch
+* human contributors must make contributions through a regular pull request workflow.
+
+With respect to the approval of pull requests, we propose that the number of approvals required will depend on the number of people in the approver role of a working group:
+* 0 approvals will be required when a working group has 3 or fewer people in the approver rule
+* 1 approval will be required when a working group has 4 or more people in the approver role.
--- a/toc/rfc/rfc-draft-branch-protection.md
+++ b/toc/rfc/rfc-draft-branch-protection.md
@@ -11,7 +11,7 @@
 
 In Q3 2022, the technical oversight committe accepted and implemented [an RFC for the managemnet of Github teams and user access](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0014-github-teams-and-access.md). This unintentionally broke the workflow and branch protection roles of some working groups.
 
-This RFC seeks to address that by defining a baseline of branch protection rules across all repositories under management in the cloufoundry GitHub org.
+This RFC seeks to address that by defining a baseline of branch protection rules across all repositories under management in the cloudfoundry GitHub org.
 
 ## Problem
 
--- a/toc/rfc/rfc-draft-branch-protection.md
+++ b/toc/rfc/rfc-draft-branch-protection.md
@@ -26,7 +26,7 @@ When the TOC approved and implemented [RFC 0013](https://github.com/cloudfoundry
 
 The CFF TOC should implement basic branch protection rules for all repositories in an automated manner, and allow working groups to extend them as they see fit. This can be achieved, technically, with [the `branchprotector` tool from `prow` toolset](https://github.com/kubernetes/test-infra/blob/master/prow/cmd/branchprotector/README.md).
 
-We propose that protection be applied to all branches matching the expressions `main` or `v[0-9]*`.
+We propose that protection be applied to all branches matching either the default branch or the expression v`v[0-9]*`.
 
 On protected branches, we propose the following rules:
 * only bots may push directly to the branch
--- a/toc/rfc/rfc-draft-branch-protection.md
+++ b/toc/rfc/rfc-draft-branch-protection.md
@@ -1,0 +1,34 @@
+# Meta
+[meta]: #meta
+- Name: Default git branch protection rules
+- Start Date: 2022-10-07
+- Author(s): AP-Hunt, Benjamintf1, rkoster
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: https://github.com/cloudfoundry/community/pull/449
+
+
+## Summary
+
+In Q3 2022, the technical oversight committe accepted and implemented [an RFC for the managemnet of Github teams and user access](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0014-github-teams-and-access.md). This unintentionally broke the workflow and branch protection roles of some working groups.
+
+This RFC seeks to address that by defining a baseline of branch protection rules across all repositories under management in the cloufoundry GitHub org.
+
+## Problem
+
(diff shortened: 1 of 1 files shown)


beyhan:
My personal opinion is:
* PR only workflow for approvers (If we could find a way to simplify this when there are CI issues would be great)
* More simplicity for bots

rkoster:
```
Every change in the revision’s history was agreed to by two trusted persons prior to submission, and both of these trusted persons were strongly authenticated. (Exceptions from [Verified History](https://slsa.dev/spec/v0.1/requirements#verified-history) apply here as well.)

The following combinations are acceptable:
Uploader and reviewer are two different trusted persons.
Two different reviewers are trusted persons.
```
https://slsa.dev/spec/v0.1/requirements#two-person-reviewed

According to SLSA it seems like PRs submitted by an Approver would only need 1 additional approval. However, I don't think this is something that could be enforced with GitHub branch protection rules. So better to just require 2 approvals for everybody.

ameowlia:
I agree with @Benjamintf1. 2 approvers sounds like _A LOT_, especially given that 95% of our PRs are written _by_ working group approvers, but when you open a PR you cannot approve your own PR. This means 3 people would be required for each PR (1 writer + 2 approvers).

In ARP we do not have any problems that could be solved by multiple approvers. Before we require multiple approvers I would want to hear a compelling reason why it is necessary.

### Example 1
* ARP Diego has 8 approvers
* Two are currently on parental leave
* If two approvers are required per PR...
  * Approver-1 writes the PR
  * Approver-2 approves the PR
  * Approver-3 approvers and merges the PR
* ⚠️ This means 50% of our current approvers who aren't on leave will be involved in every change that the team wants to make. 


### Example 2
* ARP Docs has 2 approvers
* If one approver is required per PR...
   * Approver-1 writes the PR
   * Approver-2 approves the PR
* This means _every_ PR this team makes will require participation from both approvers.
* ⚠️ This means when one member is on vacation, the other cannot get work in (without reaching out to me as the admin).

### Admin as workround is not sustainable
As the only admin for the ARP repos I have already been pinged 20+ times in the past week to fix things on an emergency basis. I worry that if approvers are required then when there is a time crunch I will be paged more often to use my admin powers.

### My request
**Make the number of approvers required configurable.** Small teams (like the docs teams above) shouldn't be forced to have _any_ approvers. And I see no reason why other teams should require more than one approver.

Gerg:
I oppose requiring more than one PR review. IMO, [diffusion of responsibility](https://en.wikipedia.org/wiki/Diffusion_of_responsibility) reduces the quality of multi-review PRs.

jrussett:
I don't think that adding another reviewer will improve the quality of the PR and will instead just increase the time it takes to get work done. It feels like we struggle enough to make sure that at least one person reviews the PR after another person has already done the acceptance, why add yet another layer of review and spent time?

AP-Hunt:
I've updated the RFC in response to feedback and discussion among the TOC.

Changes:
* Only bots will be allowed to push directly to the branch
* Humans must use a PR workflow
* WG with <=3 approvers will require 0 approvals, WG with >3 will require 1 approval.

I'll squash the commits before it gets merged.

rkoster (approved):

AP-Hunt:
@cloudfoundry/wg-leads Final comment period on this PR starts today, ends 15th November.

beyhan (approved):

pivotal-marcela-campo:
I am not comfortable with the problems it may cause to enforce this. The working group definition doesn’t have a calendar attached and we are not removing approvers each time someone takes time off or pauses their work on open source projects. 

Example:
Cloud Service Brokers team in Service management WG currently has 5 approvers

By the end of this month, one is going on leave, another one on holidays and we have a member on rotation on another team. This means we are left with 2 active approvers temporarily and hence wouldn’t require approvals, but we do because the WG definition doesn’t have a calendar attached. If one of them gets ill, or decides to take a day off then no work can be possibly merged. 

I do see the recommendations valuable and definitely think we should all follow whenever possible.

stephanme (approved):
+1 for the concept.

A minor changed needed for the version branches (remove one v).

Regarding the implementation: the branchprotector tool needs some enhancements before we can use it. At the moment it doesn't allow to configure teams that can bypass the PR flow. This is needed for bots.

toc/rfc/rfc-draft-branch-protection.md
```diff
@@ -0,0 +1,37 @@
...
+- Name: Default git branch protection rules
+- Start Date: 2022-10-07
+- Author(s): AP-Hunt, Benjamintf1, rkoster
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: https://github.com/cloudfoundry/community/pull/449
+
+
+## Summary
+
+In Q3 2022, the technical oversight committe accepted and implemented [an RFC for the managemnet of Github teams and user access](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0014-github-teams-and-access.md). This unintentionally broke the workflow and branch protection roles of some working groups.
```
emalm, line 12: Fixing some typos:
```suggestion
In Q3 2022, the Technical Oversight Committee accepted and implemented [an RFC for the management of GitHub teams and user access](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0014-github-teams-and-access.md). This unintentionally broke the workflow and branch protection rules of some working groups.
```

toc/rfc/rfc-draft-branch-protection.md
```diff
@@ -0,0 +1,37 @@
...
+
+This RFC seeks to address that by defining a baseline of branch protection rules across all repositories under management in the cloufoundry GitHub org.
```
emalm, line 14:
```suggestion
This RFC seeks to address that by defining a baseline of branch protection rules across all repositories under management in the cloudfoundry GitHub org.
```

toc/rfc/rfc-draft-branch-protection.md
```diff
@@ -0,0 +1,37 @@
...
+* allowing approvers in the working group to push to both release-line and non-release-line branches
+* allowing release bots to push commits to release-line branches
+
+When the TOC approved and implemented [RFC 0013](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0013-remove-nonstandard-github-teams.md), any GitHub teams within the cloudfoundry org that were **not** defined in working group charter were removed. This broke branch protection rules because they were now referring to non-existent teams.
+
+## Proposal
+
+The CFF TOC should implement basic branch protection rules for all repositories in an automated manner, and allow working groups to extend them as they see fit. This can be achieved, technically, with [the `branchprotector` tool from `prow` toolset](https://github.com/kubernetes/test-infra/blob/master/prow/cmd/branchprotector/README.md).
+
+We propose that protection be applied to all branches matching either the default branch or the expression v`v[0-9]*`.
```
emalm, line 29:
```suggestion
We propose that protection be applied to all branches matching either the default branch or the expression `v[0-9]*`.
```

emalm (changes requested):
In addition to these editorial corrections, let's discuss whether we should amend the RFC to include @stephanme's [idea](https://github.com/cloudfoundry/community/pull/449#issuecomment-1315399746) of allowing per-repository override of these default branch-protection rules.

toc/rfc/rfc-draft-branch-protection.md
```diff
@@ -0,0 +1,37 @@
...
+
+On protected branches, we propose the following rules:
+* only bots may push directly to the branch
```
emalm, line 32:
```suggestion
* only the bots for the Working Group owning the repository and any areas in the Working Group contributing to the repository may push directly to the branch
```

emalm (approved):

PR #453, opened 2022-10-12
Move the CF on K8s SIG call to Wednesday

gcapizzi:
We'd like to move our SIG call back to every other Wednesday. @christopherclark can you help me make sure that the calendar invite is synced with this PR?


ramiyengar:
I've moved the invite, @gcapizzi. You can check the CF Public calendar here: https://www.cloudfoundry.org/community-calendar/

PR #459, opened 2022-10-18
Add a docs area to the FI WG and add docs approvers

ameowlia:
## Problem
Technical writers need to be approvers to do their jobs. These three docs repos (docs-bbr, docs-credhub, and docs-uaa) are part of the documentation @cshollingsworth and @animatedmax maintain for Cloud Foundry.

## Solution
This fix adds a docs area to the Foundational Infrastructure WG. This fix follows the prior art set by the other working groups with docs repos[1] for how to give technical writers access.

[1] App Runtime Interfaces and App Runtime Platform

## Approvers
There is very little activity on these repos. Because of their low volume I do not think the approvers for them can be held to standard as other areas; if we did it would take years before anyone could become an approver. Below I break down how these docs writers have done the majority of work in these repos and why they should be approvers.

**[Docs-UAA](https://github.com/cloudfoundry/docs-uaa)**
* Since Jan 2021 there have been 12 commits
* 7/12 (58%) by @cshollingsworth
* 3/12 (25%) by @animatedmax

**[Docs-Credhub](https://github.com/cloudfoundry/docs-credhub/commits/master)**
* Since Jan 2020 there have been 19 commits
* 2/19 (10%) by @animatedmax

Additionally, both of these approvers have long tenure as CF docs writers, @cshollingsworth for 5 years and @animatedmax for 8 years.

### DIFF
--- a/toc/working-groups/foundational-infrastructure.md
+++ b/toc/working-groups/foundational-infrastructure.md
@@ -54,6 +54,16 @@ bots:
 - name: Cryogenics-CI
   github: Cryogenics-CI
 areas:
+- name: Docs
+  approvers:
+  - name: Chloe Hollingsworth
+    github: cshollingsworth
+  - name: Max Hufnagel
+    github: animatedmax
+  repositories:
+  - cloudfoundry/docs-bbr
+  - cloudfoundry/docs-credhub
+  - cloudfoundry/docs-uaa
 - name: Credential Management (Credhub)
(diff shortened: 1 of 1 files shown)


ameowlia:
As this is only pertaining to the FI working group, @rkoster (WG lead) has the authority to review and merge this without waiting on the TOC.

beyhan (approved):

PR #495, opened 2022-12-14
Add information about the ARI WG

beyhan:
Hi @Gerg,

I added information about the ARI slack channel but I wasn't sure about the rest. It will be great to have all the information here.


beyhan:
Closing because of missing review. @Gerg if you think this is useful pls reopen.

PR #541, opened 2023-02-21
Move stack-auditor to CLI team

ryanmoran:
The Buildpacks team does not feel that it can successfully maintain the the `stack-auditor` project. It was a repository that was historically owned by this team, but should probably be maintained elsewhere. I've suggested a move to the CLI team, but I'm open to other places that might make sense.

cc/ @cloudfoundry/wg-app-runtime-interfaces-cli-approvers

### DIFF
--- a/toc/working-groups/app-runtime-interfaces.md
+++ b/toc/working-groups/app-runtime-interfaces.md
@@ -147,7 +147,6 @@ areas:
   - cloudfoundry/libbuildpack
   - cloudfoundry/pip-pop
   - cloudfoundry/public-buildpacks-ci-robots
-  - cloudfoundry/stack-auditor
   - cloudfoundry/switchblade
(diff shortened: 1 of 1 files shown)


Gerg:
I created an interim PR to share the stack-auditor repo with areas that may be the long-term owners: https://github.com/cloudfoundry/community/pull/553

We can continue to use this PR to discuss the long-term plan.

beyhan:
@Gerg, @ryanmoran do you have better idea now what could be the long-term solution for the `stack-auditor` plugin or is this clarified already?

Gerg:
This PR has drifted some from the current ARI working group areas, and it was partially addressed by https://github.com/cloudfoundry/community/pull/553. 

I'm guessing we will still want to move stack-auditor out of the buildpacks area, but I'll open a new PR for that.

PR #554, opened 2023-03-08
Add kohara88 to CF contributors

kohara88:
I am a new docs writer from VMware and am requesting to become a CF contributor. @ameowlia


ameowlia (approved):
LGTM

PR #563, opened 2023-03-24
Add ard-wg-gitbot

jochenehret:


beyhan (changes requested):
Hi @jochenehret,

could you please also add the bot to the https://github.com/cloudfoundry/community/blob/main/org/contributors.yml otherwise the automation won't work.

beyhan (approved):

rkoster (approved):

PR #569, opened 2023-04-03
as we cannot reuse the relint-env repo.

ramonskie:
we need this repo so that we can store bbl state of the bosh-bootloader pipeline information in this repo
see https://github.com/cloudfoundry/bosh-bootloader/blob/main/ci/pipelines/bosh-bootloader.yml#L82

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -191,6 +191,13 @@ orgs:
           of your choice.
         has_projects: false
         has_wiki: false
+      bosh-bootloader-ci-envs:
+        allow_merge_commit: false
+        allow_squash_merge: false
+        default_branch: main
+        description: state files for bbl ci environments
+        has_projects: false
+        has_wiki: false
       bosh-cli:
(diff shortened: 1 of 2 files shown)


org/cloudfoundry.yml
```diff
@@ -191,6 +191,13 @@ orgs:
           of your choice.
         has_projects: false
         has_wiki: false
+      bosh-bootloader-ci-envs:
+        allow_merge_commit: false
+        allow_squash_merge: false
+        default_branch: main
+        description: state files for bbl ci environments
+        has_projects: false
+        has_wiki: false
```
rkoster, line 200: Shouldn't this repo be private?

ramonskie: yup

rkoster (approved):

beyhan (approved):

beyhan:
I was too fast with merging this but if you have any concerns we can revert the change.

PR #573, opened 2023-04-06
RFC for Python and Ruby removal from the CF stack

beyhan:
This RFC is the result of the discussions in the TOC meeting ([meeting notes](https://docs.google.com/document/d/1ng8rAcnrCbTm5RVNAKZ_ptqXiwcBIPgC8rbYp6A8lUo)) on 2023-04-04. It proposes a depracation period for the Removal of Python and Ruby from the CF stack

### DIFF
--- a/toc/rfc/rfc-0016-python-and-ruby-removal-from-cf-stack.md
+++ b/toc/rfc/rfc-0016-python-and-ruby-removal-from-cf-stack.md
@@ -4,7 +4,7 @@
 - Start Date: 2023-04-05
 - Author(s): @stephanme @sklevenz @beyhan
 - Status: Draft
-- RFC Pull Request: (fill in with PR link after you submit it)
+- RFC Pull Request: [community#573](https://github.com/cloudfoundry/community/pull/573)
 
 
 ## Summary
--- a/toc/rfc/rfc-0016-python-and-ruby-removal-from-cf-stack.md
+++ b/toc/rfc/rfc-0016-python-and-ruby-removal-from-cf-stack.md
@@ -1,0 +1,27 @@
+# Meta
+[meta]: #meta
+- Name: Proposal for Ruby and Python Removal from the cflinuxfs4 Stack with a Deprecation Period
+- Start Date: 2023-04-05
+- Author(s): @stephanme @sklevenz @beyhan
+- Status: Draft
+- RFC Pull Request: (fill in with PR link after you submit it)
+
+
+## Summary
+
+The Python and Ruby runtimes have been [removed from the cflinuxfs4 stack](https://github.com/cloudfoundry/cflinuxfs4/releases/tag/1.0.0) shortly before its GA release. This is a backwards incompatible change for the buildpacks relying on those runtimes and for some of the CF applications having runtime dependencies to Ruby or Python without using the Ruby or Python buildpack.
+
+This RFC addresses that by proposing a backwards compatible cflinuxfs4 stack and leaving time for the affected parties to adopt to the removal.
+
+## Problem
+
+The CF stack contains Ruby and Python versions coming from the Ubuntu version the stack is based on. The Ruby and the Python versions included in Ubuntu are quite old and usually have their end of life before the corresponding end of life of the Ubuntu version they are included in. The end of life of the CF stack is usually defined by the end of life of the Ubuntu version it’s based on. That means for certain time of period the CF stack is delivered with unmaintained Ruby and Python versions. This is a security issue and needs to be addressed but removing those runtimes from the stack is a backwards incompatible change because of:
+
+* Some of the available buildpacks’ implementations require Ruby or Python. That is why those buildpacks must adopt and bring their own Ruby or Python version, which means that there is a compatibility dependency between the `cflinuxfs4` stack and those buildpacks. This needs to be considered by CF application developers and we know from experience that such changes require a lot of time to be executed. There are even CF users pinning their buildpack versions which could be an issue in this case.
+
+* There are Cloud Foundry providers which already offer the new `cflinuxfs4` stack in `0.x` version to their end users and there are already applications running on that. CF applications could be impacted if there is a runtime dependency to Ruby or Python which is provided by the stack. Those applications will crash during the app evacuation happening as part of a CF update. This is an unsupported use case, but we shouldn't break those application without giving them a chance to adapt.
+
+## Proposal
+
+The CFF community should provide two versions of the `cflinuxfs4` stack one without Ruby and Python and one with those runtimes as backwards compatible version. The CFF community will deprecate the Python and Ruby provided by the `cflinuxfs4` stack and remove those runtimes from the stack latest with the `cflinuxfs5` stack. This should be part of the App Runtime Interfaces Working Group responsibilities because they own the process of making new CF stacks. Additionally, CF operators requires the option to choose between the two `cflinxufs4` stack flavours. This should be part of the App Runtime Deployments responsibilities because the working provides the refence deployment of CF.
+
--- a/toc/rfc/rfc-0016-python-and-ruby-removal-from-cf-stack.md
+++ b/toc/rfc/rfc-0016-python-and-ruby-removal-from-cf-stack.md
@@ -14,7 +14,7 @@ This RFC addresses that by proposing a backwards compatible `cflinuxfs4` stack a
 
 ## Problem
 
-The CF stack contains Ruby and Python versions coming from the Ubuntu version the stack is based on. The Ruby and the Python versions included in Ubuntu are quite old and usually have their end of life (EOF) before the corresponding EOF of the Ubuntu version they are included in. The EOF of the CF stack is usually defined by the EOF of the Ubuntu version it’s based on. That means for certain time of period the CF stack is delivered with unmaintained Ruby and Python versions. This is a security issue and needs to be addressed but removing those runtimes from the stack is a backwards incompatible change because of:
+The CF stack contains Ruby and Python versions coming from the Ubuntu version the stack is based on. The Ruby and the Python versions included in Ubuntu are quite old and usually have their end of life (EOL) before the corresponding EOL of the Ubuntu version they are included in. The EOL of the CF stack is usually defined by the EOL of the Ubuntu version it’s based on. That means for certain time of period the CF stack is delivered with unmaintained Ruby and Python versions. This is a security issue and needs to be addressed but removing those runtimes from the stack is a backwards incompatible change because of:
 
 * Some of the available buildpacks’ implementations require Ruby or Python. That is why those buildpacks must adopt and bring their own Ruby or Python version, which means that there is a compatibility dependency between the `cflinuxfs4` stack and those buildpacks. This needs to be considered by CF application developers and we know from experience that such changes require a lot of time to be executed. There are even CF users pinning their buildpack versions which could be an issue in this case.
 
@@ -22,7 +22,7 @@ The CF stack contains Ruby and Python versions coming from the Ubuntu version th
 
 ## Proposal
 
-The CFF community will provide two flavours of the `cflinuxfs4` stack one without Ruby and Python and one with those runtimes as backwards compatible option. The two flavours of `cflinuxfs4` will be supported until the EOF of `cflinuxfs4`. In the following sections the flavour without Ruby and Python is called `cflinuxfs4` and the one with `cflinuxfs4-compat`.
+The CFF community will provide two flavours of the `cflinuxfs4` stack one without Ruby and Python and one with those runtimes as backwards compatible option. The two flavours of `cflinuxfs4` will be supported until the EOL of `cflinuxfs4`. In the following sections the flavour without Ruby and Python is called `cflinuxfs4` and the one with `cflinuxfs4-compat`.
 
 ### Creation
 
--- a/toc/rfc/rfc-0016-python-and-ruby-removal-from-cf-stack.md
+++ b/toc/rfc/rfc-0016-python-and-ruby-removal-from-cf-stack.md
@@ -6,22 +6,38 @@
 - Status: Draft
 - RFC Pull Request: [community#573](https://github.com/cloudfoundry/community/pull/573)
 
-
 ## Summary
 
 The Python and Ruby runtimes have been [removed from the cflinuxfs4 stack](https://github.com/cloudfoundry/cflinuxfs4/releases/tag/1.0.0) shortly before its GA release. This is a backwards incompatible change for the buildpacks relying on those runtimes and for some of the CF applications having runtime dependencies to Ruby or Python without using the Ruby or Python buildpack.
 
-This RFC addresses that by proposing a backwards compatible cflinuxfs4 stack and leaving time for the affected parties to adopt to the removal.
+This RFC addresses that by proposing a backwards compatible `cflinuxfs4` stack and leaving time for the affected parties to adopt to the removal.
 
 ## Problem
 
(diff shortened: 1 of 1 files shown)


robdimsdale (commented):
In principle I am happy with supporting a version of `cflinuxfs4` that contains ruby and python. I have some specific questions that I don't currently see addressed in this RFC:

1. Which versions of ruby and python do we expect to see on the stack? Before we removed them, the versions were: Python 3.10.6 and Ruby 3.0.2p107. I'd assume we continue with the same versions
2. Should we continue to provide patch updates to Ruby and Python, similar to all other packages in the stack? I assume yes.
3. How long should we create and support this stack for?
4. What version should the stack be? I'd propose continuing the `v0.X.Z` version line. So we'd pick up at `v0.75.0`.
5. What's the process for deprecating ruby and python (or the entire version line)?
6. How do we communicate this deprecation to buildpack authors and app developers?

toc/rfc/rfc-0016-python-and-ruby-removal-from-cf-stack.md
```diff
@@ -0,0 +1,27 @@
...
+* There are Cloud Foundry providers which already offer the new `cflinuxfs4` stack in `0.x` version to their end users and there are already applications running on that. CF applications could be impacted if there is a runtime dependency to Ruby or Python which is provided by the stack. Those applications will crash during the app evacuation happening as part of a CF update. This is an unsupported use case, but we shouldn't break those application without giving them a chance to adapt.
+
+## Proposal
+
+The CFF community should provide two versions of the `cflinuxfs4` stack one without Ruby and Python and one with those runtimes as backwards compatible version. The CFF community will deprecate the Python and Ruby provided by the `cflinuxfs4` stack and remove those runtimes from the stack latest with the `cflinuxfs5` stack. This should be part of the App Runtime Interfaces Working Group responsibilities because they own the process of making new CF stacks. Additionally, CF operators requires the option to choose between the two `cflinxufs4` stack flavours. This should be part of the App Runtime Deployments responsibilities because the working provides the refence deployment of CF.
```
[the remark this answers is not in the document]
beyhan: IMHO the best time to remove packages from the CF stack is the release of a new one. Unfortunately, we are a little bit too late with the current one. Additionally, the CF platform in general is a multi purpose PaaS hosting apps with many unknown use case. That is why, we need to be careful when we do incompatible changes in the platform.

beyhan:
Hi @robdimsdale,

thank you for the really good questions. Here the answers: 
1. We think that the same version of Ruby and Python which are coming with Ubuntu makes sense here
2. Yes, they should be treated like any other package coming with the stack
3. A good experience for the community would be to support it until the end of life of `cflinuxfs4` and remove with the next stack version.
4. Would it be possible to have the same version but with a qualifier like `1.x.y-with-deprecated-packages`?
5. We should deprecate those packages with the `cflinuxfs4` and remove with the next stack version
6. We suggest to use the CFF channels which are appropriate to communicate the deprecation and the timelines when we have them.

Most probably it makes sense to also have a deprecation RFC. I will bring up this question in todays TOC meeting.

AP-Hunt:
How much effort is it to maintain a compatibility version like that? Is it a thing that could be done throughout the lifetime of 'fs4? Or only for so long?

robdimsdale:
> 1. We think that the same version of Ruby and Python which are coming with Ubuntu makes sense here

Great, we can include whatever versions of python 3.x and ruby 3.x are included in Jammy. This will be a breaking change from cflinuxfs3 as that included ruby 2.x and python 2.x (as well as python 3.x), but I think that's reasonable.

> 2. Yes, they should be treated like any other package coming with the stack

Agreed - just wanted to make sure we're on the same page on this point.

> 3. A good experience for the community would be to support it until the end of life of cflinuxfs4 and remove with the next stack version.
> 5 . We should deprecate those packages with the cflinuxfs4 and remove with the next stack version

I think this is where we have different perspectives. We're going to be supporting `cflinuxfs4` until 2026, and I don't love the idea of supporting two parallel versions of a stack for that long. Could we start with a shorter deprecation window? Something long enough to enable users to migrate off, but not as long as four years? Perhaps one year? I'd be happy with language in the RFC that allows us to extend this window if needed.

> 4. Would it be possible to have the same version but with a qualifier like 1.x.y-with-deprecated-packages?

I think it's more accurate (and intuitive) to keep the version line as `0.x.y`. We're essentially continuing the `0.x.y` version line that existed before we made the breaking changes in the 1.0 release. I accept that `v0.x` is a signal that there could be more breaking changes to come - if one was to interpret the version lines as semver - but I don't really think the stacks truly follow semver anyway. Therefore I think we can choose to frame the versions in terms of the actual supported set of packages.

I'm open to an alternative versioning scheme (e.g. `1.x.y-deprecated-packages`), but I'd really like to start with `0.x.y`.

> 7. We suggest to use the CFF channels which are appropriate to communicate the deprecation and the timelines when we have them.

Sure, no objections. I'm happy with this. What I'd like to avoid is adding logic to the stack/buildpacks to issue deprecation notices during the build, but it doesn't like like there's a desire for this.

beyhan:
@robdimsdale thanks for your feedback. After the discussions in the TOC meeting today we went over the questions with @stephanme and our answers are:

3. We would like to have it for the whole life time of `cflinuxfs4` because:
3.1 We have many customers already running on `cflinuxfs4` (around 1/3 of our customers) and that is why we want to avoid the incompatible change here because the customers already put efforts to move to the current version of `cflinuxfs4` and it was a lot of effort for us to coordinate and organize this.
3.2 We don't see an easy option to provide the two `cflinuxfs4` stack flavours as opt-in for the offline buildpacks. This will require efforts also on the buildpacks to provide releases for the two flavours. 
3.3 The CF stack is quite hidden from CF end users and an incompatible change in the stack will break their apps during the CF update without giving customers a chance to validate and avoid this. 
3.4 We were wondering in the TOC meeting what are the efforts to support the two flavours of the stacks. Could you please elaborate on this? When it is about validation the ARD WG could help here.
4. We think that from consumption point of view it will be easier for `cf-deployment` e.g. to consume the both flavours like it works with the buildpacks providing a `cflinuxfs3` and `cflinuxfs4` package in the same bosh-release (example: https://bosh.io/releases/github.com/cloudfoundry/java-buildpack-release?version=4.56)

robdimsdale:
Gotcha. I see that. It is unfortunate that you migrated customers to cflinuxfs4 before we'd finalized the stack, validated the buildpack compatibility, and moved it out of the `experimental` ops-file section cf-deployment.

However, I accept that it's equally unfortunate that we didn't broadly communicate our intention to make breaking changes to the stack. Ideally we could have communicated our decision earlier, before you'd started migrating, and you would have held off migrating customers until the stack was ready.

Regardless, I accept that this is the situation we find ourselves in so we have to find a compromise.

My main concern with supporting the stack is the engineering overhead of the pipeline. The longer we support it for, the longer we incur the burden. A secondary concern is the cognitive overhead of maintaining two version lines. A minor concern is financial cost of running another pipeline†. However I am overwhelmingly concerned about the first concern - the pipeline overhead.

You can see the current `cflinuxfs4` pipeline [here](https://buildpacks.ci.cf-app.com/teams/core-deps/pipelines/cflinuxfs4). We'd essentially be duplicating all which increases the maintenance burden: increased chance of needing to debug broken builds, increased surface area for future refactors/modifications.

Perhaps a compromise is that we support a simpler pipeline for the version of the stack that contains ruby and python? Maybe we can run a simpler pipeline - maybe one that doesn't run CATS?

How would you feel about this proposal:
1. The buildpacks team continues to support cflinuxfs4 without Ruby and Python in the current pipeline - running the current set of tests/validation, etc
1. The buildpacks team adds a simpler pipeline to support cflinuxfs4 with Ruby and Python (version naming TBD). This pipeline would build and publish the stack without running the full set of tests.
    * For increased confidence, we could choose to only build this stack after we have published the (fully-validated) stack without Ruby and Python. This would avoid a situation where we publish a stack with Ruby and Python when the stack without those packages has failed the extra validation.
1. The buildpacks team will continue to publish both stacks via the above mechanism until the EOL for cflinuxfs4 (currently slated for 2027/04/01)

As far as the the version for this stack (`0.x.y` vs `1.x.y-compat` or similar), I'm not following what you're saying. I don't see how `0.x.y` is incompatible with the current distribution mechanisms you describe.

---

† Currently the CI for buildpacks is paid for by VMware, not the CFF. I think it would be a good idea to start a conversation about migrating to CFF-hosted infrastructure, but I don't want to muddy this RFC with that concern.

jochenehret:
From [cf-deployment](https://github.com/cloudfoundry/cf-deployment) point of view it would be easier to have separate BOSH releases for the community cflinuxfs4 stack and the compatibility stack. The [update](https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/update-releases) and [validation](https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment) pipelines can only handle one version of a BOSH release. So we could only process the 1.x.x community release and the 0.x.x compatibility release would have to be validated elsewhere. I would expect however that building separate BOSH releases would be more effort for the stack maintainers.

We'll try to detail out these two solutions and possibly add a few more feasible approaches.

beyhan:
@robdimsdale thanks for looking into this in details and the proposal! We agree that doubling the CATS for the stack validation for such a long period is not a good effort & cost against value investment. Maybe there is a possibility to execute the CATS against the compatibility stack on the ARD WG side without generating further costs. As @jochenehret commented the ARD WG will look into details how the two stacks could be consumed and come up with more details.

jochenehret:
Here's a start to technically detail out the solutions discussed above. @robdimsdale : Please fill in for the ARI WG / Buildpack WG. Any additional ideas are of course welcome.

| Solution | ARI WG Implementation | ARI WG Effort | ARD WG Implementation | ARD WG Effort | 
| - | - | - | - | - |
| Separate BOSH releases | ? | ? | Additional ops-release in [update-releases pipeline](https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/update-releases). New ops file to replace cflinuxfs4-release with the compatibility release. Deployment and validation e.g. in [fresh-luna](https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment?group=fresh-luna). | Moderate |
| One BOSH release with different versioning schemes (0.x.x for compatibility release / 1.x.x for regular release) | ? | ? | Would have to use different [bosh-io-release-resource source configurations](https://github.com/concourse/bosh-io-release-resource#source-configuration) for the cflinuxfs4 bosh.io resource. Does not fit with the existing solutions in the [update-releases pipeline](https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/update-releases). Automatic update of compatibility release is not easily possible. | High |
| One BOSH release, one version, different packages (not sure if this works, needs a PoC) | ? | ? | No changes in [update-releases pipeline](https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/update-releases) necessary. Would need a new ops file to deploy the compatibility package from cflinuxfs4-release. Could be deployed and validated in one of the existing scenarios, e.g. [fresh-luna](https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment?group=fresh-luna) | Moderate |

robdimsdale:
Solution | ARI WG Implementation | ARI WG Effort | ARD WG Implementation | ARD WG Effort
-- | -- | -- | -- | --
Separate BOSH releases | Separate repository for BOSH release (e.g. `cloudfoundry/cflinuxfs4-compat-release`) - results in clear intent and discoverability | Low (especially if there's minimal testing before release) | Additional ops-release in update-releases pipeline. New ops file to replace cflinuxfs4-release with the compatibility release. Deployment and validation e.g. in fresh-luna. | Moderate
One BOSH release with different versioning schemes (0.x.x for compatibility release / 1.x.x for regular release) |  Interleave the BOSH final releases in the single, existing repository (i.e. `cloudfoundry/cf-linuxfs4-release`) | Low/Moderate | Would have to use different bosh-io-release-resource source configurations for the cflinuxfs4 bosh.io resource. Does not fit with the existing solutions in the update-releases pipeline. Automatic update of compatibility release is not easily possible. | High
One BOSH release, one version, different packages (not sure if this works, needs a PoC) | Interleave package (blob) updates in single, existing BOSH release repo | Moderate | No changes in update-releases pipeline necessary. Would need a new ops file to deploy the compatibility package from cflinuxfs4-release. Could be deployed and validated in one of the existing scenarios, e.g. fresh-luna | Moderate

robdimsdale:
@jochenehret @beyhan Based on the above suggestions, it seems like in aggregate, the easiest path for everyone is a separate BOSH release (GitHub repository, BOSH.io entry, etc) for the release that contains ruby + python.

I'm happy with the naming `cflinuxfs4-compat-release`, but I'm also open to alternative names.

As far as the version numbering, if it's a separate repo it doesn't really matter what version line we choose. I'd probably propose v1.x to indicate the level of support and maturity. We won't be able to keep the exact same version as the non-compat release because they will drift over time - i.e. as soon as there's a patch to ruby/python or a transitive dependency the compat-release will move ahead relative to the non compat release - so I'd propose starting at v1.0.0 with the first final release.

In terms of testing, if we are comfortable moving the CATS validation to post-release, that would result in a fairly straight forward process for the buildpacks team / ARI Working Group: we would trigger the creation and automatic publish of the compat stack when the non-compat stack is published. We could probably do that entirely within GitHub actions, which would be effectively free, as well as easy to maintain.

robdimsdale:
Great summary @beyhan !

The only thing I'd add is a bit more detail about why/how we can keep the two versions of cflinuxfs4 in sync: we agreed in the meeting that when we build the `compat` stack (based off the newly-published stack without ruby and python) we would build with whatever versions of ruby and python were available at that time. We explicitly agreed that we wouldn't try to build a new `compat` stack if/when there is a new version of ruby and/or python.

The benefits to this approach are:
* It keeps the version lines in sync, reducing complexity
* It is slightly easier to automate on the creation side

The cost of this approach is that if there is a CVE in ruby/python, or any of their transitive dependencies, that won't be resolved until the next time the upstream stack is created and the `compat` stack is rebuilt.

beyhan:
@robdimsdale thanks for the addition! I added it to the RFC. Regarding adding bosh-releases on bosh.io please check https://github.com/bosh-io/releases and this is the [pr](https://github.com/bosh-io/releases/pull/106) which was open for `cflinuxfs4`.

@cloudfoundry/toc  I have update the pr with all the discussions we had. It is now ready for review.

stephanme (approved):

AP-Hunt (approved):

rkoster (approved):

toc/rfc/rfc-0016-python-and-ruby-removal-from-cf-stack.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Summary
+
+The Python and Ruby runtimes have been [removed from the cflinuxfs4 stack](https://github.com/cloudfoundry/cflinuxfs4/releases/tag/1.0.0) shortly before its GA release. This is a backwards incompatible change for the buildpacks relying on those runtimes and for some of the CF applications having runtime dependencies to Ruby or Python without using the Ruby or Python buildpack.
+
+This RFC addresses that by proposing a backwards compatible `cflinuxfs4` stack and leaving time for the affected parties to adopt to the removal.
+
+## Problem
+
+The CF stack contains Ruby and Python versions coming from the Ubuntu version the stack is based on. The Ruby and the Python versions included in Ubuntu are quite old and usually have their end of life (EOF) before the corresponding EOF of the Ubuntu version they are included in. The EOF of the CF stack is usually defined by the EOF of the Ubuntu version it’s based on. That means for certain time of period the CF stack is delivered with unmaintained Ruby and Python versions. This is a security issue and needs to be addressed but removing those runtimes from the stack is a backwards incompatible change because of:
```
Gerg, line 17: Should `EOF` be `EOL`?

Gerg (approved):

jochenehret (approved):

toc/rfc/rfc-0016-python-and-ruby-removal-from-cf-stack.md
beyhan: Good catch :-). I fixed it. Thanks

beyhan:
@robdimsdale would it be possible to review until next Tuesday so that we can proceed with this during the next TOC meeting. You've been so active on this that is why I would like to have your review also.

robdimsdale (approved):

PR #587, opened 2023-04-21
Add routing-ci repository

plowin:
Hi @ameowlia, as discussed in our last WG meeting, we would like to create a new repo for "routing-ci", similar to e.g. "runtime-ci". It seems this repo existed in the past and I am not sure if we could revive it like this. I am also fine with moving it to the `Networking-Extensions` ~line 363 if you prefer.

We will probably not store the CI pipelines in this repo as they better stay close with the respective repo they are used in. But we would create/host the VM running these pipeline in this repo. Please review/add your thoughts.

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -1927,6 +1927,10 @@ orgs:
       routing-ci:
         archived: true
         has_projects: true
+      routing-concourse:
+        default_branch: main
+        description: Concourse CI for CF Routing Components
+        has_projects: true
       routing-datadog-config:
         has_projects: true
         private: true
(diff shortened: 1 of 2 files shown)


ameowlia (changes requested):
@plowin Unless you want to unarchive routing-ci and use the contents in it, can you come up with a new name for the repo?

plowin:
Thanks for the pointer!
I missed that this repo still exists, proposed a new name, @ameowlia

ameowlia (approved):
🎉

plowin:
In the next weeks we are prioritizing bringing a concourse for ARP WG up in this repo

PR #591, opened 2023-05-01
RFC to add Cloud Native Buildpacks to CF Deployment

Gerg:
For easier viewing: https://github.com/cloudfoundry/community/blob/rfc_cnbs/toc/rfc/rfc-draft-add-cnbs.md

cc @ryanmoran, @robdimsdale, @dsabeti, @aramprice, @dsboulder

### DIFF
--- a/toc/rfc/rfc-draft-add-cnbs.md
+++ b/toc/rfc/rfc-draft-add-cnbs.md
@@ -152,7 +152,7 @@ A `cnb-shim-builder` repository will be added to the Application Runtime
 Interfaces working group. This repository will contain the tooling necessary
 for converting existing Paketo buildpacks into shimmed CNBs.
 
-The shimmed buildpacks will be experimental bosh releases, belonging to the
+The shimmed buildpacks will be experimental BOSH releases, belonging to the
 Application Runtime Interfaces working group. The proposed list of buildpacks
 is as follows (in rough priority order):
 1. `java_native_cnb_release`
@@ -171,8 +171,8 @@ is as follows (in rough priority order):
 It is not required to shim all buildpacks during phase 1 of this project.
 Depending on the effort to shim buildpacks and other factors, it may make sense
 to move to native CNB support before all Paketo buildpacks are shimmed. During
-phase 2, the above bosh release repositories will be converted to unshimmed
-bosh releases of the corresponding Paketo buildpacks.
+phase 2, the above BOSH release repositories will be converted to unshimmed
+BOSH releases of the corresponding Paketo buildpacks.
 
 The CNBs will initially be opt-in for CF Deployment operators via an
 experimental ops file. Once the buildpacks have reached a suitable level of
@@ -186,7 +186,7 @@ ease development and adoption. Paketo provides [multiple stacks](https://paketo.
 that are compatible with the Paketo buildpacks. There is an opportunity to
 adopt some or all of the Paketo stacks into CF Deployment, in addition to the
 buildpacks. The Paketo buildpacks have greater cohesion with the Paketo
-buildpacks than with `cflinuxfs4`, and the Paketo "base" and "tiny" stacks could
+stacks than with `cflinuxfs4`, and the Paketo "base" and "tiny" stacks could
 offer security-conscious CF Deployment users stacks with fewer native
 dependencies than are currently included with `cflinuxfs4`. This RFC does not
 cover the adoption of additional stacks into CF Deployment, but it does open
--- a/toc/rfc/rfc-draft-add-cnbs.md
+++ b/toc/rfc/rfc-draft-add-cnbs.md
@@ -13,11 +13,10 @@
 buildpacks, are the current generation of buildpacks, and offer some
 improvements over the v2 buildpacks that CF Deployment currently uses. The
 Cloud Foundry Foundation already has an implementation of Cloud Native
-Buildpacks via the [Korifi](https://www.cloudfoundry.org/technology/korifi/)
-project, however these CNBs are not currently integrated with core CF
-Deployment.
+Buildpacks via the [Paketo](https://paketo.io/) project, however these CNBs are
+not currently integrated with core CF Deployment.
 
-This RFC proposes adding CNB support to Cloud Foundry and including the Korifi
+This RFC proposes adding CNB support to Cloud Foundry and including the Paketo
 buildpacks in CF Deployment.
 
 ## Problem
@@ -65,7 +64,8 @@ Cloud Foundry with OCI images, container registries, etc.
     1. [Java Native Buildpack](https://github.com/paketo-buildpacks/java-native-image)
     1. Easier buildpack customization/composition
 1. Increase cohesion and application portability between CF Deployment and
-   Korifi, via mutual Paketo integration
+   [Korifi](https://www.cloudfoundry.org/technology/korifi/), via mutual Paketo
+   integration
 1. Increase user base for the CNB lifecycle and Paketo buildpacks
 1. Open the door for eventual deprecation of the v2 buildpacks, reducing
    maintenance costs (v2 buildpack deprecation is NOT included in this RFC)
--- a/toc/rfc/rfc-draft-add-cnbs.md
+++ b/toc/rfc/rfc-draft-add-cnbs.md
@@ -1,0 +1,194 @@
+# Meta
+[meta]: #meta
+- Name: Add Cloud Native Buildpacks
+- Start Date: 2023-05-01
+- Author: @gerg
+- Contributing Authors: @ryanmoran, @robdimsdale, @dsabeti, @aramprice, @dsboulder
+- Status: Draft
+- RFC Pull Request: #591
+
+## Summary
+
+[Cloud Native Buildpacks (CNBs)](https://buildpacks.io/), also known as v3
+buildpacks, are the current generation of buildpacks, and offer some
+improvements over the v2 buildpacks that CF Deployment currently uses. The
+Cloud Foundry Foundation already has an implementation of Cloud Native
+Buildpacks via the [Korifi](https://www.cloudfoundry.org/technology/korifi/)
+project, however these CNBs are not currently integrated with core CF
+Deployment.
+
+This RFC proposes adding CNB support to Cloud Foundry and including the Korifi
+buildpacks in CF Deployment.
+
+## Problem
+
+The v2 buildpacks are effectively in maintenance mode and do not receive
+substantial new features. By not integrating with v3 buildpacks, Cloud Foundry
+is missing out on new capabilities like application
+[SBOM](https://en.wikipedia.org/wiki/Software_supply_chain) generation, new
+buildpacks like the Java Native and Web Servers CNBs, as well as any new
+features that are added to the still-actively-developed v3 buildpacks.
+
+## Proposal
+
+### Overview
+
+Cloud Foundry integration with CNBs will be executed over at least two phases.
+
+In the first phase, the Paketo buildpacks will be "shimmed" to work with Cloud
+Foundry's existing v2 buildpack interface. These shimmed Paketo buildpacks will
+be bundled with select [CNB lifecycle](https://github.com/buildpacks/lifecycle)
+binaries and lightweight orchestration binaries. At build time, the Cloud
+Foundry [buildpack lifecycle](https://github.com/cloudfoundry/buildpackapplifecycle)
+will invoke the orchestration binaries, which will map the v2 build process to
+the v3 build process.
+
+In the second phase, unmodified Paketo buildpacks will be natively supported by
+Cloud Foundry. This will be implemented by integrating the buildpack shim logic
+into a Cloud Foundry lifecycle.
+
+Depending on learnings from the first two phases of CNB integration or other
+factors, there may be additional phases of work, however these additional
+phases are not covered by this RFC.
+
+Notably, this proposal does NOT include building or running OCI images on Cloud
+Foundry. This proposal is to use Paketo buildpacks to build and run Cloud
+Foundry droplets. This proposal does not preclude future efforts to integrate
+Cloud Foundry with OCI images, container registries, etc.
+
+### Goals
+
+1. Align CF Deployment with the main focus of the Buildpacks team
+1. Bring new build capabilities to CF Deployment users, including:
+    1. [Application SBOMs](https://paketo.io/docs/howto/sbom/)
+    1. [Web Servers Buildpack](https://github.com/paketo-buildpacks/web-servers)
+    1. [Java Native Buildpack](https://github.com/paketo-buildpacks/java-native-image)
+    1. Easier buildpack customization/composition
+1. Increase cohesion and application portability between CF Deployment and
+   Korifi, via mutual Paketo integration
+1. Increase user base for the CNB lifecycle and Paketo buildpacks
+1. Open the door for eventual deprecation of the v2 buildpacks, reducing
+   maintenance costs (v2 buildpack deprecation is NOT included in this RFC)
+
+### High Level Implementation
+
+In phase one, Cloud Foundry will produce a series of experimental shimmed
+buildpacks that include:
+1. A Paketo CNB
+1. Select executables from the CNB lifecycle
+1. Orchestrator executables that conform to the v2 buildpack interface
+
+Cloud Foundry users will interact with the shimmed buildpacks the same as they
+do for any other buildpack. For example, the following commands will work as
+expected:
+
+```
+$ cf create-buildpack ruby_cnb ./ruby_cnb_shim.zip
+...
+$ cf push my-app --buildpack=ruby_cnb
+```
+
+At build time, the procedure will be as follows:
+1. The Cloud Foundry Buildpack App Lifecycle's Builder will invoke shim
+   orchestrator executables, consistent with the v2 buildpack interface
+1. The shim orchestrator will invoke bundled CNB lifecycle executables
+1. The CNB lifecycle executables will invoke Paketo buildpack executables,
+   consistent with the v3 buildpack interface
+1. The Paketo buildpack will build a set of "layer" directories
+1. The shim orchestrator will package the layer directories into a Cloud
+   Foundry droplet and return it back to the Cloud Foundry Lifecycle
+
+For a proof of concept of shimmed buildpacks, see @ryanmoran's
+[cfnb](https://github.com/ryanmoran/cfnb) project. Developing the shim
+externally at first will allow for more rapid iteration, without disrupting the
+core Cloud Foundry runtime.
+
+During phase two, the shim logic will be integrated into the Cloud Foundry
+lifecycle itself, allowing for "native" Paketo buildpack support. The details
+of this integration will be decided once phase one is complete.
+
+### Shim at Build Time
+
+At build time, the shim orchestrator will expose the following v2 buildpack
+executables:
+
+| Executable | Function |
+| --- | --- |
+| `detect` | Invoke the CNB Lifecycle `detector` executable |
+| `supply` | Invoke the CNB Lifecycle `builder` executable |
+| `finalize` | Add CNB `launcher` executable and runtime configuration files to build layers |
+| `release` | Package the generated layer directories into a Cloud Foundry droplet |
+
+These executables will be invoked by the Cloud Foundry Buildpack App Lifecycle's
+builder.
+
+### Shim at Run Time
+
+At run time, the Cloud Foundry Buildpack App Lifecycle launcher will invoke the
+CNB Lifecycle `launcher` executable that was included in the droplet at build
+time. In order to support custom start commands for apps using the CNB
+buildpack shims, the CNB launcher invocation will be prepended to the app start
+command using the [`entrypoint_prefix` hook](https://github.com/cloudfoundry/buildpackapplifecycle/commit/29feb13caeff646f35585eee865c376c818fc2ea)
+in the CF Lifecycle launcher.
+
+### Service Bindings
+
+The largest anticipated breaking change from v2 to v3 buildpacks is the change
+in how service bindings are consumed by buildpacks. The v3 buildpacks expect
+service bindings to follow the [Kubernetes Service Binding specification](https://github.com/servicebinding/spec),
+which places service binding credentials on the filesystem. Cloud Foundry and
+v2 buildpacks instead use environment variables for service binding
+credentials.
+
+In order to securely place credentials on the filesystem, Diego will need the
+ability to mount in-memory filesystems (e.g. `tmpfs`) in application
+containers. Once an in-memory filesystem is available, select `VCAP_SERVICES`
+environment variables can be translated into their Kubernetes Service Binding
+equivalents by the shim orchestrator, for consumption by the v3 buildpacks.
+
+### Governance and Release
+
+A `cnb-shim-builder` repository will be added to the Application Runtime
+Interfaces working group. This repository will contain the tooling necessary
+for converting existing Paketo buildpacks into shimmed CNBs.
+
+The shimmed buildpacks will be experimental bosh releases, belonging to the
+Application Runtime Interfaces working group. The proposed list of buildpacks
+is as follows (in rough priority order):
+1. `java_native_cnb_release`
+1. `java_cnb_release`
+1. `web_servers_cnb_release`
+1. `dotnet_core_cnb_release`
+1. `nodejs_cnb_release`
+1. `python_cnb_release`
+1. `ruby_cnb_release`
+1. `procfile_cnb_release`
+1. `go_cnb_release`
+1. `nginx_cnb_release`
+1. `php_cnb_release`
+1. `rust_cnb_release`
+
+It is not required to shim all buildpacks during phase 1 of this project.
+Depending on the effort to shim buildpacks and other factors, it may make sense
+to move to native CNB support before all Paketo buildpacks are shimmed. During
+phase 2, the above bosh release repositories will be converted to unshimmed
+bosh releases of the corresponding Paketo buildpacks.
+
+The CNBs will initially be opt-in for CF Deployment operators via an
+experimental ops file. Once the buildpacks have reached a suitable level of
+maturity and stability, likely after phase 2, they will be added to the default
+set of buildpacks installed as part of CF Deployment.
+
+### Stacks
+
+Initially, the shimmed Paketo buildpacks will be compatible with `cflinuxfs4` to
+ease development and adoption. Paketo provides [multiple stacks](https://paketo.io/docs/concepts/stacks/#what-paketo-stacks-are-available)
+that are compatible with the Paketo buildpacks. There is an opportunity to
+adopt some or all of the Paketo stacks into CF Deployment, in addition to the
+buildpacks. The Paketo buildpacks have greater cohesion with the Paketo
+buildpacks than with `cflinuxfs4`, and the Paketo "base" and "tiny" stacks could
+offer security-conscious CF Deployment users stacks with fewer native
+dependencies than are currently included with `cflinuxfs4`. This RFC does not
+cover the adoption of additional stacks into CF Deployment, but it does open
+the door to add these stacks in the future.
+
--- a/toc/rfc/rfc-draft-add-cnbs.md
+++ b/toc/rfc/rfc-draft-add-cnbs.md
@@ -32,22 +32,19 @@ features that are added to the still-actively-developed v3 buildpacks.
 
 ### Overview
 
-Cloud Foundry integration with CNBs will be executed over at least two phases.
+Cloud Foundry integration with CNBs will be executed over multiple phases.
 
-In the first phase, the Paketo buildpacks will be "shimmed" to work with Cloud
-Foundry's existing v2 buildpack interface. These shimmed Paketo buildpacks will
-be bundled with select [CNB lifecycle](https://github.com/buildpacks/lifecycle)
-binaries and lightweight orchestration binaries. At build time, the Cloud
-Foundry [buildpack lifecycle](https://github.com/cloudfoundry/buildpackapplifecycle)
+In the first phase (the focus of this RFC), the Paketo buildpacks will be
+"shimmed" to work with Cloud Foundry's existing v2 buildpack interface. These
+shimmed Paketo buildpacks will be bundled with select
+[CNB lifecycle](https://github.com/buildpacks/lifecycle) binaries and
+lightweight orchestration binaries. At build time, the Cloud Foundry
+[buildpack lifecycle](https://github.com/cloudfoundry/buildpackapplifecycle)
 will invoke the orchestration binaries, which will map the v2 build process to
 the v3 build process.
 
-In the second phase, unmodified Paketo buildpacks will be natively supported by
-Cloud Foundry. This will be implemented by integrating the buildpack shim logic
-into a Cloud Foundry lifecycle.
-
-Depending on learnings from the first two phases of CNB integration or other
-factors, there may be additional phases of work, however these additional
+Based on learnings from the first phase of CNB integration and other
+factors, there will likely be additional phases of work, however these additional
 phases are not covered by this RFC.
 
 Notably, this proposal does NOT include building or running OCI images on Cloud
@@ -72,7 +69,7 @@ Cloud Foundry with OCI images, container registries, etc.
 
 ### High Level Implementation
 
-In phase one, Cloud Foundry will produce a series of experimental shimmed
+Cloud Foundry will produce a series of experimental shimmed
 buildpacks that include:
 1. A Paketo CNB
 1. Select executables from the CNB lifecycle
@@ -104,9 +101,6 @@ For a proof of concept of shimmed buildpacks, see @ryanmoran's
 externally at first will allow for more rapid iteration, without disrupting the
 core Cloud Foundry runtime.
 
-During phase two, the shim logic will be integrated into the Cloud Foundry
-lifecycle itself, allowing for "native" Paketo buildpack support. The details
-of this integration will be decided once phase one is complete.
 
 ### Shim at Build Time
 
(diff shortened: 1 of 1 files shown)


toc/rfc/rfc-draft-add-cnbs.md
```diff
@@ -0,0 +1,194 @@
...
+in how service bindings are consumed by buildpacks. The v3 buildpacks expect
+service bindings to follow the [Kubernetes Service Binding specification](https://github.com/servicebinding/spec),
+which places service binding credentials on the filesystem. Cloud Foundry and
+v2 buildpacks instead use environment variables for service binding
+credentials.
+
+In order to securely place credentials on the filesystem, Diego will need the
+ability to mount in-memory filesystems (e.g. `tmpfs`) in application
+containers. Once an in-memory filesystem is available, select `VCAP_SERVICES`
+environment variables can be translated into their Kubernetes Service Binding
```
emalm, line 146: It looks like the identifiers for types of services may already vary between the v2 and v3 buildpack lines: for example, the [Java buildpack support for Google Stackdriver Profiler](https://github.com/cloudfoundry/java-buildpack/blob/fe4be89177bede973cbaa14176e3aa5ff4315d69/docs/framework-google_stackdriver_profiler.md) looks for a service binding of name/type/label `google-stackdriver-profiler`, but the [Paketo Google Stackdriver CNB](https://github.com/paketo-buildpacks/google-stackdriver) looks for a binding of type `GoogleStackdriverProfiler`. How will app teams use the new buildpack integrations against existing service brokers that return the old identifiers? Will the translation of `VCAP_SERVICES` data into K8s service-binding file hierarchies also include a translation of known service identifiers to the corresponding Paketo-recognized type, or do we have some other proposal in mind?

Also, the K8s service-binding spec has only 3 canonical places for metadata: the required `type` and optional `provider` fields in the binding payload, and the name of the binding directory under the service-binding root directory. Have we given any consideration to how the `VCAP_SERVICES` metadata should map to these identifiers, and whether and how any additional service-binding metadata that CF buildpacks might inspect should map to conventional fields (presumably other files in the binding-specific directory)?

toc/rfc/rfc-draft-add-cnbs.md
```diff
@@ -0,0 +1,194 @@
...
+equivalents by the shim orchestrator, for consumption by the v3 buildpacks.
+
+### Governance and Release
+
+A `cnb-shim-builder` repository will be added to the Application Runtime
+Interfaces working group. This repository will contain the tooling necessary
+for converting existing Paketo buildpacks into shimmed CNBs.
+
+The shimmed buildpacks will be experimental bosh releases, belonging to the
```
emalm, line 155:
```suggestion
The shimmed buildpacks will be experimental BOSH releases, belonging to the
```

toc/rfc/rfc-draft-add-cnbs.md
```diff
@@ -0,0 +1,194 @@
...
+1. `go_cnb_release`
+1. `nginx_cnb_release`
+1. `php_cnb_release`
+1. `rust_cnb_release`
+
+It is not required to shim all buildpacks during phase 1 of this project.
+Depending on the effort to shim buildpacks and other factors, it may make sense
+to move to native CNB support before all Paketo buildpacks are shimmed. During
+phase 2, the above bosh release repositories will be converted to unshimmed
+bosh releases of the corresponding Paketo buildpacks.
```
emalm, line 175:
```suggestion
phase 2, the above BOSH release repositories will be converted to unshimmed
BOSH releases of the corresponding Paketo buildpacks.
```

emalm (changes requested):
Thanks! Very excited to see this proposal come forward. I suggested some corrections and had a few initial questions.

toc/rfc/rfc-draft-add-cnbs.md
```diff
@@ -0,0 +1,194 @@
...
+service bindings to follow the [Kubernetes Service Binding specification](https://github.com/servicebinding/spec),
+which places service binding credentials on the filesystem. Cloud Foundry and
+v2 buildpacks instead use environment variables for service binding
+credentials.
+
+In order to securely place credentials on the filesystem, Diego will need the
+ability to mount in-memory filesystems (e.g. `tmpfs`) in application
+containers. Once an in-memory filesystem is available, select `VCAP_SERVICES`
+environment variables can be translated into their Kubernetes Service Binding
+equivalents by the shim orchestrator, for consumption by the v3 buildpacks.
```
loewenstein, line 147: Maybe https://github.com/buildpacks/libcnb/pull/227 and https://github.com/buildpacks/libcnb/pull/228 are making a change here?

@emalm I did no real thorough search yet, but I think it is mainly apm buildpacks that need service bindings, I know for sure for the dynatrace-buildpack.

toc/rfc/rfc-draft-add-cnbs.md
```diff
@@ -0,0 +1,194 @@
...
+
+It is not required to shim all buildpacks during phase 1 of this project.
+Depending on the effort to shim buildpacks and other factors, it may make sense
+to move to native CNB support before all Paketo buildpacks are shimmed. During
+phase 2, the above bosh release repositories will be converted to unshimmed
+bosh releases of the corresponding Paketo buildpacks.
+
+The CNBs will initially be opt-in for CF Deployment operators via an
+experimental ops file. Once the buildpacks have reached a suitable level of
+maturity and stability, likely after phase 2, they will be added to the default
```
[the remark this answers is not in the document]
stephanme: How would the buildpack detection work when v2 and v3 buildpacks are installed both?

toc/rfc/rfc-draft-add-cnbs.md
```diff
@@ -0,0 +1,194 @@
...
+
+This RFC proposes adding CNB support to Cloud Foundry and including the Korifi
+buildpacks in CF Deployment.
+
+## Problem
+
+The v2 buildpacks are effectively in maintenance mode and do not receive
+substantial new features. By not integrating with v3 buildpacks, Cloud Foundry
+is missing out on new capabilities like application
+[SBOM](https://en.wikipedia.org/wiki/Software_supply_chain) generation, new
```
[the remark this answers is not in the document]
robdimsdale: Buildpacks produce SBOMs at known locations on the filesystem of the OCI image. It would probably be the responsibility of the shim layer to ensure these SBOM files are processed. I assume they wouldn't be packaged into the droplet, as that's just the executable, so they would have to provided to some Cloud Foundry handler via CF-specific metadata.

Maybe you knew all that already and your question was just around how CF will define, consume, and render this information. In which case I don't have any extra information :)

toc/rfc/rfc-draft-add-cnbs.md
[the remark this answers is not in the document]
robdimsdale: > How would the buildpack detection work when v2 and v3 buildpacks are installed both?

I'd expect them to be installed under different names, so you would specific them specifically with `cf push -b v3_python_buildpack`, or rely on the buildpack ordering for detect. However, given that they will generally detect on the same things (e.g. presence of language-toolchain-specific files) I would expect whichever is first in the detection order would likely win. For backwards-compatibility we would probably want to suggest operators put the v3 buildpacks lower in the order and ask users to explicitly opt-in via `cf push -b`.

toc/rfc/rfc-draft-add-cnbs.md
```diff
@@ -0,0 +1,194 @@
...
+CNB Lifecycle `launcher` executable that was included in the droplet at build
+time. In order to support custom start commands for apps using the CNB
+buildpack shims, the CNB launcher invocation will be prepended to the app start
+command using the [`entrypoint_prefix` hook](https://github.com/cloudfoundry/buildpackapplifecycle/commit/29feb13caeff646f35585eee865c376c818fc2ea)
+in the CF Lifecycle launcher.
+
+### Service Bindings
+
+The largest anticipated breaking change from v2 to v3 buildpacks is the change
+in how service bindings are consumed by buildpacks. The v3 buildpacks expect
```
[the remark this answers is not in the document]
robdimsdale: In general service bindings aren't used by most of the v3 buildpacks. They are predominantly used in the APM buildpacks (e.g. to provide credentials to third-part systems). Most of the top-level Paketo buildpacks (e.g. python, nodejs) don't ship with APM buildpack integration, but they could be modified to include those APM buildpacks if needed. In order to provide a similar surface area to the v2 buildpacks, they probably would have to do so.

toc/rfc/rfc-draft-add-cnbs.md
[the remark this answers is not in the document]
emalm: Thanks, @robdimsdale ! Actually, I've been unclear on the specifics of how the SBOM data gets propagated even in a K8s context, especially so it can integrate with system such as Grype that can process the SBOM data. Does the OCI image metadata contain the SBOM, or do those processing systems have to unpack the filesystem layers in the image to get it from a canonical location? If the former, does kpack then propagate the SBOM data from the filesystem into the image metadata?

Anyway, I definitely think we can work out all the details of exposing CNB-generated SBOM data to the authorized set of CF users in due time, but it'd be great to refer to it at a high level as intended work in this part of the proposal.

toc/rfc/rfc-draft-add-cnbs.md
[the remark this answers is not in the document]
emalm: Great, that makes sense to me about the distinct buildpack names and v2/v3 ordering for appropriate autodetection. We'll need to choose those names carefully so that we're all happy enough with them for the long term, since presumably any app manifests or `cf push` commands that select buildpacks by name will use those new names and it's generally a pain for app teams to change them later without a good reason.

toc/rfc/rfc-draft-add-cnbs.md
```diff
@@ -0,0 +1,194 @@
...
+
+[Cloud Native Buildpacks (CNBs)](https://buildpacks.io/), also known as v3
+buildpacks, are the current generation of buildpacks, and offer some
+improvements over the v2 buildpacks that CF Deployment currently uses. The
+Cloud Foundry Foundation already has an implementation of Cloud Native
+Buildpacks via the [Korifi](https://www.cloudfoundry.org/technology/korifi/)
+project, however these CNBs are not currently integrated with core CF
+Deployment.
+
+This RFC proposes adding CNB support to Cloud Foundry and including the Korifi
```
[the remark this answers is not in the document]
Gerg: Whoops! 😅

toc/rfc/rfc-draft-add-cnbs.md
tcdowney: Yeah that's correct. If the user does nothing then they will get credentials mounted at `${SERVICE_BINDING_ROOT}/<instance-guid>/` and the `type` value will be set to `user-provided` (I don't believe we default `provider` to anything). The user has the opportunity to override these by specifying `--binding-name` during binding creation (this will mount it at `${SERVICE_BINDING_ROOT}/<binding-name>/`) and by including the `type`/`provider` fields in the credential payload like you said.

For what it's worth, I don't think the `${SERVICE_BINDING_ROOT}/<instance-guid>/` structure is very useful and perhaps not even idiomatic. I think Cloud Controller has [sufficient validations](https://github.com/cloudfoundry/cloud_controller_ng/blob/8c7df76f686f9753591e9da3bd2786763d083fe7/app/models/services/service_binding.rb#LL44C1-L44C1) in place to ensure that binding name is unique at the app level (and that an instance can only be bound once) so it should mount them all at `${SERVICE_BINDING_ROOT}/<instance-name>/` / `${SERVICE_BINDING_ROOT}/<binding-name>/` by default. The GUID is too dynamic for an app to meaningfully make use of it here.

toc/rfc/rfc-draft-add-cnbs.md
```diff
@@ -0,0 +1,194 @@
...
+Foundry's existing v2 buildpack interface. These shimmed Paketo buildpacks will
+be bundled with select [CNB lifecycle](https://github.com/buildpacks/lifecycle)
+binaries and lightweight orchestration binaries. At build time, the Cloud
+Foundry [buildpack lifecycle](https://github.com/cloudfoundry/buildpackapplifecycle)
+will invoke the orchestration binaries, which will map the v2 build process to
+the v3 build process.
+
+In the second phase, unmodified Paketo buildpacks will be natively supported by
+Cloud Foundry. This will be implemented by integrating the buildpack shim logic
+into a Cloud Foundry lifecycle.
```
beyhan, line 47: This mean that there are no changes planned for the current V2 [buildpack lifecycle](https://github.com/cloudfoundry/buildpackapplifecycle). I think that we should mention it explicitly to avoid misunderstanding here.

toc/rfc/rfc-draft-add-cnbs.md
```diff
@@ -0,0 +1,194 @@
...
+Cloud Foundry with OCI images, container registries, etc.
+
+### Goals
+
+1. Align CF Deployment with the main focus of the Buildpacks team
+1. Bring new build capabilities to CF Deployment users, including:
+    1. [Application SBOMs](https://paketo.io/docs/howto/sbom/)
+    1. [Web Servers Buildpack](https://github.com/paketo-buildpacks/web-servers)
+    1. [Java Native Buildpack](https://github.com/paketo-buildpacks/java-native-image)
+    1. Easier buildpack customization/composition
```
beyhan, line 65: Is there now change required in CF to support this goal (Easier buildpack customization/composition)? In the RFC I couldn't find anything in this direction. Or this is covered by the [multiple-buildpacks support](https://docs.cloudfoundry.org/buildpacks/use-multiple-buildpacks.html)?

toc/rfc/rfc-draft-add-cnbs.md
```diff
@@ -0,0 +1,194 @@
...
+into a Cloud Foundry lifecycle.
+
+Depending on learnings from the first two phases of CNB integration or other
+factors, there may be additional phases of work, however these additional
+phases are not covered by this RFC.
+
+Notably, this proposal does NOT include building or running OCI images on Cloud
+Foundry. This proposal is to use Paketo buildpacks to build and run Cloud
+Foundry droplets. This proposal does not preclude future efforts to integrate
+Cloud Foundry with OCI images, container registries, etc.
```
loewenstein, line 56: I like incremental approaches in principle. One problem I see with this specific one is that we are heading ~180 degree in the other direction - if we wanted to follow up with pushing towards OCI images and container registries.
At least I don't see us reusing anything of the two phases laid out.

Gerg: From the system architecture perspective, this proposal is definitely divergent from building/running OCI images.  However, I think this proposal is convergent from the app developer perspective. My hypothesis is that app developers don't care that much about how their apps are packaged/run. With this proposal, app developers can get the benefits of the Paketo buildpacks, without requiring significant changes to the CF Deployment platform engineer/operator experience or system architecture.

toc/rfc/rfc-draft-add-cnbs.md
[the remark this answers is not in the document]
Gerg: @robdimsdale Can you expand more on why you think the SBOM shouldn't be in the droplet? There are some potential advantages to bundling the SBOM with the executable. For example:
- Benefiting people who are scanning droplets directly from the blobstore (though maybe this won't be as common if we have built-in SBOM support).
- Preserving the SBOM when downloading/uploading droplets.

10k-100k lines per SBOM is probably too large to put in the CCDB, so we would likely have to store the SBOMs in the blobstore anyway.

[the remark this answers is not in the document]
robdimsdale: Maybe my mental model is wrong, but I was thinking the droplet was intended only to be the executable and its supporting filesystem. If there's already precedent for storing metadata in the droplet (or we're willing to create that precedent) then that would be fine. Storing the SBOM data in the droplet is closer to how the V3 lifecycle handles the SBOM data.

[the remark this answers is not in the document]
robdimsdale: And yes, I don't think you'd want to parse the SBOM data to store it in a structured format in CCDB. I was assuming there was already some way of storing large amounts of (potentially compressed) metadata already in CF/CAPI.

toc/rfc/rfc-draft-add-cnbs.md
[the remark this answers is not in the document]
Gerg: I agree with the above. Also, if the v3 buildpacks are using the Paketo stacks, then the app's stack will determine which set of buildpacks it uses.

toc/rfc/rfc-draft-add-cnbs.md
Gerg: I was intentionally non-prescriptive here about whether it would be appropriate to extend the current Buildpack App Lifecycle to support v3 buildpacks, or whether it made sense to create an entirely separate lifecycle (CNB App Lifecycle?).

If desired, we can update the RFC to be more opinionated on that matter.

toc/rfc/rfc-draft-add-cnbs.md
Gerg: I don't think there are any necessary changes to CF to get this capability. The main Paketo buildpacks are themselves [composites of multiple other buildpacks](https://github.com/paketo-buildpacks/ruby/blob/main/buildpack.toml), which makes it easier to add/remove sub-buildpacks to meet your custom needs.

That said, this could open the door for more advanced operator controls for buildpacks in the future. For instance, if platform operators could customize buildpacks via platform configuration instead of modifying the buildpacks themselves. Definitely outside the scope of this RFC though.

toc/rfc/rfc-draft-add-cnbs.md
Gerg: Adding @c0d1ngm0nk3y @dmikusa @samj1912, in case y'all are interested in this RFC.

dmikusa: This RFC seems awesome though!

We're working out something in libcnb#228, but hopefully, that is resolved soon and we can cut a release. Then buildpacks based on libcnb can get access to service binding info on CF in a transparent way. There are a couple of caveats though. 1. That's only for libcnb-based buildpacks and 2.) you can't really represent all of the `VCAP_SERVICES` data with K8s service bindings so you lose labels, for example.

APM buildpacks are a big user of service bindings, but they are used elsewhere also. Maven/Gradle use them for build secrets, like accessing a private repo. They are also used for injecting trusted CA certificates, and for dependency mappings also (a mechanism to redirect where the buildpack fetches a dependency from).

toc/rfc/rfc-draft-add-cnbs.md
beyhan: I read it as decided and wanted to make it clear but it is even better that you think for different options here. I agree that it makes sense to decide on this when we have more input from phase 1. Regarding extending the current buildpack lifecycle or having a CNB app lifecycle I think we should go with the approach which makes possible next phases/steps like building OCI images and integration with Container registries easier.  At the moment I'm in favour of a separate CNB app lifecycle.

loewenstein: I am not sure if you meant this that way @Gerg, but I agree with @beyhan that we should keep the longer-term in mind and sight when discussing and working on a shorter-term workaround.

I see the benefit of munching Paketo buildpacks and their execution environment (aka CNB lifecycle) together and wrap it up in a v2 compatible shim. Any user can try `cf push -b v3_java_buildpack` on any foundation and give feedback on if and how their app still works.

When starting efforts inside the Cloud Foundry codebase, we should take into account how this can end up in running CNBs natively. We anyway need to run the CNB lifecycle in a Diego scheduled container. We should make sure this stays usable as is and could be pushed into a registry, even if we add logic that extracts a droplet from the result.

For what it's worth, we have written a Cloud Native Buildpack platform with capabilities similar to `pack build` that can run fully inside a container. The [`cnbBuild` step](https://github.com/SAP/jenkins-library/blob/master/cmd/cnbBuild.go) is a bit buried in a larger codebase for a whole CI/CD system / library, but we could certainly help to extract (and maintain) it in a different place or try to push some of the platform logic directly into the `lifecycle`. I've never tried to run in in a Diego Task, but it sure doesn't need a Docker Daemon.

toc/rfc/rfc-draft-add-cnbs.md
loewenstein: I agree with the app dev's point of view, but I also do see the platform engineer/operator view. I don't think that we should increase the system architecture's complexity with a shim in the mid- to long-term.

FWIW we need a CNB lifecycle run in a Diego scheduled container anyway, right? So the main addition to the system architecture would be a registry, right? I would assume that when [configured to use a blobstore like S3](https://github.com/distribution/distribution/blob/main/docs/configuration.md) this [registry implementation](https://github.com/distribution/distribution) will be mostly if not fully stateless.

toc/rfc/rfc-draft-add-cnbs.md
dmikusa: >(I understood that's rather a priority thing than anyone vetoing that) or provide similar logic directly in packit.

It is something we want to do, but I don't know the priority. We've been working for a libcnb v2 for a while now and when that releases, the changes in it will enable packit to base on top of libcnb v2. I don't know how long that will take though. Just getting libcnb v2 shipped has taken some time, but I hope we're getting close. It needs one more thing, support for buildpack extensions.

>did you have a case in mind where this direction of information loss could pose a problem?

I was mostly thinking that it could be a change for application developers. Right now, as an app dev using CF and say the java-cfenv, you can search by things like the label or tags for a class of service. That wouldn't be possible, at least not with the current implementation. If an application were doing that, it would need to change to look by vendor or search by name. The binding type is set to the vendor name, and the service name is set to the binding name.

loewenstein: Well, application code can just continue to consume VCAP_* directly I would assume.

dmikusa: Correct, but there will be users that don't read the fine manual and just build with v3 CNBs and expect it to work. If there are pitfalls, we should try to make them obvious pitfalls, like making the app build break with v3 thereby forcing the user to decide.

Gerg: So, if I can try to distill this discussion down some, it seems like our two options are:

### 1) CF Emulates K8s Service Bindings
This is what is currently proposed in the RFC. Since there (probably) isn't an automatic way to map VCAP_SERVICES into equivalent k8s bindings, this will have to be the subset of bindings that the Paketo Buildpacks care about. This is more of the "buildpack-aware platform" implementation.

It seems like we don't want CF itself to be responsible for mapping VCAP_SERVICES to k8s bindings, because then buildpacks won't be able to add new binding types without a corresponding CF change. Maybe we should contain this logic in a buildpack instead?

**Pros:**
1. Doesn't require any buildpacks changes

**Cons:**
1. Apps can get "double" configuration when the same binding appears in both VCAP_SERVICES and the k8s binding files
1. Depends on adding tmpfs to CF containers

### 2) Paketo Buildpacks Use VCAP_SERVICES
This is continuing down the path of https://github.com/buildpacks/libcnb/pull/227. When writing this RFC, I had assumed this wasn't feasible, but it would be convenient. In this world, the buildpacks would be responsible for consuming either k8s service bindings or VCAP_SERVICES. This is more of the "platform-aware buildpacks" implementation.

**Pros:**
1. Doesn't require any CF changes

**Cons:**
1. Only CF-aware v3 buildpacks will work with VCAP_SERVICES bindings. Maybe reduces number of compatible CNBs (outside of Paketo).

---

Let me know what I missed here. Trying to figure out how to start working this back into the RFC.

loewenstein: > Maybe reduces number of compatible CNBs (outside of Paketo).

FWIW `libcnb` is maintained by the Cloud Native Buildpacks project, not by Paketo. This doesn't invalidate the "only CF-aware buildpacks" restriction. But it is not only about Paketo buildpack.

loewenstein: Also
> It seems like we don't want CF itself to be responsible for mapping VCAP_SERVICES to k8s bindings, because then buildpacks won't be able to add new binding types without a corresponding CF change. Maybe we should contain this logic in a buildpack instead?

If not making CF k8s binding aware, this will limit the solution to relatively CF-aware buildpacks as well.

dmikusa: Yes, that sounds right. Although, I'm not 100% sure I follow the first option. 

You mentioned, `Maybe we should contain this logic in a buildpack instead?` but you can't have a buildpack that translates `VCAP_SERVICES` to bindings because the buildpack detection process runs in parallel so one buildpack cannot intercept and translate them before the others run. Many of the buildpacks check for service bindings at detect time to determine if they will execute their builds, so that is important. 

Whatever performs the translation step needs to do it before the buildpacks run. The libcnb approach works because the buildpacks use that library to fetch bindings, so it can translate just in time. The other option would be to translate them before you run the buildpack lifecycle, but then you'd have the translation logic in the platform, i.e. CF, and it seems like that's not desirable.

Gerg: > You mentioned, Maybe we should contain this logic in a buildpack instead? but you can't have a buildpack that translates VCAP_SERVICES to bindings because the buildpack detection process runs in parallel so one buildpack cannot intercept and translate them before the others run. Many of the buildpacks check for service bindings at detect time to determine if they will execute their builds, so that is important.

Ah, that's good to know.

> The other option would be to translate them before you run the buildpack lifecycle, but then you'd have the translation logic in the platform, i.e. CF, and it seems like that's not desirable.

I think it would be desirable, if we had a way to automatically convert all CF bindings into their equivalent k8s bindings.  For example, using @emalm's [example above](https://github.com/cloudfoundry/community/pull/591#discussion_r1182004165), CF binding name/type/label `google-stackdriver-profiler` becomes k8s binding type `GoogleStackdriverProfiler`. If we could say "Always replace hyphens with intercapitalization" the mapping would be forwards-compatible. If we have to do a manual mapping between CF and k8s bindings, then I'm worried that a CNB adding support for a new binding type will also require a corresponding update to the mapping logic in Cloud Foundry.

This is outside of my core area of knowledge, so let me know if what I'm saying doesn't make sense or is missing something.

dmikusa: I definitely understand your concern there. I think the problem with an automatic mapping like that will be that they don't match up to what the Paketo buildpacks currently expect. For example, we look for `StackdriverProfiler` and `StackdriverDebugger`.

I think that on either side of the service binding interface, you have a problem that the implementor can pick arbitrary values. For example, whoever implements the service broker on CF can pick the values they use for the provider and credentials. Similarly, whoever implements the buildpacks can pick different types to look for in the service bindings.

Since we're talking about CF and Paketo and we're both part of the CFF, it seems reasonable that we agree upon either a standard way of mapping them or that we agree on a well-known set that should be used. We can't stop using the current types in Paketo buildpacks, but we can certainly add new types to look for that are compatible with well-known CF services.

FYI, there are some well-known secret entries in the [service binding spec](https://github.com/servicebinding/spec#well-known-secret-entries), and we should definitely try to use these. Spring Cloud Bindings also has a number of [types that it knows about and reacts to](https://github.com/spring-cloud/spring-cloud-bindings), so we'll want to take that into consideration as well since we have a lot of Spring users.

toc/rfc/rfc-draft-add-cnbs.md
stephanme: I think there is more needed than 'just a registry' to switch from droplets to OCI images. Droplets don't contain the stack and app developers are used to get stack updates automatically for buildpack apps without having to redeploy/restage their apps.
If the CNB lifecycle would produce OCI images, a stack rebase needs to be added to the runtime evacuation process if keep this user experience.

toc/rfc/rfc-draft-add-cnbs.md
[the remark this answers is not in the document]
AP-Hunt: TOC discussion summary: 
* out of scope to have them available for lifecycle
* can iterate on SBOMs stuff after first pass
* don't know what tooling we want 
* resolve to note in the RFC that the SBOM data will reside in the droplet

toc/rfc/rfc-draft-add-cnbs.md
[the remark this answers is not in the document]
AP-Hunt: TOC discussion:
* resolve to provide recommendations for naming, so people can knowingly opt in

toc/rfc/rfc-draft-add-cnbs.md
AP-Hunt: TOC discussion:
* what buildpacks need VCAP_SERVICES at build time? Is it just those that are libcnb based?
    * some buidpacks (e.g. datadog) looking for env variables and injecting an agent at build time
* having buildpacks be aware of VCAP_SERVICES is a viable path
* resolve to update RFC to say buildpacks require access to VCAP_SERVICES at build time, but leave the implementation as an open question to be answered with another RFC

toc/rfc/rfc-draft-add-cnbs.md
AP-Hunt: TOC discussion:
* follow on RFC required once we've learned lessons (phase 2)

toc/rfc/rfc-draft-add-cnbs.md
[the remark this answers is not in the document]
Gerg: https://github.com/cloudfoundry/community/commit/5a7ebf4f81696d4f4dcd6c40b9dcca8dc4597197

toc/rfc/rfc-draft-add-cnbs.md
Gerg: https://github.com/cloudfoundry/community/commit/894bfb3d20057ecd5052032a135eafa31d84e7ef

toc/rfc/rfc-draft-add-cnbs.md
Gerg: https://github.com/cloudfoundry/community/commit/6d6bd816cefc07d71882af16300c93876f8e301a

toc/rfc/rfc-draft-add-cnbs.md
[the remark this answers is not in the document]
Gerg: https://github.com/cloudfoundry/community/commit/ba6fa1a04b256c2e506c7457e1665f2970bd641a

Gerg:
I've made a round of edits, based on feedback from the last TOC meeting. I propose we move this RFC into the final comment period.

beyhan (approved):

stephanme (approved):

emalm (approved):

AP-Hunt (approved):

krismarc:
@Gerg any chance for the community update on this RFC implementation? Is there any additional place where we could read/keep track on it?

I'm excited by this one. I was thinking that paketo buildpacks will be only available in Korifi. This brings back a hope for the main buildpack used by us. Actually, IBM gave up on maintaining the old Liberty buildpack and switched over all their efforts on a new one.

https://github.com/cloudfoundry/ibm-websphere-liberty-buildpack
https://github.com/paketo-buildpacks/liberty

I'm just wondering how I could anyhow inflence on making this buildpack to participate in this initiative. Is there any chance we could loop it into this project? :)

loewenstein:
@krismarc There has been a "successor" RFC with https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0028-cnb-lifecycle.md and all implementation pull requests are open already and ready for review.

There is also an additional RFC to include Cloud Native Buildpacks as system buildpacks with https://github.com/cloudfoundry/community/pull/830.

krismarc:
Sorry to necro this once again. Not sure in which repository I should raise my questions about this feature.

Our CF deployment is completely air-gapped and it's unable to fetch internet dependencies.

I'm trying to run nodejs sample app using cnb described here:
https://docs.cloudfoundry.org/buildpacks/cnb/index.html

..however, I get this:
```
   Writing application metadata: /home/vcap/layers/paketo-buildpacks_ca-certificates/launch.toml <= {Labels:[] Processes:[] Slices:[] BOM:[{Name:helper Metadata:map[layer:helper names:[ca-certificates-helper] version:3.9.0] Launch:true Build:false}]}
   Paketo Buildpack for Node Engine 5.1.4
   Resolving Node Engine version
   Candidate version sources (in priority order):
   <unknown> -> ""
   Selected Node Engine version (using <unknown>): 20.18.1
   Checking if layer /home/vcap/layers/paketo-buildpacks_node-engine/node can be reused
   Checksum of dependency: sha256:c6fa75c841cbffac851678a472f2a5bd612fff8308ef39236190e1f8dbb0e567
   Checksum of layer:
   Build requirements match: false
   Launch requirements match: false
   Executing build process
   Installing Node Engine 20.18.1
   failed to fetch dependency: failed to make request: Get "https://nodejs.org/dist/v20.18.1/node-v20.18.1-linux-x64.tar.xz": dial tcp 104.20.23.46:443: i/o timeout
   ERROR: failed 'build' phase, error: exit status 1
   Error: building failed
   Exit status 234
   Cell 00b37c58-f656-4f88-8cfe-dc709198043d stopping instance 5d5d2e31-bbfd-41ee-9d85-ba9bfbcd695f
   Cell 00b37c58-f656-4f88-8cfe-dc709198043d destroying container for instance 5d5d2e31-bbfd-41ee-9d85-ba9bfbcd695f
CNBBuildFailed - cnb: building failed
FAILED
```

I tried to use binding as described here:
https://paketo.io/docs/howto/configuration/#dependency-mappings

```
# tree bindings/
bindings/
└── node-engine
    └── node
        ├── c6fa75c841cbffac851678a472f2a5bd612fff8308ef39236190e1f8dbb0e567
        └── type
```
unfortunately, with no luck. 

My question is, if it's anyhow supported? How can I use our mirrors? or bind individual dependencies?

loewenstein-sap:
> My question is, if it's anyhow supported? How can I use our mirrors? or bind individual dependencies?

I have never tried this, but if it works at all - you will have to provide the binding in VCAP_SERVICES form I believe. The documentation in Paketo reads like the name wouldn't matter, the type should be `dependency-mapping` and the content a mapping from sha256 to URI.

krismarc:
@loewenstein, could you come up with an example so I could try myself? Not sure if I fully understood properly. Do you mean as 'user-provided' service which will became part of VCAP_SERVICES?

@beyhan I believe this is something to be generally considered, supported and documented as there might be many companies behind sort of proxies and own mirrors which do not let direct downloads from the internet.

krismarc:
> > My question is, if it's anyhow supported? How can I use our mirrors? or bind individual dependencies?
> 
> I have never tried this, but if it works at all - you will have to provide the binding in VCAP_SERVICES form I believe. The documentation in Paketo reads like the name wouldn't matter, the type should be `dependency-mapping` and the content a mapping from sha256 to URI.

This part of RFC seems to explain what you have said:
https://github.com/cloudfoundry/community/blob/8c7298337a8515d7dfae058b3bd1f88ad0eeaf95/toc/rfc/rfc-0030-add-support-for-file-based-service-binding.md#app-runtime-interfaces-wg

loewenstein-sap:
> I managed it to work with BP_DEPENDENCY_MIRROR. https://paketo.io/docs/howto/configuration/#dependency-mirrors
> 
> Yet not clear to me how and if possible to use bindings directly. Paketo instruction doesn't really provide any real examples.

I have no example of individual dependency-mappings in Cloud Foundry available, but it should be about `cf cups` and using the right `type`. The buildpacks themselves are prepared to read from file system at `SERVICE_BINDING_ROOT` or `VCAP_SERVICES`.

However, why would you not stick to `BP_DEPENDENCY_MIRRORS` over of individual bindings?

beyhan:
> could you come up with an example so I could try myself? Not sure if I fully understood properly. Do you mean as 'user-provided' service which will became part of VCAP_SERVICES?

Currently, CF supports only the [VCAP_SERVICES](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-SERVICES) way of defining service bindings. I think what @loewenstein-sap is saying that you have to define the bindings in that format because CF doesn't understand formats. The RFC you found is about introducing another service binding option based on servicebindings.io but it is still in implementation phase. 

> I believe this is something to be generally considered, supported and documented as there might be many companies behind sort of proxies and own mirrors which do not let direct downloads from the internet.

That functionality is something provided by the CNBs and is not a CF feature. If you manage to get the bindings approach also working then we could think where it belongs.  There is another [RFC in progress](https://github.com/cloudfoundry/community/issues/965) which will introduce system buildpacks support for CNB. Providing offline buildpacks as system buildpacks in CF is an approach which works well for air gapped environments.

krismarc:
> > I managed it to work with BP_DEPENDENCY_MIRROR. https://paketo.io/docs/howto/configuration/#dependency-mirrors
> > Yet not clear to me how and if possible to use bindings directly. Paketo instruction doesn't really provide any real examples.
> 
> I have no example of individual dependency-mappings in Cloud Foundry available, but it should be about `cf cups` and using the right `type`. The buildpacks themselves are prepared to read from file system at `SERVICE_BINDING_ROOT` or `VCAP_SERVICES`.
> 
> However, why would you not stick to `BP_DEPENDENCY_MIRRORS` over of individual bindings?

@loewenstein-sap  We don't have mirrors of all possible sources and the path differs in some cases. I was thinking about building some automation which would dynamically craft service bindings depending on our own sources locations. Yet a lot of considerations ahead of me. I've just started playing with CNBs in CF 😄 Another option would be using this: https://paketo.io/docs/howto/configuration/#setting-hostname-mirrors

> That functionality is something provided by the CNBs and is not a CF feature. If you manage to get the bindings approach also working then we could think where it belongs. 

@beyhan, sure, thx! I'll give it a try.

> There is another https://github.com/cloudfoundry/community/issues/965 which will introduce system buildpacks support for CNB. Providing offline buildpacks as system buildpacks in CF is an approach which works well for air gapped environments.

That's what we are currently using. We do provide our own dependencies repos, or we relay on default dependencies shipped with the buildpack, eg. with java buildpack.

loewenstein-sap:
> We don't have mirrors of all possible sources and the path differs in some cases. I was thinking about building some automation which would dynamically craft service bindings depending on our own sources locations. Yet a lot of considerations ahead of me. I've just started playing with CNBs in CF 😄 Another option would be using this: https://paketo.io/docs/howto/configuration/#setting-hostname-mirrors

Got it. https://github.com/dmikusa/binding-tool might be worth a look, although it is focusing on the `pack build` use case and hence creating a file system layout.

krismarc:
> > We don't have mirrors of all possible sources and the path differs in some cases. I was thinking about building some automation which would dynamically craft service bindings depending on our own sources locations. Yet a lot of considerations ahead of me. I've just started playing with CNBs in CF 😄 Another option would be using this: https://paketo.io/docs/howto/configuration/#setting-hostname-mirrors
> 
> Got it. https://github.com/dmikusa/binding-tool might be worth a look, although it is focusing on the `pack build` use case and hence creating a file system layout.

Thanks for the tool. 

I've contacted our 'mirrors' guys and they could create any mirror for me. I'd just need to provide them with the list of all sources locations. In this case, I'd need to build some sort of scrapper and probably buildpacks tomls are going to be the best source as @dmikusa does in his tool [here.](https://github.com/dmikusa/binding-tool/blob/c2d749ea61bf94af0821e33c7490614b8b283eaf/src/deps.rs#L85) Eventually he could add such a function which would just list dependencies locations instead of downloading them. Unfortunately, I don't know Rust yet to contribute.

I had no time yet. However, I'll test individual service bindings via VCAP_SERVICES anyway. You never know when it might be needed with some edge cases.

PR #601, opened 2023-05-12
Archive pip-pop repo

robdimsdale:
This PR archives the https://github.com/cloudfoundry/pip-pop repository, as it is no longer used by the buildpacks team.

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -1816,6 +1816,7 @@ orgs:
         archived: true
         has_projects: true
       pip-pop:
+        archived: true
         has_projects: true
       platform-certification:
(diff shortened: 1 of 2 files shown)


beyhan (approved):

emalm (changes requested):
Thanks, @robdimsdale ! Could you also please remove `pip-pop` from the set of repos in https://github.com/cloudfoundry/community/blob/main/toc/working-groups/app-runtime-interfaces.md? Looks like it's only in the Buildpacks Python area.

robdimsdale:
@emalm done

AP-Hunt (approved):

emalm (approved):

PR #604, opened 2023-05-17
2023 elections voter lists

christopherclark:


emalm (approved):

emalm:
Merging so that the election descriptions, voter lists, and other data are up to date in Elekto. If there are other errors, we'll file additional PRs to correct them.

PR #441, opened 2022-10-05
[ARI] Add temp teams for managing GitHub secrets

Gerg:
- Maintainers currently only have "write" access
- Write access does not allow modifying GitHub secrets
- Teams using GitHub actions depend on secrets
- Current workaround requires working group lead to edit secrets on behalf of the team
- These teams will temporarily give maintainers access to manage secrets while the TOC works on a long-term solution.

PR #618, opened 2023-06-08
Add cflinxufs4-compat-release repo and give sophiewigmore admin access

sophiewigmore:
I will need admin access while I am working on getting initial automation out the door the first release. I may need to amend secrets while we're ironing out any kinks in out automation so it's easier to have the repository access this way

~~I also added an entry for the cflinuxfs4-compat-release here, because I assumed we'd need one.~~

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -2313,9 +2313,11 @@ orgs:
         - arjun024
         - brayanhenao
         - ryanmoran
+        - sophiewigmore
         members: []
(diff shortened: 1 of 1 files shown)


Gerg:
This is a follow-up to https://github.com/cloudfoundry/community/pull/441

Gerg (approved):
I'm going to merge this since the override team was already approved by the TOC, and this PR is updating the maintainers and repos consistent with the ARI WG charter.

PR #630, opened 2023-06-26
[RFC 630] add readiness healthchecks for apps

ameowlia:
For easier viewing: https://github.com/cloudfoundry/community/blob/rfc-readiness-healthchecks/toc/rfc/rfc-draft-readiness-healthchecks.md

### DIFF
--- a/toc/rfc/rfc-draft-readiness-healthchecks.md
+++ b/toc/rfc/rfc-draft-readiness-healthchecks.md
@@ -41,7 +41,7 @@ than the current option.
 
 ## Types of readiness healthcheck
 
-Readiness healthcheck can be either "http" or "tcp" type. The format of healthcheck
+Readiness healthcheck can be either "http" or "port" type. The format of healthcheck
 type is [similar to liveness
 healthcheck](https://docs.cloudfoundry.org/devguide/deploy-apps/healthchecks.html).
 The "process" healthcheck type will not be supported since it doesn't make sense
@@ -90,6 +90,10 @@ applications:
     readiness-health-check-type: http                  # 👈 new property
 ```
 
+New `routable` field in CC API [process stats
+object](https://v3-apidocs.cloudfoundry.org/version/3.141.0/index.html#the-process-stats-object)
+with values `true` or `false` will display the routable status of the process.
+
 ### LRP Design
 
 The readiness healthcheck data will be apart of the desired LRP object.
@@ -117,13 +121,41 @@ The readiness healthcheck data will be apart of the desired LRP object.
   },
 ```
 
+### CLI Changes
+
+The `routable` field of the process stats API object property will be used in
+CF CLI `cf app` output.
+
+```
+     state     routable   since                  cpu    memory          disk         logging            details
+#0   running   yes        2023-06-27T15:07:14Z   0.6%   46.8M of 192M   179M of 1G   0/s of unlimited
+#1   running   no         2023-06-27T15:11:43Z   0.0%   0 of 0          0 of 0       0/s of 0/s
+#2   running   yes        2023-06-27T15:11:43Z   0.0%   0 of 192M       0 of 1G      0/s of 0/s
+```
+
+New CLI options will be added to `cf push` command:
+
+* `--readiness-endpoint` will allows to set the endpoint for http readiness
+  checks.
+* `--readiness-health-check-type` will allows to set the type of the readiness
+  check: "http" or "port".
+
+New CLI commands will be added:
+
+* `get-readiness-health-check` shows the readiness health check performed on an
+  app instance
+* `set-readiness-health-check` updates the readiness health check performed on
+  an app instance
+
 ### Logging and Metrics
 
 #### App logs
 
 When AI readiness healthcheck succeeds a log line is printed to AI logs:
-"Container became ready". When AI readiness healthcheck fails a log line is
-printed to AI logs: "Container became not ready".
+"Container passed the readiness health check. Container marked ready and added
+to route pool". When AI readiness healthcheck fails a log line is printed to AI
+logs: "Container failed the readiness health check. Container marked not ready
+and removed from route pool".
 
 #### App events
 
--- a/toc/rfc/rfc-draft-readiness-healthchecks.md
+++ b/toc/rfc/rfc-draft-readiness-healthchecks.md
@@ -4,7 +4,7 @@
 - Start Date: 2023-06-26
 - Author(s): @ameowlia, @mariash
 - Status: Draft
-- RFC Pull Request: (fill in with PR link after you submit it)
+- RFC Pull Request: https://github.com/cloudfoundry/community/pull/630
 
 
 ## Summary
--- a/toc/rfc/rfc-draft-readiness-healthchecks.md
+++ b/toc/rfc/rfc-draft-readiness-healthchecks.md
@@ -71,7 +71,7 @@ applications:
     health-check-invocation-timeout: 2
     health-check-type: http
     timeout: 80
-    readiness-health-check-http-endpoint: /health      # 👈 new property
+    readiness-health-check-http-endpoint: /ready       # 👈 new property
     readiness-health-check-invocation-timeout: 2       # 👈 new property
     readiness-health-check-type: http                  # 👈 new property
 ```
--- a/toc/rfc/rfc-draft-readiness-healthchecks.md
+++ b/toc/rfc/rfc-draft-readiness-healthchecks.md
@@ -41,12 +41,9 @@ than the current option.
 
 ## Types of readiness healthcheck
 
-Readiness healthcheck can be either "http" or "port" type. The format of healthcheck
-type is [similar to liveness
-healthcheck](https://docs.cloudfoundry.org/devguide/deploy-apps/healthchecks.html).
-The "process" healthcheck type will not be supported since it doesn't make sense
-to have "process" readiness healthcheck type. Once any defined process exits AI
-is marked as crashed.
+[Similar to liveness healthchecks](https://docs.cloudfoundry.org/devguide/deploy-apps/healthchecks.html), readiness healthcheck can be of type "http", "port", or "process".
+However, when a user selects the "process" healthcheck type, nothing will be passed to the LRP, because once a process exits the AI
+is marked as crashed and Diego will attempt a restart. The default readiness healthcheck type is "process", which is backwards compatible.
 
 ## Rolling deploys
 
@@ -88,6 +85,7 @@ applications:
     readiness-health-check-http-endpoint: /ready       # 👈 new property
     readiness-health-check-invocation-timeout: 2       # 👈 new property
     readiness-health-check-type: http                  # 👈 new property
+    readiness-health-check-interval: 5                 # 👈 new property
 ```
 
 New `routable` field in CC API [process stats
@@ -135,10 +133,12 @@ CF CLI `cf app` output.
 
 New CLI options will be added to `cf push` command:
 
-* `--readiness-endpoint` will allows to set the endpoint for http readiness
+* `--readiness-endpoint` will set the endpoint for http readiness
   checks.
-* `--readiness-health-check-type` will allows to set the type of the readiness
-  check: "http" or "port".
+* `--readiness-health-check-type` will set the type of the readiness
+  check: "http", "port", or "process".
+* `--readiness-health-check-interval` will set the interval of the readiness
+  check. Must be an integer greater than 0."
 
 New CLI commands will be added:
 
--- a/toc/rfc/rfc-draft-readiness-healthchecks.md
+++ b/toc/rfc/rfc-draft-readiness-healthchecks.md
@@ -1,0 +1,111 @@
+# Meta
+[meta]: #meta
+- Name: Readiness Healthchecks
+- Start Date: 2023-06-26
+- Author(s): @ameowlia, @mariash
+- Status: Draft
+- RFC Pull Request: (fill in with PR link after you submit it)
+
+
+## Summary
+
+Add a readiness healthcheck option for apps. When the readiness healthcheck
+passes, the app is marked "ready" and the app will be routable. When the
+readiness healthcheck fails, the app is marked as "not ready" and its route will
+be removed from gorouter's route table.
+
+## Problem
+
+With the current implementation of application healthchecks, when the
+application healthcheck detects that an app instance (AI) is unhealthy, then
+Diego will stop the AI, delete the AI, and reschedule a new AI.
+
+This is too aggressive from some apps. There could be many reasons why a single
+request could fail, but the app is actually running fine. Additionally, many
+applications have a warm up period where they are not ready to receive requests
+immediately. For example, apps might need to populate caches, load data, or wait
+for external services before they mark themselves as routable. In these cases,
+the app should be kept alive, but in a non-routable state.
+
+## Proposal
+
+### Summary
+We intend to support readiness healthchecks. (This was requested previously in
+this [issue](https://github.com/cloudfoundry/cloud_controller_ng/issues/1706).)
+This would be an additional healthcheck that app developers could configure.
+When the readiness healthcheck passes, the app is marked "ready" and the app
+will be routable. When the readiness healthcheck fails, the app is marked as
+"not ready" and its route will be removed from gorouter's route table.
+This new readiness healthcheck will give users a healthcheck option that is less
+drastic than the current option.
+
+### Architecture Overview
+This feature will require changes in the following releases
+
+* CF CLI
+* Cloud Controller
+* Diego
+* Routing
+
+1. The cloud controller will store this new data, before passing it onto the BBS
+   as part of the desired LRP.
+2. The Diego executor will see these new readiness healthchecks on the desired
+   LRP and will run the healthchecker binary in the app container with
+   configuration provided.
+3. When the readiness healthcheck succeeds, the container will be marked as
+   "ready". When the readiness healthcheck fails, the container will be marked
+   as "not ready".
+4. When the route emitter gets route information, it will inspect if the AI is
+   ready or not ready. It will emit registration or unregistration messages as
+   appropriate for the gorouter to consume
+
+### CC Design
+Users will be able to set the healthcheck via the app manifest.
+
+```
+applications:
+- name: test-app
+  processes:
+  - type: web
+    health-check-http-endpoint: /health
+    health-check-invocation-timeout: 2
+    health-check-type: http
+    timeout: 80
+    readiness-health-check-http-endpoint: /health      # 👈 new property
+    readiness-health-check-invocation-timeout: 2       # 👈 new property
+    readiness-health-check-type: http                  # 👈 new property
+```
+
+### LRP Design
+
+The readiness healthcheck data will be apart of the desired LRP object.
+
+"check_definition": {
+    "checks": [
+      {
+        "http_check": {
+          "port": 8080,
+          "path": "/health",
+          "request_timeout_ms": 10000
+        },
+      }
+    ],
+    "readiness_checks": [                  # 👈 new property
+      {
+        "tcp_check": {
+          "port": 8080,
+          "connect_timeout_ms": 10000
+        },
+      }
+    ],
+    "log_source": ""
+  },
+
+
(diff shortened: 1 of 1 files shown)


beyhan (commented):
In general I really like the proposal. I added some minor comments and questions I have.

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,111 @@
...
+      }
+    ],
+    "readiness_checks": [                  # 👈 new property
+      {
+        "tcp_check": {
+          "port": 8080,
+          "connect_timeout_ms": 10000
+        },
+      }
+    ],
```
beyhan, line 100: - The manifest section defines `readiness-health-check-type: http` but in the data section we have `tcp_check` for the readiness health check. Is this a type?
- The readiness check data doesn't define a `path` field. Is this intended?

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,111 @@
...
+    "log_source": ""
+  },
+
+
+### Open Questions
```
stephanme, line 105: Additional open question:
* How shall a rolling app deployment behave? Use the health check as today or use (if configured) the readiness check to ensure that the updated app instance is routable before an old instances is stopped.

I tend to use the readiness check (if configured).

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,111 @@
...
+```
+applications:
+- name: test-app
+  processes:
+  - type: web
+    health-check-http-endpoint: /health
+    health-check-invocation-timeout: 2
+    health-check-type: http
+    timeout: 80
+    readiness-health-check-http-endpoint: /health      # 👈 new property
```
maxmoehl, line 74: (minor) The two checks should use different endpoints:

```suggestion
    readiness-health-check-http-endpoint: /ready       # 👈 new property
```

toc/rfc/rfc-draft-readiness-healthchecks.md
maxmoehl: I thought this was intended to show that a readiness healthcheck doesn't have to be a HTTP check and can be any type of healthcheck. But it would probably be more realistic to define the healthcheck on layer 4 and the readiness check on layer 7.

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,111 @@
...
+    readiness-health-check-invocation-timeout: 2       # 👈 new property
+    readiness-health-check-type: http                  # 👈 new property
```
peanball, line 76: The example is a little misleading, as it checks the same endpoint with the same timeout. Maybe we can highlight that there could be different timings or different endpoints ("liveness" via health check vs. "readiness").

```suggestion
  - type: web
    health-check-http-endpoint: /health
    health-check-invocation-timeout: 2
    health-check-type: http
    timeout: 80
    readiness-health-check-http-endpoint: /ready       # 👈 new property
    readiness-health-check-invocation-timeout: 2       # 👈 new property
    readiness-health-check-type: http                  # 👈 new property
```

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,111 @@
...
+   configuration provided.
+3. When the readiness healthcheck succeeds, the container will be marked as
+   "ready". When the readiness healthcheck fails, the container will be marked
+   as "not ready".
+4. When the route emitter gets route information, it will inspect if the AI is
+   ready or not ready. It will emit registration or unregistration messages as
+   appropriate for the gorouter to consume
+
+### CC Design
+Users will be able to set the healthcheck via the app manifest.
```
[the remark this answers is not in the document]
beyhan: I also would vote to add the option to configure the timeout with `--timeout` in `set-health-check` as described in this [issue](https://github.com/cloudfoundry/cli/issues/2366). In my opinion It is getting more relevant to update the timeout with `set-health-check` with this new type of health check.

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,111 @@
...
+    readiness-health-check-http-endpoint: /health      # 👈 new property
+    readiness-health-check-invocation-timeout: 2       # 👈 new property
+    readiness-health-check-type: http                  # 👈 new property
+```
+
+### LRP Design
+
+The readiness healthcheck data will be apart of the desired LRP object.
+
+"check_definition": {
```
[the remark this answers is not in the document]
maxmoehl: 😉 
````suggestion
```
"check_definition": {
```
````

(but since your comment only is on the one line I can't make a proper suggestion)

File: toc/rfc/rfc-draft-readiness-healthchecks.md [hunk 7]
```diff
@@ -0,0 +1,111 @@
...
+    "readiness_checks": [                  # 👈 new property
+      {
+        "tcp_check": {
+          "port": 8080,
+          "connect_timeout_ms": 10000
+        },
+      }
+    ],
+    "log_source": ""
+  },
```
peanball, line 102: Putting this into a code fence would be easier to read in the markdown representation.
````suggestion
```
"check_definition": {
    "checks": [
      {
        "http_check": {
          "port": 8080,
          "path": "/health",
          "request_timeout_ms": 10000
        },
      }
    ],
    "readiness_checks": [                  # 👈 new property
      {
        "tcp_check": {
          "port": 8080,
          "connect_timeout_ms": 10000
        },
      }
    ],
    "log_source": ""
  },
```
````

File: toc/rfc/rfc-draft-readiness-healthchecks.md (hunk 1 above)
Comment (beyhan) [reply toc/rfc/rfc-draft-readiness-healthchecks.md:100]: This brings me to the point that we have to clarify in this RFC which types of health checks will be possible/allowed here because I can imagine that a `process` type of health check won't make much sense here and most probably technically not possible.

**Comment by maxmoehl** [comment]
Currently diego replaces the envoy certificate when the application instance is going down. By doing so gorouter will fail to connect to envoy because of the wrong certificate while the app is known to be unavailable and the de-registration has not yet been propagated to gorouter.

The same mechanism could be used for readiness checks to prevent requests from being set to the application in the small timeframe between the readiness check going down and the de-registration being propagated to gorouter.

File: toc/rfc/rfc-draft-readiness-healthchecks.md [hunk 8]
```diff
@@ -0,0 +1,111 @@
...
+
+
+### Open Questions
+* What logging and metrics would be helpful for app devs and operators?
```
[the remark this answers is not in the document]
mariash: @beyhan so far we are adding log line to app logs similar to how we log when container becomes healthy/unhealthy. "Container became ready", "Container became not ready". 

For app events or metrics would be interesting to hear use cases, when event would be useful vs metric.

toc/rfc/rfc-draft-readiness-healthchecks.md
mariash: 1. For rolling deploys, I think it makes sense to account for the application instance readiness.
2. Yes, how about routable field with values => yes/no?

toc/rfc/rfc-draft-readiness-healthchecks.md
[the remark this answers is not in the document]
mariash: As a first iteration we can configure retrieve information about readiness health checks in the manifest. But eventually, cf CLI should provide options to configure it. 
How about push command flags: `--readiness-endpoint` and `--readiness-health-check-type`. Same for CLI commands `get-readiness-health-check` and `set-readiness-health-check`

mariash:
@maxmoehl I am not sure I see the benefit of updating envoy certificate when readiness health check fails, since this is not an immediate action either given that envoy is watching changes on the filesystem.

toc/rfc/rfc-draft-readiness-healthchecks.md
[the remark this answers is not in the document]
maxmoehl: Yes, still the changes to the CLI should be outlined in the RFC. We might decide to implement them at a later stage but this RFC should describe the feature as a whole as much as possible.

maxmoehl:
> @maxmoehl I am not sure I see the benefit of updating envoy certificate when readiness health check fails, since this is not an immediate action either given that envoy is watching changes on the filesystem.

what I'm saying is that the logic to [replace the certificates](https://github.com/cloudfoundry/executor/blob/2c29c7e2fbee4bfa11901a61b0246b67a47c0555/depot/containerstore/proxy_config_handler.go#L251-L263) already exists and diego is [making use of it](https://github.com/cloudfoundry/executor/blob/2c29c7e2fbee4bfa11901a61b0246b67a47c0555/depot/containerstore/credmanager.go#L270) when an AI goes down. Since all the logic is already in place we might as well re-use it for the readiness checks since they have very similar mechanisms and it would help with potentially dropped/delayed nats messages.

toc/rfc/rfc-draft-readiness-healthchecks.md
beyhan: 2. I'm fine with routable with values => yes/no

toc/rfc/rfc-draft-readiness-healthchecks.md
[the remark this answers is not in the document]
beyhan: Sounds good and it is consistent with what we have today available in CF CLI for health check configuration.

toc/rfc/rfc-draft-readiness-healthchecks.md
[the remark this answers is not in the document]
beyhan: We have for example an alert notification service which works based on events. I know that people configure alerts for app.crash events. I can imagine that people could be interested in configuring alerts for apps which are not routable.

Regarding the metrics I don't have any concrete use case. In the linked comment App Autoscaling was mentioned. Additionally, monitoring dashboards could be another use case where routable instances could be visualised. To my knowledge we don't have any sophisticated metrics for apps natively supported in CF. We have only the [container metrics](https://docs.cloudfoundry.org/loggregator/container-metrics.html). Usually, CF users use third party tools to collect metrics (mainly around resource usage) but I can imagine that third party tools will have hard time to generate a metric about whether an app instance is routable or not.

toc/rfc/rfc-draft-readiness-healthchecks.md
beyhan: Thanks for the updates!

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Summary
+We intend to support readiness healthchecks. (This was requested previously in
+this [issue](https://github.com/cloudfoundry/cloud_controller_ng/issues/1706).)
+This would be an additional healthcheck that app developers could configure.
+When the readiness healthcheck passes, the app is marked "ready" and the app
+will be routable. When the readiness healthcheck fails, the app is marked as
+"not ready" and its route will be removed from gorouter's route table.
+This new readiness healthcheck will give users a healthcheck option that is less
+drastic than the current option.
```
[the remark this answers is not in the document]
beyhan: Thanks!

toc/rfc/rfc-draft-readiness-healthchecks.md
beyhan: My questions got answered here. @stephanme can we resolve this?

toc/rfc/rfc-draft-readiness-healthchecks.md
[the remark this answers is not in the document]
beyhan: @mariash the proposed CLI options are missing in the RFC. I think it makes sense to add them or?

toc/rfc/rfc-draft-readiness-healthchecks.md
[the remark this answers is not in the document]
beyhan: The RFC has now a proposal for an event. Regarding the metrics part I don't have any concrete use case, that is why we can leave it for now. I think it makes sense to have a separate RFC which looks into possible app metrics. @mariash do you have any additional comments here?

ChrisMcGowan (approved):

toc/rfc/rfc-draft-readiness-healthchecks.md
stephanme: An extra `routable` field in the [process stats object](https://v3-apidocs.cloudfoundry.org/index.html#the-process-stats-object) should work for the CF API. Maybe we can add this to the CC section above.

My questions are answered as well.

toc/rfc/rfc-draft-readiness-healthchecks.md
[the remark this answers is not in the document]
mariash: @beyhan makes sense. We can add rfc for metrics later. I am also updating the log line to clarify that instance becomes un-routable per suggestion by @ameowlia.

toc/rfc/rfc-draft-readiness-healthchecks.md
mariash: Updating RFC with this information.

peanball:
We just had a discussion for a customer scenario issue that might be interesting for this RFC.

The readiness probe is a scheduled polling, like the health check (e.g. every n seconds). One the readiness state changes, the cascade of route updates, etc. takes place.

What we would like to add is an immediate control from the app to the readiness state checker, to be able to say: "My app is not ready, right now!".

The other way around this of course also works and could save some time, but the case where an app becomes _not ready_ is from our point of view much more important.

It would allow app developers to protect their app from an overload scenario.

Consequences or requests for the "immediate" part of the state change:
* New request should not be accepted right away. This means that possibly the TLS termination proxy (i.e. Envoy) could be adjusted to serve an invalid TLS cert, as it does for stopping instances.
* The cascade of messages to Gorouter should still be done. "Disabling" the TLS proxy is only to increase the responsiveness of the "not ready" state.

What do you think about this, and do you see that this might still fit in the frame of this RFC?

toc/rfc/rfc-draft-readiness-healthchecks.md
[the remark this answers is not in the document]
beyhan: Thanks!

toc/rfc/rfc-draft-readiness-healthchecks.md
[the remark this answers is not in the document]
beyhan: This is now incorporated into the RFC. I'm resolving. Thanks @mariash for the updates!

mariash:
@beyhan Can this be resolved by lowering the poll interval?

beyhan:
> @beyhan Can this be resolved by lowering the poll interval?

@mariash at the moment the interval is 30s which is also documented in [here](https://docs.cloudfoundry.org/devguide/deploy-apps/healthchecks.html). What options to lower the interval we have? It could help but won't be perfect because there is a gap depending on how long the poll interval is in which the app instance could get requests.

@ameowlia for completeness could you please add the concerns you expressed during the TOC meeting yesterday regarding the [replace the certificates](https://github.com/cloudfoundry/executor/blob/2c29c7e2fbee4bfa11901a61b0246b67a47c0555/depot/containerstore/proxy_config_handler.go#L251-L263) suggestion.

ameowlia:
@mariash - can you add to the RFC about exposing the interval? It won't be a perfect fix, for this issue, but could improve it.

peanball:
No objections to continue without immediate state change for this RFC and consider it in a larger context.

We are currently not using c2c networking because of the missing load balancing across instances and other bits (e.g. retries) that we get via gorouter.

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,172 @@
...
+ready" and its route will be removed from gorouter's route table. This new
+readiness healthcheck will give users a healthcheck option that is less drastic
+than the current option.
+
+## Types of readiness healthcheck
+
+Readiness healthcheck can be either "http" or "port" type. The format of healthcheck
+type is [similar to liveness
+healthcheck](https://docs.cloudfoundry.org/devguide/deploy-apps/healthchecks.html).
+The "process" healthcheck type will not be supported since it doesn't make sense
```
sethboyles, line 47: Does this need to be updated?  In the PRs for this feature, `process` is supported as an option from the API (even if it effectively means there is no readiness check)

ameowlia: Good catch. Will update.

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,172 @@
...
+to have "process" readiness healthcheck type. Once any defined process exits AI
+is marked as crashed.
```
ameowlia, line 49:
```suggestion
[Similar to liveness healthchecks](https://docs.cloudfoundry.org/devguide/deploy-apps/healthchecks.html), readiness healthcheck can be of type "http", "port", or "process".
However, when a user selects the "process" healthcheck type, nothing will be passed to the LRP, because once a process exits the AI
is marked as crashed and Diego will attempt a restart. The default readiness healthcheck type is "process", which is backwards compatible.
```

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,172 @@
...
+- name: test-app
+  processes:
+  - type: web
+    health-check-http-endpoint: /health
+    health-check-invocation-timeout: 2
+    health-check-type: http
+    timeout: 80
+    readiness-health-check-http-endpoint: /ready       # 👈 new property
+    readiness-health-check-invocation-timeout: 2       # 👈 new property
+    readiness-health-check-type: http                  # 👈 new property
```
ameowlia, line 90:
```suggestion
    readiness-health-check-type: http                  # 👈 new property
    readiness-health-check-interval: 5                 # 👈 new property
```

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,172 @@
...
+#1   running   no         2023-06-27T15:11:43Z   0.0%   0 of 0          0 of 0       0/s of 0/s
+#2   running   yes        2023-06-27T15:11:43Z   0.0%   0 of 192M       0 of 1G      0/s of 0/s
+```
+
+New CLI options will be added to `cf push` command:
+
+* `--readiness-endpoint` will allows to set the endpoint for http readiness
+  checks.
+* `--readiness-health-check-type` will allows to set the type of the readiness
+  check: "http" or "port".
```
ameowlia, line 141:
```suggestion
* `--readiness-health-check-type` will set the type of the readiness
  check: "http", "port", or "process".
* `--readiness-health-check-interval` will set the interval of the readiness
  check. Must be an integer greater than 0."
```

toc/rfc/rfc-draft-readiness-healthchecks.md
```diff
@@ -0,0 +1,172 @@
...
+```
+     state     routable   since                  cpu    memory          disk         logging            details
+#0   running   yes        2023-06-27T15:07:14Z   0.6%   46.8M of 192M   179M of 1G   0/s of unlimited
+#1   running   no         2023-06-27T15:11:43Z   0.0%   0 of 0          0 of 0       0/s of 0/s
+#2   running   yes        2023-06-27T15:11:43Z   0.0%   0 of 192M       0 of 1G      0/s of 0/s
+```
+
+New CLI options will be added to `cf push` command:
+
+* `--readiness-endpoint` will allows to set the endpoint for http readiness
```
ameowlia, line 138:
```suggestion
* `--readiness-endpoint` will set the endpoint for http readiness
```

ameowlia:
Starting Final Comment Period. FCP should end on Aug 22, 2023.

stephanme (approved):

rkoster (approved):

ChrisMcGowan (approved):

joergdw:
The link on the top of this PR does not work for me. <https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0020-readiness-healthchecks.md> however does.

PR #637, opened 2023-06-30
Add draft RFC for using Paketo stacks in CF Deployment

Gerg:
For easier viewing: https://github.com/cloudfoundry/community/blob/rfc_paketo_stacks/toc/rfc/rfc-draft-add-paketo-stacks.md

cc @robdimsdale, @dsabeti, @dsboulder, @sophiewigmore, @geofffranks

### DIFF
--- a/toc/rfc/rfc-draft-add-paketo-stacks.md
+++ b/toc/rfc/rfc-draft-add-paketo-stacks.md
@@ -1,0 +1,133 @@
+# Meta
+[meta]: #meta
+- Name: Add Paketo Stacks to CF Deployment
+- Start Date: 2023-06-30
+- Author: @gerg
+- Contributing Authors: @robdimsdale, @dsabeti, @dsboulder, @sophiewigmore, @geofffranks
+- Status: Draft
+- RFC Pull Request: TBD
+
+## Summary
+
+This RFC is a continuation of
+[RFC 0017: Add Cloud Native Buildpacks](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0017-add-cnbs.md)
+introduced by
+[community#591](https://github.com/cloudfoundry/community/pull/591).
+
+The [Paketo](https://paketo.io/) project releases a series of stacks. These
+stacks are rootFS for the build and run containers for apps using the Paketo
+buildpacks. Now that CF Deployment is adopting the Paketo Cloud Native
+Buildpacks (CNBs), there is an opportunity to also adopt the Paketo stacks.
+Adopting Paketo stacks would further unify the Cloud Foundry ecosystem and
+provide some benefits to Cloud Foundry users, as discussed in this RFC.
+
+## Problem
+
+Security scanning technology has advanced significantly in recent years,
+allowing Cloud Foundry operators to more easily scan app containers for CVEs.
+This, combined with a renewed industry focus on CVE management, has increased
+Cloud Foundry operator sensitivity to the CVEs present in app containers.
+
+Currently, a major source of these CVEs is the collection of Ubuntu packages
+that are included in cflinuxfs4. Cloud Foundry has made some improvements in
+this area, for example by excluding Ruby and Python from cflinuxfs4, but there
+is still room for improvement. Adopting the Paketo stacks in CF Deployment will
+provide Cloud Foundry operators and app developers with tools to reduce CVE
+exposure in running app containers.
+
+## Proposal
+
+### Background
+
+A primary benefit of Paketo is that built app images have a reduced number of
+operating system packages. Paketo accomplishes this through two mechanisms:
+lightweight stacks and separate build and run images.
+
+#### Lightweight Stacks
+
+Ignoring Windows stacks, CF Deployment currently only offers a single stack:
+cflinuxfs4. In contrast,
+[Paketo offers four "weights" of stacks](https://paketo.io/docs/concepts/stacks/#what-paketo-stacks-are-available)
+with varying sets of included Operating System packages:
+[Full](https://github.com/paketo-buildpacks/jammy-full-stack),
+[Base](https://github.com/paketo-buildpacks/jammy-base-stack),
+[Tiny](https://github.com/paketo-buildpacks/jammy-tiny-stack),
+and [Static (experimental)](https://github.com/paketo-buildpacks/jammy-static-stack)
+. The Paketo Full stack is roughly equivalent to cflinuxfs4, whereas the Base,
+Tiny, and Static stacks include progressively fewer Ubuntu packages.
+
+#### Separate Build and Run Images
+
+CF Deployment uses the same rootFS for building and running applications. This
+means that packages that are only needed for building applications (e.g.
+compilers like `gcc`) are also present in running app containers. The Paketo
+stacks instead [come with both Build and Run images](https://paketo.io/docs/concepts/stacks/#what-is-a-stack)
+, where build-time Ubuntu packages are only included in the Build image.
+
+#### Conclusion
+
+Accounting for both of these factors, there are two "dimensions" to Paketo
+stacks that affect the number of Ubuntu packages included in app containers:
+weight and build/run. CF Deployment could one day offer all of these options,
+so Cloud Foundry users can select the stack that matches their app's Ubuntu
+package needs and security posture.
+
+### Goals
+
+1. Align CF Deployment with the main focus of the Buildpacks & Stacks team
+1. Bring new capabilities to CF Deployment users, including:
+    1. Lightweight Stacks
+    1. Separate Build and Run Images
+1. Increase cohesion and app portability between CF Deployment and Korifi, via
+   shared Paketo stacks
+1. Increase user base for the Paketo stacks
+1. Open the door for eventual deprecation of the cflinuxfs* stacks, reducing
+   maintenance costs (cflinuxfs* deprecation is NOT included in this RFC)
+
+### High-Level Implementation
+
+Instead of using cflinuxfs4, apps using the new CNBs will use the
+[Paketo Jammy Full stack](https://github.com/paketo-buildpacks/jammy-full-stack)
+. The Jammy Full stack will include two rootFSs: build and run. The Paketo
+Jammy Full stack is very similar to cflinuxfs4, which will reduce the
+difficulty of migrating existing Cloud Foundry apps to the Paketo stacks and
+buildpacks. This will also lay the groundwork for eventually supporting the
+remaining Paketo Jammy stacks, which will offer better security posture for
+interested Cloud Foundry users.
+
+#### App Developer Experience
+
+The Paketo Full stack will be modeled as a single Cloud Foundry stack with two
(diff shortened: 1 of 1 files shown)


toc/rfc/rfc-draft-add-paketo-stacks.md
```diff
@@ -0,0 +1,133 @@
...
+```sh
+cf push my-app --stack=paketo_jammy_full
+```
+
+#### Initial Scope
+
+1. Produce a tarball of the Paketo Full stack (instead of an OCI image)
+1. Create a BOSH release for the Paketo Full stack
+1. Update CNB shims to use the Paketo Full stack instead of cflinuxfs4
+1. Add CF support for separate build and run rootFSs
```
beyhan, line 116: I have two points here:

1. My understanding from the [RFC for the first phase](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0017-add-cnbs.md) was that it aims to introduce CNB experience in CF without requiring any changes in the CF platform itself. Is that correct or this RFC will be required to try the shimmed CNBs in a CF environment? 
2. I really like the idea to introduce support for build and run rootFSs in CF. This could be beneficial also for the current V2 buildpacks.  My question is around the plan to create a tarball of the Paketo Full stack and introduce a rootFSs for the CNB use case. If I understand that correctly the approach will deviate a lot from the [CNB Platform Interface Specification](https://github.com/buildpacks/spec/blob/main/platform.md#stacks) and I think it will make an eventual support of OCI based app builds at a later point of time harder to achieve. My suggestion is that we should evaluate at this point what does it mean to go with a more CNB closed lifecycle approach vs. V2 based one. We should put out of scope the registry part in the [CNB specification](https://github.com/buildpacks/spec/blob/main/platform.md#exporter) here but at least the rest should be taken in consideration. 

I'm really interested to get other opinions on this.

Gerg: > My understanding from the [RFC for the first phase](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0017-add-cnbs.md) was that it aims to introduce CNB experience in CF without requiring any changes in the CF platform itself. Is that correct or this RFC will be required to try the shimmed CNBs in a CF environment?

Without this RFC, the plan is to release the CNB shims on the cflinuxfs4 stack. If this RFC is adopted, then that plan may be modified.

Buildpacks can be compatible with multiple stacks, so it is an option to release the CNB shims for both cflinuxfs4 and the Paketo stacks, especially since they are both based on Ubuntu 22.04 Jammy. It's mostly just a question of whether it will be worth the effort to release on both stacks.

I'm happy to amend this RFC based on what reviewers think the best option is for CNB stack compatibility.

Gerg: > I really like the idea to introduce support for build and run rootFSs in CF. This could be beneficial also for the current V2 buildpacks.

It's certainly a possibility that we could extend this to the v2 buildpacks. I don't think we currently have bandwidth to do that, since most of the focus of the buildpacks team is on Paketo and CF Deployment Paketo integration right now. It would also be tricky to roll out for cflinuxfs4, since it would be a breaking change for the stack to lose a bunch of OS packages at runtime.

Gerg: > My question is around the plan to create a tarball of the Paketo Full stack and introduce a rootFSs for the CNB use case. If I understand that correctly the approach will deviate a lot from the [CNB Platform Interface Specification](https://github.com/buildpacks/spec/blob/main/platform.md#stacks) and I think it will make an eventual support of OCI based app builds at a later point of time harder to achieve.

From our experiments, it's fairly easy to translate between the Paketo OCI images and CF rootFS tgzs, so this was intended to be a low-effort step to incremental Paketo adoption in CF Deployment.

From my perspective, this RFC steps us slightly closer to CNB Platform adoption by:
1. Using the Paketo stacks
2. Enabling differences between the build and run environments

That said, I'm by no means an expert on the CNB Platform spec. I'd love to hear more details about your concerns on how this might make future steps more difficult.

cc @robdimsdale

robdimsdale: Paketo stacks are OCI images, which is to say they are essentially a filesystem with metadata. Cloud Foundry stacks are just filesystems in a tgz format. Therefore the simplest way to get a Paketo stack into Cloud Foundry is to get a tgz of the stack image's filesystem.

It's true that this approach discards the OCI image metadata associated with the filesystem, but that doesn't prevent the filesystem from working with Cloud Foundry today. Additionally, I don't see how adopting tarballs of the filesystems as a first step precludes future support for the OCI image stacks directly.

Am I missing something here?

rkoster: > - I’m not so familiar with the SBOM generation but does the generated SBOM describe the stack used for the app?
>   - If yes, then a future update of the stack via the CF platform will invalidate the generated SBOM of an app. We should decide how to deal with this depending on our decisions.

This is an excellent point, I'm also curious about how the SBOM generation part works with the build-and-run stacks. If the SBOM is generated on the build stack it will show dependencies that might not be available on the run stack.

loewenstein:
With the [Remove Stacks & Mixins](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md) RFC as well as the [work on Extensions](https://github.com/buildpacks/rfcs/blob/main/text/0105-dockerfiles.md) close to being done - I wonder how long an integration of Cloud Native Buildpacks that separates stacks from droplets can carry us.

I agree with the [Add Cloud Native Buildpacks](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0017-add-cnbs.md) RFC focusing on quick evaluation. Wrapping Cloud Native Buildpacks in a v2 API that produces droplets that can be run on top of a Cloud Foundry stack enables applications with minimal or no change in Cloud Foundry at all. This will enable our community to check the Paketo buildpacks for compatibility and will enable us to improve that to the point of getting feasible for applications to consider switching over.

I do believe that any step further should take into account how Cloud Native Buildpacks work and are packaged. The most feasible thing to deliver and control for a platform seems to be Cloud Native Builders - combining root file systems and buildpacks. Indeed, the new use cases enabled with [Extensions](https://github.com/buildpacks/rfcs/blob/main/text/0105-dockerfiles.md) will likely only be available when we take this into account.

Given that the cflinuxfs4 and Paketo Full Stack are likely very similar if not identical - I would hope that for the early phase of discovery we don't need a Paketo specific Cloud Foundry Stack. As I tried to picture above, I don't think that we should have a Paketo specific Cloud Foundry Stack for anything beyond.

FWIW Keeping the separation could lead to the Paketo project non being able to leverage the full potential of changes in the Cloud Native Buildpack ecosystem.

robdimsdale:
> Given that the cflinuxfs4 and Paketo Full Stack are likely very similar if not identical - I would hope that for the early phase of discovery we don't need a Paketo specific Cloud Foundry Stack. As I tried to picture above, I don't think that we should have a Paketo specific Cloud Foundry Stack for anything beyond.

We explored this and unfortunately they're not identical. For example the CF stacks require a `vcap` user, but the Paketo stacks use `cnb` user. Either we provide both users in the Paketo stacks, or we make separate stacks.

toc/rfc/rfc-draft-add-paketo-stacks.md
robdimsdale: For the Paketo stacks, we provide SBOMs for both build and run images (e.g. [build receipt](https://github.com/paketo-buildpacks/jammy-base-stack/releases/download/v0.1.68/jammy-base-stack-0.1.68-run-receipt.cyclonedx.json), [run receipt](https://github.com/paketo-buildpacks/jammy-base-stack/releases/download/v0.1.68/jammy-base-stack-0.1.68-run-receipt.cyclonedx.json)). Similar we provide SBOMs for each of the buildpacks.

Merging the two is a responsibility of the Platform. E.g. `pack`/`kpack` are responsible for this. In a CF world, the tool that provides the V2<->V3 API wrapper would be acting as the Platform, and so would have to collate and expose the SBOMs from across the image layers, and present that to the CF API.

Gerg: In the current world where stacks are not included in droplets, we would only want to include buildpack SBOMs, right? The stack can change for the application without requiring a restage.

Gerg:
> > FWIW Keeping the separation could lead to the Paketo project non being able to leverage the full potential of changes in the Cloud Native Buildpack ecosystem.
>
> I assume this is meant to read:
>
> Keeping the separation could lead to the Cloud Foundry project non being able to leverage the full potential of changes in the Cloud Native Buildpack ecosystem.

My reading of this is that if CF doesn't implement the CNB specification properly, then Paketo may get caught in a situation where they can't incorporate newer versions of the CNB specification without breaking CF.

loewenstein:
> My reading of this is that if CF doesn't implement the CNB specification properly, then Paketo may get caught in a situation where they can't incorporate newer versions of the CNB specification without breaking CF.

Exactly @Gerg.

robdimsdale:
Ah ok, thanks. I made a few assumptions that lead me to not consider much risk to Paketo:
1. We would have to maintain separate stacks (for the different users mentioned above)
2. We wouldn't maintain separate buildpacks, nor would we make CNB-incompatible changes to the buildpacks
3. The concept of a builder wouldn't exist in CF, because there is already a mechanism for bridging stacks and buildpacks

beyhan:
In this RFC we have some open questions and looks to me like we need to discuss those during a meeting. At leats this is my personal opinion. Looking into the schedule for working group updates for the TOC the ARI WG is scheduled at the moment for the 3rd of October. If you think that we need a live discussion would that be an option? Or are there any other suggestions?

Gerg:
I need to do a deeper reading of the CNB platform specification, which I haven't had a chance to yet. I'm going on vacation for a couple weeks, but I hope to re-engage with this RFC when I return. 

At that point @ryanmoran may be back from leave as well, and he may have some to contribute.

beyhan:
The current scheduled for the ARI WG update during the [TOC meeting](https://www.cloudfoundry.org/community-calendar/) is 26th of September. @Gerg I think it makes sense to add this RFC as discussion point for that meeting. Do you think it is a good idea? If yes, we can announce it here, so that who ever is interested can join the discussion.

Gerg:
I'm going to retract this RFC PR for now. I appreciate everyone who has contributed to it up to this point and provided sharp and thoughtful feedback. To move this effort forward, I think it will require a substantial re-working of the proposal, and this area unfortunately hasn't been a focus for me recently.

If anyone is interested in trying out Paketo stacks on CF, I believe there are sufficient hooks in place today to add a custom Paketo Stack, even if it isn't formally integrated into CF Deployment. If anyone is interested, I'd be happy to chat more about this on Slack.

If there is continued interest in this area, I hope to eventually submit a successor RFC that better integrates with the CNB specification, or at least justifies intentional non-integration with the spec.

PR #640, opened 2023-07-06
RFC for integrating pcap-release with BOSH

domdom82:
This PR adds an RFC to propose integration of [pcap-release](https://github.com/cloudfoundry/pcap-release) with BOSH.

For easier viewing: https://github.com/domdom82/community/blob/rfc-bosh-pcap/toc/rfc/rfc-draft-pcap-bosh.md


toc/rfc/rfc-draft-pcap-bosh.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Integrate PCAP CLI as a "pcap" keyword into BOSH CLI
+
+- This would allow PCAP BOSH CLI to share context with BOSH CLI and have access to the user's login session as well as the currently used BOSH environment.
+
+- Another benefit would be that users would not have to download the PCAP BOSH CLI separately. 
+
+### Co-locate PCAP AGENT with BOSH AGENT
+
+- This would make network capture a "built-in" feature of BOSH. It removes the need to manually add PCAP AGENT as a job to deployment manifests or configure it as an addon.
```
maxmoehl, line 53: On agent side the proposal is to have the `pcap` functionality integrated directly into the BOSH agent by importing our library. Unless there is a compelling reason not to do so?

toc/rfc/rfc-draft-pcap-bosh.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Co-locate PCAP API with BOSH DIRECTOR
+
+- This would make it very easy for PCAP CLI to find the right endpoint of PCAP API (it would be equal to the current `BOSH_ENVIRONMENT`)
```
maxmoehl, line 57: `BOSH_ENVIRONMENT` is just the alias to look for in the BOSH config. The availability of PCAP API would probably be announced via the `/info` endpoint which is already used to point to services adjacent to the BOSH director.

toc/rfc/rfc-draft-pcap-bosh.md
```diff
@@ -0,0 +1,59 @@
...
+- No route-registrar or gorouter would be needed to announce any routes. The endpoint would be implicitly known via the BOSH CLI.
```
maxmoehl, line 58: It is not strictly required right now either, you could directly access the API VM. But yes, gorouter would then be completely out of the game.

toc/rfc/rfc-draft-pcap-bosh.md
```diff
@@ -0,0 +1,59 @@
...
+- Possible caveat: Large network dumps could strangle network bandwidth on BOSH DIRECTOR. Therefore an option to have several PCAP API nodes (load-balanced via BOSH DNS) deployed separately may be considered as an alternative.
```
maxmoehl, line 59:
```suggestion
- Large network dumps could exhaust resources on the BOSH director which wouldn't be a concern with a dedicated, horizontally scalable deployment.
```

toc/rfc/rfc-draft-pcap-bosh.md
domdom82: Fair point, I didn't know they rewrote it from ruby to go already. I will rephrase to "integrate with BOSH AGENT".
The details of the integration (submodule or not etc.) will be discussed with the community. We only need to maintain the ability to also have pcap-agent source on a separate repo so that it may be integrated with CF also.

Who knows, could also become a separate "pcap-bosh-agent" if integrating while retaining original sources is not a feasible option.

maxmoehl: We will not split the source. All go code related to pcap will remain in a single repository (and go module), in this case we would probably just have a dependency to that module from the bosh-agent.

toc/rfc/rfc-draft-pcap-bosh.md
domdom82: true. I'd leave it in though to describe the difference of how we would likely deploy today (in conjunction with CF).

toc/rfc/rfc-draft-pcap-bosh.md
domdom82: I disagree with this suggestion. The goal is to co-locate with director, which would be the most convenient. There is a downside to that convenience and there should be an option to remove that downside. Your suggestion, as it reads, implies that there will be a dedicated deployment to avoid resource exhaustion. That contradicts the overall proposal of "co-locating with BOSH director". My original sentence wanted to retain both options with the co-location as the default.

toc/rfc/rfc-draft-pcap-bosh.md
domdom82: Agreed. Changed draft accordingly.

toc/rfc/rfc-draft-pcap-bosh.md
domdom82: @domdom82 what about
```
Large network dumps could exhaust resources on the BOSH director, therefore an option for a dedicated, horizontally scalable deployment should exist.
```
?

toc/rfc/rfc-draft-pcap-bosh.md
```diff
@@ -0,0 +1,59 @@
...
+
+- This would allow PCAP BOSH CLI to share context with BOSH CLI and have access to the user's login session as well as the currently used BOSH environment.
+
+- Another benefit would be that users would not have to download the PCAP BOSH CLI separately. 
+
+### Integrate PCAP AGENT with BOSH AGENT
+
+- This would make network capture a "built-in" feature of BOSH. It removes the need to manually add PCAP AGENT as a job to deployment manifests or configure it as an addon.
+
+### Co-locate PCAP API with BOSH DIRECTOR
```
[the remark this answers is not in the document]
domdom82: I'd say we should mention this as a hook for additional discussions with the community. That's why I would like to include the option in the RFC so that it won't be dismissed later. The goal of the RFC should be to trigger a discussion and address possible options and concerns, not provide a finite solution. 
I think the "director bottleneck" caveat will be definitively discussed. If you can think of more aspects that could need addressing, feel free to add them as you see fit.

toc/rfc/rfc-draft-pcap-bosh.md
```diff
@@ -0,0 +1,59 @@
...
+
+- This would make it very easy for PCAP CLI to find the right endpoint of PCAP API, likely via the `/info` endpoint of BOSH DIRECTOR.
+- No route-registrar or gorouter would be needed to announce any routes. The endpoint would be implicitly known via the BOSH CLI.
+- Large network dumps could exhaust resources on the BOSH director, therefore an option for a dedicated, horizontally scalable deployment should exist.
```
plowin, line 59:
```suggestion
- Large network dumps could exhaust resources on the BOSH director and this approach adds a runtime dependency to the singleton BOSH director, therefore an option for a dedicated, horizontally scalable deployment should exist.
```

toc/rfc/rfc-draft-pcap-bosh.md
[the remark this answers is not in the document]
plowin: Added a proposal to include the "singleton" aspect to the existing caveat

domdom82:
@beyhan we're planning to set this RFC to "ready for review" by EOB July 12. If possible, please review / add your thoughts by then. Thanks!

toc/rfc/rfc-draft-pcap-bosh.md
```diff
@@ -0,0 +1,59 @@
...
+- Status: Draft
+- RFC Pull Request: https://github.com/cloudfoundry/community/pull/640
+
+
+## Summary
+
+Members of the [App Runtime Platform WG](https://github.com/cloudfoundry/community/blob/main/toc/working-groups/app-runtime-platform.md) are developing a BOSH release called [pcap-release](https://github.com/cloudfoundry/pcap-release) which adds network traffic capturing tools to both BOSH deployments and Cloud Foundry applications.
+The release was created after an initial discussion was started on the [cf-deployment](https://github.com/cloudfoundry/cf-deployment/issues/980) repository.
+
+Since then, the project has evolved into a BOSH use case and a Cloud Foundry use case. This RFC discusses a proposal to get the BOSH use case integrated with BOSH director and BOSH cli.
```
peanball, line 15:
```suggestion
Since then, the project has evolved into a BOSH use case and a Cloud Foundry use case. This RFC discusses a proposal to get the BOSH use case integrated with BOSH director, BOSH agent and BOSH cli.
```

toc/rfc/rfc-draft-pcap-bosh.md
```diff
@@ -0,0 +1,59 @@
...
+The main goal of the pcap feature is to allow users (BOSH operator in this case) to easily capture network traffic across multiple VMs by specifying various parameters in PCAP CLI:
+- BOSH deployment
+- BOSH instance group(s)
+- BPF [Berkeley Packet Filter](https://www.tcpdump.org/manpages/pcap-filter.7.html) filter to apply
+
+PCAP CLI will form a capture request and send it to PCAP API. The capture will then take place in real-time similar to running [tcpdump](https://www.tcpdump.org/manpages/tcpdump.1.html) in a distributed fashion across multiple machines. The PCAP API will contact PCAP AGENT on all nodes involved and trigger a network capture. The PCAP AGENTs will concurrently stream all captured traffic back to PCAP AGENT where all streams are merged into one contiguous stream that is sent back to the PCAP CLI where it is then stored as a PCAP FILE on disk.
+
+Trust is established as follows:
+1. PCAP CLI uses the BOSH CLI `access_token` and sends them to PCAP API
+2. PCAP API verifies `access_token` against BOSH UAA and establishes mTLS with PCAP AGENT
```
peanball, line 26:
```suggestion
2. PCAP API verifies `access_token` against BOSH UAA, verifies the token scope and establishes mTLS with PCAP AGENT
```

toc/rfc/rfc-draft-pcap-bosh.md
```diff
@@ -0,0 +1,59 @@
...
+## Problem
+
+The current architecture for "PCAP for BOSH" is depicted as follows:
+
+![PCAP BOSH Architecture](rfc-draft-pcap-bosh/tcpdump-for-bosh.svg)
+
+
+There are several flaws with this architecture that are the result of missing integration:
+1. The PCAP BOSH CLI does not share context with BOSH CLI (because unlike CF CLI, BOSH CLI has no plugin mechanism).
+2. PCAP BOSH CLI reads the user's `access_token` and `refresh_token`, talk to BOSH UAA to get a new `access_token` using the `refresh_token` and write both back to `.bosh/config` - a file which is also accessed by BOSH CLI.
```
peanball, line 38:
```suggestion
2. PCAP BOSH CLI reads the user's `access_token` and `refresh_token`, talks to BOSH UAA to get a new `access_token` using the `refresh_token` and writes both back to `.bosh/config` - a file which is also accessed by BOSH CLI.
```

toc/rfc/rfc-draft-pcap-bosh.md
```diff
@@ -0,0 +1,59 @@
...
+
+Since then, the project has evolved into a BOSH use case and a Cloud Foundry use case. This RFC discusses a proposal to get the BOSH use case integrated with BOSH director and BOSH cli.
+
+The main goal of the pcap feature is to allow users (BOSH operator in this case) to easily capture network traffic across multiple VMs by specifying various parameters in PCAP CLI:
+- BOSH deployment
+- BOSH instance group(s)
+- BPF [Berkeley Packet Filter](https://www.tcpdump.org/manpages/pcap-filter.7.html) filter to apply
+
+PCAP CLI will form a capture request and send it to PCAP API. The capture will then take place in real-time similar to running [tcpdump](https://www.tcpdump.org/manpages/tcpdump.1.html) in a distributed fashion across multiple machines. The PCAP API will contact PCAP AGENT on all nodes involved and trigger a network capture. The PCAP AGENTs will concurrently stream all captured traffic back to PCAP AGENT where all streams are merged into one contiguous stream that is sent back to the PCAP CLI where it is then stored as a PCAP FILE on disk.
+
```
peanball, line 23: I think it's important to mention that we _have to_ use HTTP/2 listeners. Consequentially, the integration would need to support those as well.
```suggestion

All communication between PCAP CLI, PCAP API and PCAP AGENT(s) is carried out via gRPC over HTTP/2.

```

toc/rfc/rfc-draft-pcap-bosh.md
```diff
@@ -0,0 +1,59 @@
+# Meta
+[meta]: #meta
+- Name: Integrate pcap feature into BOSH
+- Start Date: 2023-07-06
+- Author(s): @domdom82 @maxmoehl
```
peanball, line 5:
```suggestion
- Author(s): [@domdom82](https://github.com/domdom82)  [@maxmoehl](https://github.com/maxmoehl)
```

domdom82: I don't think those are supposed to be links. Userhandles should be fine. (double check other RFCs, e.g. [here](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0010-stemcell-cleanup.md) )

maxmoehl:
@domdom82 I think we can set it to ready for review now?

domdom82:
Thanks for everyone's review and comments! Setting to ready for review now.

domdom82:
@cloudfoundry/toc Here is the PR to discuss integrating the pcap functionality with BOSH.
The proposal was originally discussed with @rkoster on CF Day 2023. Feel free to chime in and discuss!

jpalermo:
It feels confusing to have the "architecture we're not going to be using" as the first thing you find in the RFC.

Feels like "We want to build this" would be a better structure, and maybe talk in some footnotes about "we tried this, and it was painful because of X,Y,Z", but I'm also fine if you just left all the history out since the new plan just seems fundamentally cleaner and easier.

domdom82:
@jpalermo point taken. The diagram shows the "as is" state today. It should be noted that this way works but had some inconvenience to it and the integration with CF won't go away as an option. Your point is valid though, we will provide another diagram that shows how (we think) an integration with BOSH could look like.

rkoster:
I agree with @jpalermo about the need to focus on the proposed solution, not on the current non-integrated architecture (which is confusing for people who do not already have context).

While reading through the proposal I also looked at a similar use case that already exists in the bosh cli (log streaming) which apparently is implemented over ssh: https://github.com/cloudfoundry/bosh-cli/blob/main/cmd/logs.go#L104

Would streaming over ssh and deduplicating in the bosh cli (so integrating a bit of the server-side component in the cli) be sufficient? I can see the following benefits which this approach:
1. Only need to touch the cli, people would be able to use the functionality in older environments by just updating the cli. 
2. Heavy lifting will be performed by the machine executing the cli instead of the bosh director
3. No need to open up additional ports (ssh from director to vms, when using director as gateway should already work in most environments)

If plain tcpdump would not be enough, we could consider adding a pcap binary to the stemcells, but this will mean people will have to update their stemcell to be able to use this functionality.

peanball:
@maxmoehl @rkoster, this is an interesting idea. We have some unique features in the pcap-api (central component) that weren't described in the RFC yet. We may need to flesh out the scenarios and error resilience in more detail.

Using gRPC with bidirectional streams allows, from my point of view, unique features that shine with the central dedicated component (i.e. pcap-api):
1. Two-way communication via gRPC streams for **clean shutdown of a capture**, flushing all data captured and buffered up until that point.
2. **Congestion control**. When streaming data to a pcap-client, pcap-api uses a limited buffer towards each of the clients. When the buffer runs full, it can shed load (drop packets) and inform the client that it did so. This is an indication for the user to e.g. refine the filter.  
   The same happens between pcap-agent and pcap-api. When a single agent generates more traffic than it can send to the pcap-api (or the pcap-api can accept), it also sheds packets and sends a message that it did so. This message is passed on to the client as well.
3. **Capture Stop on Disconnect**. When a client disappears, pcap-api will actively stop ongoing captures destined for that client on the pcap-agents.

Independent of, but supported by, gRPC we also have concurrent capture limitations per client IP address. The limit is configurable but ultimately should avoid abusing the feature, whether accidentally or on purpose.

A final word on using SSH: The log forwarder seems to be a reasonably simple case of "collect the logs and send them on" without any interactivity. From my point of view this could be limiting for the pcap-release.

Today we already have some scripts that allow automating calls to tcpdump on BOSH VMs and streaming via SSH. This is not particularly reliable, error prone and hard to control. Based on the experiences with those scripts we set out to design pcap-release, as it addresses many of those issues.

I agree with @maxmoehl's concern that adding an entirely different implementation for data exchange, i.e. replace the gRPC mechanism with SSH or rather _add_ SSH in addition to gRPC, will be difficult to maintain.

I also agree that we should not consider our current implementation for this RFC, but we should consider the anticipated use case for CF and the potential for reuse between those two. A single code base that is maintained, debugged and improved will benefit both use cases at the same time, instead of spreading developers even thinner on supporting two almost disparate code bases.

maxmoehl:
> Today we already have some scripts that allow automating calls to tcpdump on BOSH VMs and streaming via SSH. This is not particularly reliable, error prone and hard to control. Based on the experiences with those scripts we set out to design pcap-release, as it addresses many of those issues.

@peanball I would blame that mainly on the fact that those are shell scripts which are lacking tests, proper error handling and rely on some weird hacks that would not be needed if this were implemented in a language like Go.

domdom82:
@rkoster streaming via SSH technically works but it's not very performant. The main issue is that SSH is made for text and pcap is a binary stream. There are ways to encode pcap as hex-encoded text in tcpdump (-xx flag) but you'll lose the timestamp format and it's no fun parsing. 
In fact, we have such a solution in place right now, but people ended up running tcpdump remotely and storing the capture file in /tmp and then download all of the pcaps via scp and then merge them using mergecap. Oh well.. it "works" 😄 

The idea to do the deduplication on the client and avoid streaming via the BOSH director / API sitting on it seems appealing, though it would mean a significant deviation from the CF case where this won't be possible for security reasons. So we would probably end up with a different code base for BOSH vs. CF for pcap.

Further down the road, there will be discussions about bandwidth usage. When you capture a deployment with lots of instances with a coarse grained filter and try to stream all of that back to the client, you'll run into a congestion problem.
For that scenario, we've thought about a second use case for that in-between API node:
- The PCAP agents would not stream large network dumps directly back to the client.
- Instead they stream to a cloud storage like S3 and return the object id to the API
- During streaming, the client only sees meta data (e.g. how many packets have been captured, how much data has been transferred etc.)
- When the client stops the capture, the API will make sure to collect all of the S3 objects related to this capture, download them and stream them back to the client as one pcap file URL that the client can download.
- The API is also responsible for security: To make sure the client only downloads their captures, not yours.

You could say all of that could also be done on the CLI, true. But the security shouldn't sit on the client side imo. A lot of server-side logic would be required on the CLI to make this work.

plowin:
Hi all, a verbal explanation and demo of the current version was just advertised by @ramiyengar, feel free to listen in tomorrow!
> Link to stream: [https://www.cloudfoundry.org/event/tcpdump-for-everyone/](https://eur03.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.cloudfoundry.org%2Fevent%2Ftcpdump-for-everyone%2F&data=05%7C01%7Cpatrick.lowin%40sap.com%7Cfecca3792ee84f3bb3eb08db88520dce%7C42f7676cf455423c82f6dc2d99791af7%7C0%7C0%7C638253658222753757%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=ThQy3FSCJSnC5DgZquk31PY8YOXeg9INmlHCn05Fp7s%3D&reserved=0)
Time: 4 pm CEST/3 pm London/10 am Eastern
Date: 20th July 2023

maxmoehl:
@rkoster @domdom82 when this was initially raised I wanted to know how much effort this would be, turns out the effort is fairly minimal as you can see in [this gist](https://gist.github.com/maxmoehl/d71acb80c5f8c90e0d90ef32113d1505). Sure we still need to find the right VMs and do the merging but we've implemented those things one way or another already.

beyhan:
Going over all the discussions and the summary provided by @domdom82  my personal favorite is the `ssh` approach because:
- we can address basic needs for `tcpdump` with a simple solution
- we can collect feedback and based on that decide for further steps
- working with bosh for some time now I personally found my way to debug network issues with the available BOSH commands but I can see use cases where a `tcpdump` could be helpful
- I think that the CF app developer use case for `tcpdump` and the BOSH operator use case for `tcpdumps` are different. 

In this [comment](https://github.com/cloudfoundry/community/pull/640#issuecomment-1641563630) @domdom82 raised some security concerns. I don't think that they are relevant for the `bosh ssh` approach. @domdom82 is that correct?

rkoster:
@maxmoehl could you update to RFC to reflect the `bosh ssh` based direction? If you also agree with this direction (I'm assuming so based on the fact that you added a 👍 to Beyhan's comment above).

maxmoehl:
> @maxmoehl could you update to RFC to reflect the `bosh ssh` based direction? If you also agree with this direction (I'm assuming so based on the fact that you added a 👍 to Beyhan's comment above).

We are currently working on an updated version of the RFC and will share that soon.

peanball:
@rkoster, we have an update almost ready. It's being reviewed internally. Should be able to post it soon.

peanball:
@rkoster it's addressed now in the latest state.

The RFC text changed significantly in terms of structuring, so please have another look. Thanks!

beyhan:
@peanball thanks for the update! I read the latest version and I don't have any requests for changes. I added this RFC as a discussion topic to our [meeting notes](https://docs.google.com/document/d/1pLKTANki82JsCqPr4DE8vu4je_omA20eH96CUuRW_8Q/edit#heading=h.hey8qvaww5e6) for the next FI WG meeting this Thursday 3th of August. If any of you would like to participate in the discussions please join our FI WG meeting. The details are available in the [CF community](https://www.cloudfoundry.org/community-calendar/) calendar.

domdom82:
> In this [comment](https://github.com/cloudfoundry/community/pull/640#issuecomment-1641563630) @domdom82 raised some security concerns. I don't think that they are relevant for the `bosh ssh` approach. @domdom82 is that correct?

@beyhan yes, that's correct. In the "streaming only" scenario there is no security concern, because the pcaps are not stored elsewhere. That would only occur if the pcap was first buffered on an external datastore.

domdom82:
@beyhan Motion to trigger Final Comment Period (FCP) for this RFC. I think we have reached the end of discussions for now.

beyhan:
FCP should end on 15.08.2023

rkoster (approved):

stephanme (approved):

ameowlia (approved):

ChrisMcGowan (approved):

PR #641, opened 2023-07-07
Add draft RFC for aggregate metric egress with OTel Collector

ctlong:
For easier viewing: https://github.com/loggregator/community/blob/rfc-otel-collector/toc/rfc/rfc-draft-aggregate-metric-egress-with-opentelemetry-collector.md

cc @mkocher, @acrmp, @rroberts2222

### DIFF
--- a/toc/rfc/rfc-draft-aggregate-metric-egress-with-opentelemetry-collector.md
+++ b/toc/rfc/rfc-draft-aggregate-metric-egress-with-opentelemetry-collector.md
@@ -21,7 +21,7 @@ Our agent suite currently provides metrics via a prometheus endpoint. However us
 
 ## Proposal
 
-Diagram link: https://drive.google.com/file/d/193khcTpwObgWMcNt5VOZ68huezp3FLB6/view?usp=sharing
+Diagram link: https://drive.google.com/file/d/1-cuOwPdKokeeNvUoZ4bpcbh60Vv5J5JX/view?usp=sharing
 
 ### OpenTelemetry Collector Add-On
 
@@ -40,9 +40,11 @@ exporters:
       endpoint: otelcol2:4317
 ```
 
-### Metrics Agent
+The BOSH job would register itself with the Forwarder Agent to receive OTLP metrics.
 
-Update the Metrics Agent “/metrics” endpoint to serve all metrics, rather than just metrics from components that do not have a Prom Scraper configuration file.
+### Forwarder Agent
+
+Update the Forwarder Agent to support forwarding metrics to the OpenTelemetry Collector via OTLP.
 
 ### Option to Shut Down the Firehose
 
--- a/toc/rfc/rfc-draft-aggregate-metric-egress-with-opentelemetry-collector.md
+++ b/toc/rfc/rfc-draft-aggregate-metric-egress-with-opentelemetry-collector.md
@@ -1,0 +1,69 @@
+# Meta
+[meta]: #meta
+- Name: Add OTel Collector for aggregate metric egress
+- Start Date: 2023-07-07
+- Author(s): @ctlong, @mkocher, @acrmp, @rroberts2222
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (fill in with PR link after you submit it)
+
+
+## Summary
+
+We propose adding an [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/) to every CF-D VM to forward all app and platform metrics on those VMs to external drains. Operators would be able to choose between different protocols and settings provided by the Collector out-of-the-box when inputting their metric destinations.
+
+We believe the Collector deployed in this manner would not only solve our current problem. In the long run we could leverage it for other interesting new functionality with logs and traces, as well as reducing our agent suite footprint.
+
+## Problem
+
+Many operators and Cloud Foundry maintainers have problems with the firehose. The firehose requires two sets of coordinating vms, which is inherently unscalable, and a Cloud Foundry-specific nozzle translation layer, which no one wants to maintain. We would like to retire the firehose, and to that end built out the syslog drain system. However, we continue to see operators stay on the firehose, using barely supported nozzles. In order to retire the firehose completely, operators need a way to egress metrics, as well as logs, off the platform to their monitoring tools of choice.
+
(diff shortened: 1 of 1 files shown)


chombium:
I really like and support the idea of adding OpenTelemetry to Cloud Foundry, but I find the motivation and the problem statement a bit misleading.

Yes, the Firehose architecture has its own disadvantages, because of the aggregation components (Dopplers and RLPs), the effects of misbehaving noisy applications on the flow of log envelopes and some unmaintained plugins and libraries, but it supports collection and forwarding of logs and metrics. As we all work with the `Never touch a running system` motto it is hard to believe that someone would do something if we don't say that we will deprecate and remove the Firehose.

On the other side, the Syslog drains require some more effort to do the implementation right in case  someone wants to use the Syslog protocol and the HTTP based Syslog drains does not support high throughput rates, but they also forward logs and metrics.

By using OpenTelemetry we could solve these problems, but I'm wondering few things:
- Why does this RFC cover only forwarding of metrics? Would this be an initial setup/implementation only for metrics as there are existing out-of-the-box Open Telemetry Collector [receivers](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver) and [exporters](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter) which will be used so we can start experimenting with OpenTelemetry and we will add support for logs and Zipkin traces later? The OpenTelemetry Collector, the way it is defined in this proposal cannot be a replacement neither for the Firehose nor for the Syslog Drains.

- Why is the existence of multiple Open Telemetry Collector exporters (the ability to export the data in different formats) promoted as main selling point which will help the metric consumers? IMO this is a positive side effect of using OpenTelemetry.

OpenTelemetry's main goal is to standardize instrumentation tools for applications and combine logs, metrics and traces in order to provide better observability. I think that it would be better instead of hooking the receiver to the `metrics-agent`, that we build a custom receiver which will be connected directly to the `forwarder agent` and have application and platform component metrics and application logs at hand. We could even do some changes to the [syslog-release](https://github.com/cloudfoundry/syslog-release) to forward the platform component logs to the collector. This way we could build a custom processor, correlate all logs and metrics, enrich them with some spans and extra attributes. That would be essential help for the Cloud Foundry operators as they will have all connected data at hand to quickly see what's going on and immediately take the actions needed.

- There are some planed changes to the `metrics-agent`. We should be careful, not to break anything and stay backward compatible.


Generally speaking, I think that this RFC is a good start to introduce OpenTelemetry to Cloud Foundry, but there will be long way to go until an OpenTelemetry based solution can replace the Firehose or even the Syslog drains.

ctlong:
Hey @chombium, thanks for the detailed response! Let me address it in parts.

> Why does this RFC cover only forwarding of metrics? Would this be an initial setup/implementation only for metrics as there are existing out-of-the-box Open Telemetry Collector [receivers](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver) and [exporters](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter) which will be used so we can start experimenting with OpenTelemetry and we will add support for logs and Zipkin traces later?

I think you're correct in your assessment. In the short-term, we think it's the combination of the OpenTelemetry Collector (for metrics) and Syslog Drains (for logs) that represents a realistic replacement for the Firehose. And experimenting with the OpenTelemetry Collector now helps set us up for a potential long-term future where we could start using the Collector for logs and traces as well.

> Why is the existence of multiple Open Telemetry Collector exporters (the ability to export the data in different formats) promoted as main selling point which will help the metric consumers? IMO this is a positive side effect of using OpenTelemetry.

I assume there's a "not" missing in here somewhere, which I hope is in the first sentence 😁. I do think that the existence of many OpenTelemetry exporters is a strong reason to include it, and if we can rephrase something in our solution statement to better encapsulate that please let me know what you'd change.

> I think that it would be better instead of hooking the receiver to the metrics-agent, that we build a custom receiver which will be connected directly to the forwarder agent and have application and platform component metrics and application logs at hand... There are some planed changes to the metrics-agent. We should be careful, not to break anything and stay backward compatible.

We think so too! Utilizing the Metrics Agent was our preferred approach initially because it seemed faster. We wouldn't have to build OTLP functionality into Forwarder Agent, or a loggregator envelope receiver into the OpenTelemetry Collector. However, we've actually been backtracking on that approach in favor of one of the other two options already, so we're open to reevaluating the steel thread of the proposal to not include Metrics Agent at all.

I'll call out that we explored the codebase and couldn't find a reason for why the Metrics Agent doesn't serve all metrics on the `/metrics` endpoint, and our thinking was that might have been an oversight that was never recognized due to lack of use of the component.

chombium (approved):

beyhan:
Final period for comment ends 25th of July.

chombium:
@ctlong I have few questions regarding the internal Syslog aggregate drain to Log Cache which we have at the moment: 
 - At the moment Log Cache consumes the app logs, application container metrics and platform component metrics via a Syslog aggregate drain. 
 - The aggregate drains are defined in the cf-deployment 
 
As we want to enable more different options to consume the metrics (and eventually logs) in short term and improve the CF's observability in long term on one hand, and on the other hand, we want to provide the Syslog Drain + OTel Collector combination as an alternative to the Firehose, what would happen with the internal Syslog drain to Log Cache when we activate the OpenTelemetry Collector? Will this change affect the aggregate Syslog drains in any way? 

I assume the newly planned OTel Collector targets only external metrics (and later logs) consumers, right?

How will the CF users configure the new metric aggregate egress with OTel Collector?

ctlong:
@acrmp and I just updated the RFC to remove Metrics Agent changes, for reasons discussed in previous comments. Instead the Forwarder Agent can be altered in a backwards compatible way to forward metrics to the OTel Collector.

stephanme (approved):

rkoster (approved):

beyhan:
For the records:
The FCP ended on 25th of July 2023. During the TOC meeting on that day we agreed that there are no open discussions in this RFC and it can be approved. Additionally, we decided that the introduced change described in [this comment](https://github.com/cloudfoundry/community/pull/641#issuecomment-1648599488) during the FCP period doesn't have such an impact that we should exend the FCP period. Additionally, it was even mentioned as option in the initial version of this RFC.

PR #643, opened 2023-07-10
Archive resolvconf-manager projects

ystros:
When developing the Bionic stemcell, the team originally planned to introduce a wrapper around managing resolv.conf within the agent and BOSH DNS. However, this work was reverted (likely due to backward compatibility concerns) and abandoned. This idea was not incorporated at all into the Jammy stemcells.

As such, we should not continue actively maintaining the associated repositories.

Original implementation story:
https://www.pivotaltracker.com/story/show/167763426

Revert: 
https://www.pivotaltracker.com/n/projects/2238419/stories/168790375 
https://github.com/cloudfoundry/bosh-dns-release/commit/aac14421fd63aa7e266f93cf16d614ec6fce60d4

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -1966,8 +1966,10 @@ orgs:
         has_projects: true
       resolvconf-manager:
         has_projects: true
+        archived: true
       resolvconf-manager-index:
         has_projects: true
+        archived: true
       route-emitter:
         allow_merge_commit: false
         description: Registers and unregisters processes running on executors with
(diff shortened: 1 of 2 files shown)


rkoster (changes requested):
Could you also remove both repos from the foundational infrastructure working group charter: https://github.com/cloudfoundry/community/blob/main/toc/working-groups/foundational-infrastructure.md?

ystros:
Good catch. Updated!

rkoster (approved):

beyhan (approved):

stephanme (approved):

ameowlia (approved):

PR #512, opened 2023-01-18
Consolidate bosh-packages and bosh-io into cloudfoundry github org

rkoster:
This PR is the first step in consolidating the bosh-packages and bosh-io github organizations into the main cloudfoundry org.

After this PR gets approved and merged I would like to propose that repo creation is temporary enable for the cloudfoundry organization so that foundational infrastructure approvers who are also admin in the to be moved repo's can do so at their own pace (there will be some ci fixes that need to be done while moving).

The to-be-moved repos will also be renamed to clarify their scope (for example bosh-io/web will become cloudfoundry/bosh-io-web).

PR #653, opened 2023-07-21
Find the right place for bosh-package-java-release and bosh-package-cf-cli-release

beyhan:
With [Consolidate bosh-packages and bosh-io into cloudfoundry github org](https://github.com/cloudfoundry/community/pull/512) the FI WG got two repositories which don't really belong to the WG. This pr suggests a possible move but our intention is to start a discussion to find right place for them. cc @rkoster & @jpalermo

### DIFF
--- a/toc/working-groups/app-runtime-interfaces.md
+++ b/toc/working-groups/app-runtime-interfaces.md
@@ -117,6 +117,7 @@ areas:
   - cloudfoundry/java-buildpack-system-test
   - cloudfoundry/java-test-applications
   - cloudfoundry/jvmkill
+  - cloudfoundry/bosh-package-java-release
 
 - name: Buildpacks .Net Core
   approvers:
@@ -361,6 +362,7 @@ areas:
   - cloudfoundry/cli-workstation
   - cloudfoundry/cli-private
   - cloudfoundry/cli-pools
+  - cloudfoundry/bosh-package-cf-cli-release
   - cloudfoundry/jsonry
   - cloudfoundry/stack-auditor
   - cloudfoundry/ykk
--- a/toc/working-groups/foundational-infrastructure.md
+++ b/toc/working-groups/foundational-infrastructure.md
@@ -302,9 +302,7 @@ areas:
   - cloudfoundry/bosh-io-worker
   - cloudfoundry/bosh-linux-stemcell-builder
   - cloudfoundry/bosh-openstack-cpi-release
-  - cloudfoundry/bosh-package-cf-cli-release
   - cloudfoundry/bosh-package-golang-release
-  - cloudfoundry/bosh-package-java-release
   - cloudfoundry/bosh-package-nginx-release
   - cloudfoundry/bosh-package-python-release
   - cloudfoundry/bosh-package-ruby-release


rkoster (approved):

jpalermo:
Seems good to me, but it would also be good to have somebody from those WGs weigh in too.

beyhan:
@dmikusa or @anthonydahanne could you please comment regarding the move of `cloudfoundry/bosh-package-java-release`?

@reedr3 or @a-b could you please comment regarding the move of `cloudfoundry/bosh-package-cf-cli-release`?

toc/working-groups/app-runtime-interfaces.md
```diff
@@ -117,6 +117,7 @@ areas:
   - cloudfoundry/java-buildpack-system-test
   - cloudfoundry/java-test-applications
   - cloudfoundry/jvmkill
+  - cloudfoundry/bosh-package-java-release
```
a-b, line 120: I'm not convinced `bosh-package-java-release` belongs to app runtime interfaces.

a-b: `bosh-package-cf-cli-release` LGTM

a-b: Ok, looks like all buildpacks are there, so I guess it's ok.

a-b (approved):

Gerg:
Adding @pivotal-david-osullivan for cloudfoundry/bosh-package-java-release

Gerg:
@beyhan @rkoster What is the rationale for moving `bosh-package-java-release` out of FI, but not the other language packages (`package-ruby-release`, `package-golang-release`, etc)?

rkoster:
The java package according to bosh.io is only being used by the autoscaler: https://github.com/cloudfoundry/app-autoscaler-release/tree/main/packages/openjdk-17

Maybe it would be more appropriate to move it into the Autoscaler area.

Gerg:
Right, but what is the reason for moving some of the `bosh-package-*` releases out of FI? 

Why is `bosh-package-ruby-release` staying in FI? Just because BOSH uses it?

Why is `bosh-package-python-release`, for instance, staying in FI?

What if another release wants to start using `bosh-package-java-release`?

jpalermo:
We certainly know that FI has no expertise with the java package or bosh releases using java or how these packages should be built.

It would be good to have an owner that has a vested interest in it, but maybe that doesn't exist...

rkoster:
The alternative is to archive `bosh-package-java-release`, given that we have no approvers in the FI working group who have context on this release (UAA & Credhub do not use it).

Gerg:
What is the history for why that package was introduced in the first place? Just for language completeness?

beyhan:
I don't the history. @jpalermo, @rkoster do you know?

jpalermo:
Don't have the full history, but based on some old tracker stories (with virtually zero details), the commit history, and some wild speculation; my guess would be it was created because we were doing it for other languages and we had at least one bosh release using Java.

Gerg:
> we had at least one bosh release using Java

Which release was it?

jpalermo:
Looks like UAA/Credhub have never used it. It was used by a closed source release from 2018 to 2020, and then app autoscaler has been using it since 2022.

2018 isn't TOO far off from when it was created, but it was ~9 months after.

So my guess is it was created because we had some Java releases, primarily UAA/Credhub, and they never adopted it.

beyhan:
I would suggest then to archive the `bosh-package-java-release`.

jpalermo:
If we did want to archive it, we'd need to give [app-autoscaler-release](https://github.com/cloudfoundry/app-autoscaler-release) a chance to move off of it first.

Gerg:
Should we consider transferring ownership to the [ARI Autoscaler](https://github.com/cloudfoundry/community/blob/6bd1a5a9a916be6e2f4dbfb300d0edaa8226e03a/toc/working-groups/app-runtime-interfaces.md#L51-L70) area then?

beyhan:
Hi @silvestre,

With the implementation of https://github.com/cloudfoundry/community/pull/512 the FI WG got the [bosh-package-java-release](https://github.com/cloudfoundry/bosh-package-java-release) repository which is not used in the FI WG and we don't have expertise to maintain this repository. It looks like the ARI Autoscaler area is the only use of this repository. Do want to take ownership of this repository? The alternative is to archive it.

silvestre:
Hi @beyhan, @Gerg,

thanks for reaching out! 

As `app-autoscaler-release` is the only user, I don't think it makes sense for us to maintain this but we will switch back to packaging Java ourselves.

Feel free to archive it at any time.

beyhan:
Thanks @silvestre for the response. Based on the discussion I open https://github.com/cloudfoundry/community/pull/687 to archive `bosh-package-java-release`. I will close this pr because it includes more things than just the `bosh-package-java-release`.

PR #654, opened 2023-07-21
Add Runtime Platform repos to App Runtime platform Github project

MarcPaquette:
The Project (41) is private at this point. I would also need to add the bot to the project so it has permissions.


rkoster:
There is no bot, automation is using GitHub app auth. We will figure if additional permissions for private projects are needed, but given that this is managed at org level it should be easy to fix.

rkoster (approved):

beyhan (approved):

stephanme (approved):

ChrisMcGowan (approved):

PR #656, opened 2023-07-24
Update branch protection for the CF-on-K8s repositories

gcapizzi:
This restores the branch protection rules the WG had before turning on `config.generate_rfc0015_branch_protection_rules`. In particular:
* slightly stricter branch protection the `korifi` repository, requiring checks to have passed before merging;
* no protection on `korifi-ci` and `cf-k8s-secrets`.

Not sure if this makes `config.generate_rfc0015_branch_protection_rules` needed at all.

### DIFF
--- a/org/branchprotection.yml
+++ b/org/branchprotection.yml
@@ -5,7 +5,7 @@ branch-protection:
     cloudfoundry:
       repos:
         # example repo configured as if the project would belong to ARD WG, area CF Deployment
-        # automation would generate repo/branch config for all repos belonging to a WG unless an explicit configuration exists here 
+        # automation would generate repo/branch config for all repos belonging to a WG unless an explicit configuration exists here
         branchprotection-test:
           protect: true
           enforce_admins: true
@@ -58,7 +58,7 @@ branch-protection:
           allow_deletions: false
           allow_disabled_policies: true
           include: ["^master$"]
-          required_status_checks: 
+          required_status_checks:
             contexts: ["concourse-ci/unit-tests","concourse-ci/acceptance-tests"]
           required_pull_request_reviews:
             dismiss_stale_reviews: true
@@ -73,7 +73,7 @@ branch-protection:
           allow_deletions: false
           allow_disabled_policies: true
           include: ["^main$"]
-          required_status_checks: 
+          required_status_checks:
             contexts: ["concourse-ci/unit-tests","concourse-ci/acceptance-tests"]
           required_pull_request_reviews:
             dismiss_stale_reviews: true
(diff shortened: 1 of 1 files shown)


org/branchprotection.yml
```diff
@@ -81,3 +81,32 @@ branch-protection:
...
+          protect: false
+        korifi:
+          protect: true
+          enforce_admins: true
+          allow_force_pushes: false
+          allow_deletions: false
+          allow_disabled_policies: true
+          include: ["^main$"]
+          required_status_checks: 
+            checks:
```
stephanme, line 98: The syntax of required_status_checks is not correct.
Doc: https://docs.prow.k8s.io/docs/components/optional/branchprotector/ 
See e.g. configuration of pcap-release above

org/branchprotection.yml
```diff
@@ -81,3 +81,32 @@ branch-protection:
...
+              - context: "controllers-tests"
+              - context: "linter"
+              - context: "job-task-runner-tests"
+              - context: "kpack-image-builder-tests"
+              - context: "statefulset-runner-tests"
+              - context: "tools-tests"
+              - context: "Concourse / e2e-tests (pull_request)"
+              - context: "Concourse / e2e-eks-tests (pull_request)"
+              - context: "EasyCLA"
+          required_pull_request_reviews:
```
stephanme, line 109: Not sure, but you may want to bypass the PR requirement for your bot users. At least this is how the branchprotection rules would get generated by the automation. Since you overwrite the automation, you need to take care yourself.

stephanme (changes requested):
config.generate_rfc0015_branch_protection_rules is not needed anymore since you configure all your 3 repos explicitly.
You could keep the setting, so that any new repo that you add to your WG gets the branchprotection automatically.

org/branchprotection.yml
gcapizzi: We went for this syntax as the [Github docs](https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2022-11-28#update-branch-protection) mark `contexts` as deprecated - and thought it was ok since the checks passed?

org/branchprotection.yml
gcapizzi: @cloudfoundry/wg-cf-on-k8s what do we think about this? I believe we didn't have the bypass on before switching to the automation, but I'm happy to add it!

georgethebeatle: I am fine either way, but if we have not had to bypass the bots in the past and for the sake of minimalism we do not have to add this rule now. I think korifi bot would open PRs to re-vendor some test dependencies anyway and it is our only bot that commits code as far as I can remember. We can always add the rule in the future should we need to do so.

org/branchprotection.yml
stephanme: I would say that the branchprotector tool still requires the `contexts` node: https://github.com/kubernetes/test-infra/blob/master/prow/github/types.go#L633

The `org-config-generation-check (pull_request)` includes a schema validation for branchprotection.yml but the schema includes only the fields that are needed by the org automation. It doesn't include all the detailed fields supported by the branchprotector tool. This could be improved. Problem is that there is little documentation available and would have to reverse engineer the schema from the branchprotector sources - but I could start with the fields that are used so far in branchprotection.yml and extend case by case.
For now, the final validation comes when this PR got merged...

org/branchprotection.yml
gcapizzi: Let's keep stuff as it is for now - we'll change it if we feel like we need it!

stephanme (approved):

rkoster (approved):

PR #665, opened 2023-08-04
[RFC] User Account Management in Cloud Foundry with Multiple Identity Zones

philippthun:
Link for easier viewing: https://github.com/philippthun/community/blob/rfc-multiple-identity-zones/toc/rfc/rfc-draft-multiple-identity-zones.md

### DIFF
--- a/toc/rfc/rfc-draft-multiple-identity-zones.md
+++ b/toc/rfc/rfc-draft-multiple-identity-zones.md
@@ -4,7 +4,7 @@
 - Start Date: 2023-08-04
 - Authors: @FloThinksPi, @klaus-sap, @philippthun
 - Status: Draft
-- RFC Pull Request: (fill in with PR link after you submit it)
+- RFC Pull Request: https://github.com/cloudfoundry/community/pull/665
 
 ## Summary
 
--- a/toc/rfc/rfc-draft-multiple-identity-zones.md
+++ b/toc/rfc/rfc-draft-multiple-identity-zones.md
@@ -1,0 +1,51 @@
+# Meta
+[meta]: #meta
+- Name: User Account Management in Cloud Foundry with Multiple Identity Zones
+- Start Date: 2023-08-04
+- Authors: @FloThinksPi, @klaus-sap, @philippthun
+- Status: Draft
+- RFC Pull Request: (fill in with PR link after you submit it)
+
+## Summary
+
+This RFC proposes enhancing Cloud Foundry (CF) to support users from multiple identity zones within the CF User Account and Authentication (UAA) system. The objectives are to enable efficient user management, facilitate data isolation and allow customized security measures.
+
+## Problem
+
+CF operators are currently limited to the default identity zone within the UAA system for user account management. However, the need to store user accounts in different zones may arise from two distinct operational requirements:
+1. CF installations servicing multiple customers or tenants (e.g. public PAAS) must ensure absolute isolation of user accounts. Each customer or tenant must have their user accounts securely stored in separate identity zones to prevent unauthorized access to sensitive data. It is crucial that users from one customer cannot view, modify, or access the user accounts of other customers.
+1. Even in cases where CF is deployed for a single customer or organization, there may be a need to employ multiple identity zones. This could arise when customers wish to segregate user accounts based on roles, affiliations, or subsidiaries. By creating distinct zones, organizations can enforce specific security measures tailored to different user groups, such as implementing Multi-Factor Authentication (MFA) for certain privileged users.
+
+## Proposal
+
+To address the operational and security requirements outlined above, this RFC proposes enhancing CF to support users from multiple identity zones within the CF UAA system. This requires changes to various components as outlined in this section.
+
+### User Account and Authentication
+
+1. Currently, web-based applications like service dashboards or other client applications like "cf ssh" run the authorization code flow against the default identity zone of UAA, which is advertised by the foundation’s api root (`/`) and info endpoints. This flow shall be extended to work with user accounts from other identity zones, too.
+   1. The login page of UAA shall allow switching from the default identity zone (empty subdomain) to another identity zone/subdomain. UAA may store the favorite zone/subdomain in a cookie.
+   1. The `/oauth/authorize` endpoint in the default identity zone shall be able to access security sessions from other identity zones and return authorization codes belonging to other identity zones.
+   1. The `/oauth/token` endpoint in the default identity zone shall accept authorization codes belonging to other identity zones and return access tokens from the right zone, i.e. tokens with claim `zid` set to the zone ID of the user who has logged in.
+   1. The `/check_token` endpoint in the default identity zone should be able to verify tokens from all identity zones, i.e. the claim `zid` must be considered to choose the right identity zone for token verification.
+1. With these changes it should be sufficient to create clients with the authorized grant type `authorization_code` in the default zone only. Along with these clients, the OAuth issuer and verification keys of all zones should be taken from the default zone.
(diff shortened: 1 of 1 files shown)


philippthun:
@cloudfoundry/toc - FYI
@cloudfoundry/wg-foundational-infrastructure - Please provide feedback, especially to the UAA related changes.
@cloudfoundry/wg-app-runtime-interfaces - Please provide feedback, especially to the CAPI and CF CLI related changes.

Thanks,
@FloThinksPi, @klaus-sap, Philipp

beyhan:
@philippthun Thank you for the proposed RFC. Could you please add a link for easier viewing like in the other RFCs, e.g. [this one](https://github.com/cloudfoundry/community/pull/640).

peterhaochen47:
Hello, thanks for the proposal. Here are my comments:
## About the problem statements 
> CF installations servicing multiple customers or tenants (e.g. public PAAS) must ensure absolute isolation of user accounts. Each customer or tenant must have their user accounts securely stored in separate identity zones to prevent unauthorized access to sensitive data. It is crucial that users from one customer cannot view, modify, or access the user accounts of other customers.

I believe this problem has a solution currently. Imagine a public PAAS based on CF offered & operated by company Foo-Inc. Only the UAA admin users (users with UAA admin scopes; they should be employees from Foo-Inc, operating the PAAS) have the power to view/modify any other user accounts. Currently, a normal / non-admin user (aka a customer of the PAAS) cannot view, modify, or access the user accounts of other customers. Please provide specific concerns if you believe UAA currently does not guarantee this.

> Even in cases where CF is deployed for a single customer or organization, there may be a need to employ multiple identity zones. This could arise when customers wish to segregate user accounts based on roles, affiliations, or subsidiaries. By creating distinct zones, organizations can enforce specific security measures tailored to different user groups, such as implementing Multi-Factor Authentication (MFA) for certain privileged users.

This is true, currently the log-in config is applied uniformly to all users within an identity zone. However, could you provide more specific examples of these security measures that you want to apply differently to different sets of users? (UAA's native MFA feature is deprecated.) There could be existing solutions already.

## About the proposed changes
- The proposed changes weakens the isolation of identity zones. 
  - Currently, a UAA instance's different identity zones can almost be treated as different instances of UAA (the exception is that the UAA admin users in the default system zone can access & manage all identity zones); the user data/access from different zones do not cross-pollinate. This was the design intention of the identity zone features (so that different tenants can safely share the same UAA without fear of leakage). If anything, one might argue that this isolation is not complete enough; after all, different zones still use the same database tables & are run by the same VM (hence, contains some risks of accidental cross-pollination). For the long term, it's probably more secure to deprecate identity zones all together (and people should deploy new UAA instances if they need another isolated user store).
  - The 4 proposed changes under "item 1" (cross-linking login domains, sharing security sessions, sharing authorization code, allow one zone to check tokens issued by another zones) seem to break down some of the barriers between zones & might result in more concerns for leakage. 
 - For item 3, "In addition, it must not be possible for customers to break out from their identity zone. Customers must not be able to assign authorizations to users, which grant access to other identity zones." I believe this is true today. 
 - For item 4, "With authorization uaa.admin or wildcard scopes across all zones (e.g. zones.*.scim.read) it should be possible to list all users from all zones with one single API call." This is true today (except you have to make multiple API calls, each targeting one zone. Modifying UAA API to allow one request to access all zones's users seems risky in that it further breaks down the barriers between zones.

toc/rfc/rfc-draft-multiple-identity-zones.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Problem
+
+CF operators are currently limited to the default identity zone within the UAA system for user account management. However, the need to store user accounts in different zones may arise from two distinct operational requirements:
+1. CF installations servicing multiple customers or tenants (e.g. public PAAS) must ensure absolute isolation of user accounts. Each customer or tenant must have their user accounts securely stored in separate identity zones to prevent unauthorized access to sensitive data. It is crucial that users from one customer cannot view, modify, or access the user accounts of other customers.
+1. Even in cases where CF is deployed for a single customer or organization, there may be a need to employ multiple identity zones. This could arise when customers wish to segregate user accounts based on roles, affiliations, or subsidiaries. By creating distinct zones, organizations can enforce specific security measures tailored to different user groups, such as implementing Multi-Factor Authentication (MFA) for certain privileged users.
+
```
stephanme, line 18: One additional problem that this RFC tries to solve: Customers want to have full control of their CF users, i.e. create users, assign CF roles, and finally also remove users. It should be possible for customers to implement a central user provisioning based on their needs.

This is not possible with a central UAA using the default identity zone only because we can't grant admin access for the default zone to customers. Admin access for a customer identity zone would solve this.

peterhaochen47: I see. The use case is ultimately about having multiple organizations sharing the same CF foundation, where each of the orgs needs to self-manage their CF users.

torsten-sap: @peterhaochen47 The operational costs itself are imho a showstopper for org-specific UAA instances. Apart from that, the UAA integration with central landscape components (which need access to all UAAs) becomes overly complex and would most probably not scale.

peterhaochen47: @torsten-sap how about this option:
> You might even be able to just add one non-system identity zone as [an external identity provider](https://docs.cloudfoundry.org/api/uaa/version/76.18.0/index.html#create) of the main system UAA zone (on the UAA instance). But I haven't verified that it works (but I don't see why not).

strehle: yes but the resulted user token is then with zid:uaa means that users are within the system (uaa) zone, but we need and what separate users (and other related entities like idps or groups) into an own zones, so that a zone can be configured independently from another zone. That is what we support today in UAA, but not in cloud controller... 

The move of groups and IdPs into the customer zone is not a big challenge, but the oauth clients (e.g. existing dashboards https://docs.cloudfoundry.org/services/dashboard-sso.html ) and their configuration with secrets is something I think about improvements, so that not all dashboard clients need to be copied into each single zone, therefore we want extend the central check_token endpoint.  The details we should discuss in our call

strehle: > It's similar to that when you log into uaa system zone via Azure AD and get a UAA token, the user is still located in Azure AD.

true, but then we have the same as today. What we want have is a separation of users so that different administrators can manage these different users. Then we would need a new administrator, the origin admin who only can manage users and groups for a certain origin.

toc/rfc/rfc-draft-multiple-identity-zones.md
```diff
@@ -0,0 +1,51 @@
...
+1. With authorization `uaa.admin` or wildcard scopes across all zones (e.g. `zones.*.scim.read`) it should be possible to list all users from all zones with one single API call.
+
+### Cloud Controller
+
+1. The Cloud Controller (CC) shall authenticate users with tokens from non-default identity zones. The OAuth issuer and verification keys shall be taken from the default zone.
+1. User lookups must be performed within the identity zone of the authenticated user. SCIM queries shall be done with the `X-Identity-Zone-Id` header set to the `zid` of the user token. When a user interacts with the CC and requests information about other users who are part of their organization, the following behavior shall be implemented:
+   1. For users who reside in the same zone as the authenticated user, the CC shall display their usernames.
+   1. For users who belong to a different zone from the authenticated user, the CC must not display their usernames, but only the user GUIDs.
```
Gerg, line 39: What is a case where a user would be able to see a user from a different zone? If these identity zones are segmented by organization, then I don't think a non-admin user would encounter a user from another zone.

Asking because I'd prefer to avoid adding this complexity to CC, if we can avoid it.

klaus-sap:
Thank you for the detailed feedback. It seems to me that the proposal from this RFC is a bit too complicated. Thus, we would like to propose a simpler approach, see https://github.com/cloudfoundry/uaa/issues/2505

philippthun:
Thanks for all the comments. I'm going to close (withdraw) this RFC.

PR #681, opened 2023-09-12
Promote Cristhian Camilo Peña  as CLI approver

a-b:
Cristian has contributed to CF CLI for many years and is familiar with this project. It would be a good idea to promote him as an approver.


Gerg:
Closing this in favor of https://github.com/cloudfoundry/community/pull/685

PR #683, opened 2023-09-13
Update docs.md

anita-flegg:
Added to the scope to include using inclusive language and maintaining internal consistency of the doc styles

### DIFF
--- a/toc/working-groups/docs.md
+++ b/toc/working-groups/docs.md
@@ -15,8 +15,9 @@ To document the Cloud Foundry user experience.
 ## Scope
 
 - Merge and edit all doc changes
-- Update with the latest best practices for SEO
-- Remove out-dated information
+- Update with the latest best practices for SEO and inclusive language
+- Maintain internal consistency of doc style, including Notes and tables, for example
+- Remove out-dated information (ongoing)
 - Advise contributors on best practices
 
 ## Non-Goals


toc/working-groups/docs.md
```diff
@@ -15,8 +15,9 @@ To document the Cloud Foundry user experience.
 ## Scope
 
 - Merge and edit all doc changes
-- Update with the latest best practices for SEO
-- Remove out-dated information
+- Update with the latest best practices for SEO and inclusive language
+- Maintain internal consistency of doc style, including Notes and tables, for example
+- Remove out-dated information (ongoing)
```
beyhan, line 20: I don't think that we need the status `(ongoing)` to be part of the scope. Most probably this will be a task which is done when some parts of the docs are outdated. @anita-flegg Or do you want to express something else with this?

anita-flegg: Thanks, @beyhan. I added (ongoing) because it's not a one time effort. That may be understood, and I can remove it if you prefer.

beyhan: I got it. Thanks!

beyhan (approved):

beyhan:
I agree with the change and approved in case you want to merge it.

anita-flegg:
Thanks!

PR #694, opened 2023-09-23
Rename repo as we no longer use gke for the concourse

jpalermo:


beyhan (approved):

rkoster (changes requested):
Woudn't `fiwg` or `fi` be more appropriate? Or is this repo still reusable enough to justify using the `cff` suffix?

jpalermo:
Good question. That's something @ramonskie and I were talking about.

It's not currently in that state, and would require some work to make it more generic. Is there a desire for that, or should we make it `fi` specific?

beyhan:
Other WG decided to go with a slightly different approach available in https://github.com/cloudfoundry/app-runtime-interfaces-infrastructure/. I would say that currently there is no desire and we shouldn't do the extra effort. I would vote for `fiwg` in this case.

jpalermo:
PR updated with new name

beyhan (approved):

beyhan:
ping @rkoster

rkoster (approved):

PR #738, opened 2023-12-20
Add Pranay Raj

jbooherl:
Pranay will be taking over metric store product and needs to be added as a cloudfoundry contributor.


beyhan:
Hi @jbooherl,
could you please sign the CLA.

PR #741, opened 2024-01-03
New repo for performance tests BOSH errand

jochenehret:


beyhan:
New repository will contain a BOSH errand which will execute [cf-performance-tests](https://github.com/cloudfoundry/cf-performance-tests).

beyhan (approved):

PR #743, opened 2024-01-08
rfc: generic per-route features

maxmoehl:
🚀 [Link for easy viewing.](https://github.com/sap-contributions/cf-community/blob/rfc-route-features/toc/rfc/rfc-draft-generic-per-route-features.md)

/cc @domdom82 @peanball

### DIFF
--- a/toc/rfc/rfc-draft-generic-per-route-features.md
+++ b/toc/rfc/rfc-draft-generic-per-route-features.md
@@ -35,8 +35,8 @@ The following constraints apply (types are as specified in [RFC 8259](https://rf
 
 #### CLI
 
-There MAY be either individual command line flags or a single flag which supports generic key-value
-pairs for the `map-route` sub-command.
+There SHOULD be a single flag which supports generic key-value pairs for the `map-route`
+sub-command. The flag should accept a JSON structure like the `-c` argument of `cf create-service`.
 
 #### Cloud Controller
 
--- a/toc/rfc/rfc-draft-generic-per-route-features.md
+++ b/toc/rfc/rfc-draft-generic-per-route-features.md
@@ -85,7 +85,10 @@ components.
 
 Gorouter MUST accept the new field and parse its contents. If the map cannot be parsed it MUST
 be ignored. If individual key-value pairs are invalid those MUST be ignored and other key-value
-pairs MUST still be parsed and acted upon.
+pairs MUST still be parsed and acted upon. For any option which could not be parsed gorouter MUST
+NOT fail but fall back to the value provided in the gorouter config. When encountering an invalid
+option gorouter MUST log the error at least once, the log SHOULD indicate what it tried to parse,
+why it failed and which value will come into effect.
 
 It MAY be decided to implement one feature as part of this RFC which would be implemented together
 with the generic logic for supporting the new map.
--- a/toc/rfc/rfc-draft-generic-per-route-features.md
+++ b/toc/rfc/rfc-draft-generic-per-route-features.md
@@ -76,6 +76,11 @@ The diego components MUST add support for the new field specified by the API and
 contents as-is without modification or, if the field exceeds technical limitations, MUST raise an
 error.
 
+### Route-Registrar
+
+Route-registrar MUST offer the new features in its config as they are added to the other
+components.
+
 #### Gorouter
 
 Gorouter MUST accept the new field and parse its contents. If the map cannot be parsed it MUST
--- a/toc/rfc/rfc-draft-generic-per-route-features.md
+++ b/toc/rfc/rfc-draft-generic-per-route-features.md
@@ -104,8 +104,8 @@ To-Do:
 
 ### Specifying Features
 
-It's the responsibility of the App Runtime Platform Working Group (ARD-WG) to specify new
-features based on what is specified in this RFC. When a new feature is proposed the ARD-WG uses
+It's the responsibility of the App Runtime Platform Working Group (ARP-WG) to specify new
+features based on what is specified in this RFC. When a new feature is proposed the ARP-WG uses
 the process it would use for any new feature to decide whether and how it should be implemented. It
 MAY involve other working groups in the process.
 
--- a/toc/rfc/rfc-draft-generic-per-route-features.md
+++ b/toc/rfc/rfc-draft-generic-per-route-features.md
@@ -124,9 +124,6 @@ not final):
 }
 ```
 
-To-Do:
-* Can we somehow report errors from gorouter back to the user?
-
 ### Specifying Features
 
 It's the responsibility of the App Runtime Platform Working Group (ARP-WG) to specify new features
@@ -145,9 +142,6 @@ per-route features:
 - Option to trim path mapping on request (i.e. an app mapped to `/some-path/` receiving a request
   `GET /some-path/books` will see the request as `GET /books` if the `trim-path` option is set,
   instead of the full path)
-
-To-Do:
-* Collect more feedback and extend the list.
 
 All of these options are already (or could easily be made) configurable for the entire CF
 installation. Changing them will almost certainly break existing scenarios which is why we are
--- a/toc/rfc/rfc-draft-generic-per-route-features.md
+++ b/toc/rfc/rfc-draft-generic-per-route-features.md
@@ -1,0 +1,127 @@
+# Meta
+[meta]: #meta
+- Name: Generic Per-Route Features
+- Start Date: 2024-01-08
+- Author(s): @maxmoehl, @domdom82
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (fill in with PR link after you submit it)
+
+## Summary
+
+## Problem
+
+Implementing per-route features in Cloud Foundry is a tedious process because it requires
+changes to multiple components in multiple working groups. Most components will only need to be
+made aware of new fields without adding any significant logic but the overhead of making the
+change still exists.
+
+On the other side are the components that actually need to be aware of the change. Mostly the
+routing components (gorouter, maybe route-emitter / -registrar), the API (cloud controller)
+and UIs (CLI).
+
+## Proposal
+
+To reduce the overall overhead of implementing new per-route features we propose to add a generic
+key-value map to the definition of a route across all components. All components handling the map
+SHOULD ignore the contents of the map except gorouter, cloud controller and cf CLI which MUST handle
+the map and its values accordingly. Other components MAY limit the size of the map or size of
+keys / values for technical reasons.
+
+The following constraints apply (types are as specified in [RFC 8259](https://rfc-editor.org/rfc/rfc8259)):
+* The map MUST be representable as an object.
+* The map MUST only use strings as keys.
+* The map MUST only use numbers, strings and the literal values `true` and `false` as values.
+
+### Required Changes
+
+#### CLI
+
+Support for a generic map MUST be added to the manifest. There MAY be either individual command line
+flags or a single flag which supports generic key-value pairs for the `map-route` sub-command.
+
+#### Cloud Controller
+
+The API MUST specify and implement a new field to carry the map inside the route object. The API
+MUST implement per-field validations as features of the map are specified. The route MUST be
+rejected with an appropriate error message if the provided map is invalid and MUST be passed on
+as-is otherwise. The API also MUST store the map as a generic key-value map to ensure changes to
+the map do not require a change to the database schema.
+
+Implementation details:
+* Add a new field to the [route model](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/models/runtime/route.rb).
+* Add validation to the [route validator](https://github.com/cloudfoundry/cloud_controller_ng/blob/e719d017ea4625397a97c2c14352ebdee66febe9/lib/cloud_controller/route_validator.rb#L2).
+
+#### BBS, Rep, Executor, Route-Emitter
+
+The diego components MUST add support for the new field specified by the API and forward the
+contents as-is without modification or, if the field exceeds technical limitations, MUST raise an
+error.
+
+#### Gorouter
+
+Gorouter MUST accept the new field and parse its contents. If the map cannot be parsed it MUST
+be ignored. If individual key-value pairs are invalid those MUST be ignored and other key-value
+pairs MUST still be parsed and acted upon.
+
+It MAY be decided to implement one feature as part of this RFC which would be implemented together
+with the generic logic for supporting the new map.
+
+As new features are specified they MUST be implemented in gorouter.
+
+Example `router.register` NATS message containing additional options (field names and values are not
+final):
+
+```json
+{
+  "host": "127.0.0.1",
+  "port": 4567,
+  "tls_port": 1234,
+  "protocol": "http1",
+  "uris": [
+    "my_first_url.localhost.routing.cf-app.com",
+    "my_second_url.localhost.routing.cf-app.com/some-path"
+  ],
+  "tags": {
+    "another_key": "another_value",
+    "some_key": "some_value"
+  },
+  "options": {
+    "lb-algo": "least-conn",
+    "conn-limit": 100,
+    "cookie": "my-session-cookie",
+    "trim-path": true
+  },
+  "app": "some_app_guid",
+  "stale_threshold_in_seconds": 120,
+  "private_instance_id": "some_app_instance_id",
+  "isolation_segment": "some_iso_seg_name",
+  "server_cert_domain_san": "some_subject_alternative_name"
+}
+```
+
+To-Do:
+* Can we somehow report errors from gorouter back to the user?
+
+### Specifying Features
+
+It's the responsibility of the App Runtime Platform Working Group (ARD-WG) to specify new
+features based on what is specified in this RFC. When a new feature is proposed the ARD-WG uses
+the process it would use for any new feature to decide whether and how it should be implemented. It
+MAY involve other working groups in the process.
+
+### Features Which Could be Implemented
+
+So far we have collected a list of features which would benefit from a more flexible approach to
+per-route features:
+
+- Custom Load-Balancing Algorithm
+- Custom Connection Limits
+- Custom session cookie name (currently set on platform level, mostly to `JSESSIONID`)
+- Option to trim path mapping on request (i.e. an app mapped to `/some-path/` receiving a request `GET /some-path/books` will see the request as `GET /books` if the `trim-path` option is set, instead of the full path)
+
+To-Do:
+* Collect more feedback and extend the list.
+
+All of these options are already (or could easily be made) configurable for the entire CF
+installation. Changing them will almost certainly break existing scenarios which is why we
+are looking for a cheap way to gradually allow for more configuration per route.
--- a/toc/rfc/rfc-draft-generic-per-route-features.md
+++ b/toc/rfc/rfc-draft-generic-per-route-features.md
@@ -36,8 +36,8 @@ The following constraints apply (types are as specified in [RFC 8259](https://rf
 
 #### CLI
 
-Support for a generic map MUST be added to the manifest. There MAY be either individual command line
-flags or a single flag which supports generic key-value pairs for the `map-route` sub-command.
+There MAY be either individual command line flags or a single flag which supports generic key-value
+pairs for the `map-route` sub-command.
 
 #### Cloud Controller
 
@@ -46,6 +46,25 @@ MUST implement per-field validations as features of the map are specified. The r
 rejected with an appropriate error message if the provided map is invalid and MUST be passed on
 as-is otherwise. The API also MUST store the map as a generic key-value map to ensure changes to
 the map do not require a change to the database schema.
+
+Support for a generic map MUST be added to the manifest. The updated routes section in a manifest
+MAY look like this:
+
+```yml
+version: 1
+applications:
+- name: test
+  routes:
+  - route: example.com
+    options:
+      loadbalancing-algorithm: round-robin
+      connection-limit: 15
+      session-cookie: FOOBAR
+      trim-path: true
+  - route: www.example.com/foo
+    protocol: http2
+  - route: tcp-example.com:1234
+```
 
 Implementation details:
 * Add a new field to the [route model](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/models/runtime/route.rb).
--- a/toc/rfc/rfc-draft-generic-per-route-features.md
+++ b/toc/rfc/rfc-draft-generic-per-route-features.md
@@ -10,22 +10,21 @@
 
 ## Problem
 
-Implementing per-route features in Cloud Foundry is a tedious process because it requires
-changes to multiple components in multiple working groups. Most components will only need to be
-made aware of new fields without adding any significant logic but the overhead of making the
-change still exists.
+Implementing per-route features in Cloud Foundry is a tedious process because it requires changes to
+multiple components in multiple working groups. Most components will only need to be made aware of
+new fields without adding any significant logic but the overhead of making the change still exists.
 
 On the other side are the components that actually need to be aware of the change. Mostly the
-routing components (gorouter, maybe route-emitter / -registrar), the API (cloud controller)
-and UIs (CLI).
+routing components (gorouter, maybe route-emitter / -registrar), the API (cloud controller) and
+UIs (CLI).
 
 ## Proposal
 
 To reduce the overall overhead of implementing new per-route features we propose to add a generic
 key-value map to the definition of a route across all components. All components handling the map
 SHOULD ignore the contents of the map except gorouter, cloud controller and cf CLI which MUST handle
-the map and its values accordingly. Other components MAY limit the size of the map or size of
-keys / values for technical reasons.
+the map and its values accordingly. Other components MAY limit the size of the map or size of keys /
+values for technical reasons.
 
 The following constraints apply (types are as specified in [RFC 8259](https://rfc-editor.org/rfc/rfc8259)):
 * The map MUST be representable as an object.
@@ -41,11 +40,11 @@ pairs for the `map-route` sub-command.
 
 #### Cloud Controller
 
-The API MUST specify and implement a new field to carry the map inside the route object. The API
-MUST implement per-field validations as features of the map are specified. The route MUST be
+The API MUST specify and implement a new field to carry the map inside the route object. The
+API MUST implement per-field validations as features of the map are specified. The route MUST be
 rejected with an appropriate error message if the provided map is invalid and MUST be passed on
-as-is otherwise. The API also MUST store the map as a generic key-value map to ensure changes to
-the map do not require a change to the database schema.
+as-is otherwise. The API also MUST store the map as a generic key-value map to ensure changes to the
+map do not require a change to the database schema.
 
 Support for a generic map MUST be added to the manifest. The updated routes section in a manifest
 MAY look like this:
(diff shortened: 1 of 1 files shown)


peanball:
Per-Route features? (Nicht pre-route)On Jan 8, 2024, at 16:54, Maximilian Moehl ***@***.***> wrote:﻿/cc @domdom82 @peanball



You can view, comment on, or merge this pull request online at:
  https://github.com/cloudfoundry/community/pull/743

Commit Summary

  7d449a6  wip: rfc: generic pre-route features


File Changes  (1 file)

  
    A
    toc/rfc/rfc-draft-generic-per-route-features.md
    (88)
  


Patch Links:

  https://github.com/cloudfoundry/community/pull/743.patch
  https://github.com/cloudfoundry/community/pull/743.diff

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,88 @@
...
+MAY involve other working groups in the process.
+
+### Features Which Could be Implemented
+
+So far we have collected a list of features which would benefit from a more flexible approach to 
+per-route features:
+
+- Custom Load-Balancing Algorithm
+- Custom Connection Limits
+
```
domdom82, line 82:
```suggestion
- Custom session cookie name (currently set on platform level, mostly to JSESSIONID)
- Option to trim path mapping on request (i.e. an app mapped to `/some-path/` receiving a request `GET /some-path/books` will see the request as `GET /books` if the `trim-path` option is set, instead of the pull path)
```

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,88 @@
...
+The diego components MUST add support for the new field specified by the API and forward the
+contents as-is without modification or, if the field exceeds technical limitations, MUST raise an 
+error.
+
+#### Gorouter
+
+Gorouter MUST accept the new field and parse its contents. If the map cannot be parsed it MUST 
+be ignored. If individual key-value pairs are invalid those MUST be ignored and other key-value 
+pairs MUST still be parsed and acted upon.
```
domdom82, line 58: we should provide a sample NATS message where the options field is set so people can see how it could look like

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,88 @@
...
+the map do not require a change to the database schema.
+
+To-Do:
+* Make the validation required?
+
+#### BBS, Rep, Executor, Route-Emitter
+
+The diego components MUST add support for the new field specified by the API and forward the
+contents as-is without modification or, if the field exceeds technical limitations, MUST raise an 
+error.
```
domdom82, line 52: I'll try to come up with what data structures would have to change in BBS, CC, database etc.

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,90 @@
...
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (fill in with PR link after you submit it)
+
+## Summary
+
+## Problem
+
+Implementing per-route features in Cloud Foundry is a tedious process because it requires
+changes to multiple components in multiple working groups. Most components will only need to be
+made aware of new fields without adding any significant logic but the over-head of making the
```
peanball, line 15:
```suggestion
made aware of new fields without adding any significant logic but the overhead of making the
```

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,90 @@
...
+- Custom Load-Balancing Algorithm
+- Custom Connection Limits
+- Custom session cookie name (currently set on platform level, mostly to `JSESSIONID`)
+- Option to trim path mapping on request (i.e. an app mapped to `/some-path/` receiving a request `GET /some-path/books` will see the request as `GET /books` if the `trim-path` option is set, instead of the full path)
+
+To-Do:
+* Collect more feedback and extend the list.
+
+All of these options are already (or could easily be made) configurable for the entire CF 
+installment. But changing them will almost certainly break existing scenarios which is why we
```
peanball, line 89:
```suggestion
installation. Changing them will almost certainly break existing scenarios which is why we 
```

peanball (commented):
Some minor editorial stuff

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,90 @@
...
+#### Gorouter
+
+Gorouter MUST accept the new field and parse its contents. If the map cannot be parsed it MUST 
+be ignored. If individual key-value pairs are invalid those MUST be ignored and other key-value 
+pairs MUST still be parsed and acted upon.
+
+It MAY be decided to implement one feature as part of this RFC which would be implemented together
+with the generic logic for supporting the new map.
+
+As new features are specified they MUST be implemented in gorouter.
```
domdom82, line 63:
```suggestion
As new features are specified they MUST be implemented in gorouter.

**Example router.register NATS message containing additional options**

```
{
  "host": "127.0.0.1",
  "port": 4567,
  "tls_port": 1234,
  "protocol": "http1",
  "uris": [
    "my_first_url.localhost.routing.cf-app.com",
    "my_second_url.localhost.routing.cf-app.com"
  ],
  "tags": {
    "another_key": "another_value",
    "some_key": "some_value"
  },
  "options": {
    "lb-algo": "least-conn",
    "conn-limit": 100,
    "cookie": "my-session-cookie" 
  },
  "app": "some_app_guid",
  "stale_threshold_in_seconds": 120,
  "private_instance_id": "some_app_instance_id",
  "isolation_segment": "some_iso_seg_name",
  "server_cert_domain_san": "some_subject_alternative_name"
}
```
```

domdom82: provided an example NATS message that contains all data types we would want to support (excluding floats).

maxmoehl: There are only [numbers](https://datatracker.ietf.org/doc/html/rfc8259#section-6) in JSON :)

maxmoehl:
ping @geofffranks @ameowlia

toc/rfc/rfc-draft-generic-per-route-features.md
domdom82: As discussed. @maxmoehl to add file pointers which would need to change in CC, e.g.:
- route model
- route controller
- route validator

maxmoehl: I've added that to the CC section above.

beyhan:
@maxmoehl I see that this RFC is not marked as draft any more. I assume that this is ready for review and add it to the list of RFCs which the TOC will look into next Tuesday. If it still not the case please let me know.

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,127 @@
...
+* The map MUST only use numbers, strings and the literal values `true` and `false` as values.
+
+### Required Changes
+
+#### CLI
+
+Support for a generic map MUST be added to the manifest. There MAY be either individual command line
+flags or a single flag which supports generic key-value pairs for the `map-route` sub-command.
+
+#### Cloud Controller
```
maxmoehl, line 42: Think about turning off from the operator. This could be done by providing feature toggles in CC to reject any or some custom parameters / values.

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,127 @@
...
+* Add a new field to the [route model](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/models/runtime/route.rb).
+* Add validation to the [route validator](https://github.com/cloudfoundry/cloud_controller_ng/blob/e719d017ea4625397a97c2c14352ebdee66febe9/lib/cloud_controller/route_validator.rb#L2).
+
+#### BBS, Rep, Executor, Route-Emitter
+
+The diego components MUST add support for the new field specified by the API and forward the
+contents as-is without modification or, if the field exceeds technical limitations, MUST raise an
+error.
+
+#### Gorouter
```
maxmoehl, line 60: How will potential inconsistency for parameters from stale routes be handled?

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,127 @@
...
+SHOULD ignore the contents of the map except gorouter, cloud controller and cf CLI which MUST handle
+the map and its values accordingly. Other components MAY limit the size of the map or size of
+keys / values for technical reasons.
+
+The following constraints apply (types are as specified in [RFC 8259](https://rfc-editor.org/rfc/rfc8259)):
+* The map MUST be representable as an object.
+* The map MUST only use strings as keys.
+* The map MUST only use numbers, strings and the literal values `true` and `false` as values.
+
+### Required Changes
```
maxmoehl, line 35: Add route registrar to the scope.

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,127 @@
...
+
+The diego components MUST add support for the new field specified by the API and forward the
+contents as-is without modification or, if the field exceeds technical limitations, MUST raise an
+error.
+
+#### Gorouter
+
+Gorouter MUST accept the new field and parse its contents. If the map cannot be parsed it MUST
+be ignored. If individual key-value pairs are invalid those MUST be ignored and other key-value
+pairs MUST still be parsed and acted upon.
```
maxmoehl, line 64: Specify in more detail what the failure mode should be and which options are taken when the provided options are invalid. This should include logging.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl, line 35: Clarify how users can be made aware when their configuration does not apply:
* App logs?
* App stats?

maxmoehl (commented):
Adding comments raised during the ARP WG meeting.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: Question to reviewers: Is there anyone who would make use of such toggles? If there's no demand for it, we would not provide them to simplify the implementation.

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,127 @@
...
+The following constraints apply (types are as specified in [RFC 8259](https://rfc-editor.org/rfc/rfc8259)):
+* The map MUST be representable as an object.
+* The map MUST only use strings as keys.
+* The map MUST only use numbers, strings and the literal values `true` and `false` as values.
+
+### Required Changes
+
+#### CLI
+
+Support for a generic map MUST be added to the manifest. There MAY be either individual command line
```
ameowlia, line 39: The manifest is managed by the Cloud Controller not the CLI. 
❓ Can you move this to the cc section?
❓ Can you show an example manifest with optional params?

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,127 @@
...
+- Start Date: 2024-01-08
+- Author(s): @maxmoehl, @domdom82
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (fill in with PR link after you submit it)
+
+## Summary
+
+## Problem
+
+Implementing per-route features in Cloud Foundry is a tedious process because it requires
```
ameowlia, line 13: Another layer to this problem is many of these configurations are on a foundation level only. It doesn't make sense for the foundation to control these things for _every_ app. App devs want to control these properties but they can't. We _could_ expose them 1 by 1, but it takes work (as you say here).

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Features Which Could be Implemented
+
+So far we have collected a list of features which would benefit from a more flexible approach to
+per-route features:
+
+- Custom Load-Balancing Algorithm
+- Custom Connection Limits
+- Custom session cookie name (currently set on platform level, mostly to `JSESSIONID`)
+- Option to trim path mapping on request (i.e. an app mapped to `/some-path/` receiving a request `GET /some-path/books` will see the request as `GET /books` if the `trim-path` option is set, instead of the full path)
```
ameowlia, line 120: Are these in the order that you think you would implement the features?

ameowlia (changes requested):

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: More or less, the first two items are the most requested ones from our stakeholders, the other two only came up once or twice but they would be easy to implement with this RFC.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: Thanks! I've adjusted it, please check again.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: There is the option to discard this RFC and take a different route. For example we could spend some time collecting all the features we want to expose and implement them at once to reduce the "fixed" overhead from implementing each feature on its own. But we prefer the flexibility as there will most certainly be new features down the road.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: I've added some more details so I'll resolve this for now.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: We already have this problem and don't really deal with it. The latest message simply overwrites the metadata of any previous message. So two route emitters sending conflicting data is already an issue, see for example the [route service URL](https://github.com/cloudfoundry/gorouter/blob/main/route/pool.go#L295).

I do see this as a valid concern but I'm not sure if it is in scope for this document to deal with this.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: I'm not into the details of this topic, maybe someone has ideas on how we could implement this?

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,148 @@
...
+
+There MAY be either individual command line flags or a single flag which supports generic key-value
+pairs for the `map-route` sub-command.
+
+#### Cloud Controller
+
+The API MUST specify and implement a new field to carry the map inside the route object. The
+API MUST implement per-field validations as features of the map are specified. The route MUST be
+rejected with an appropriate error message if the provided map is invalid and MUST be passed on
+as-is otherwise. The API also MUST store the map as a generic key-value map to ensure changes to the
```
tcdowney, line 46: >  The API also MUST store the map as a generic key-value map to ensure changes to the
map do not require a change to the database schema.

A note for others who might not be aware:

There is precedence for this and it is typically done in Cloud Controller by creating a `String` column and serializing the data in and out as JSON.

https://github.com/cloudfoundry/cloud_controller_ng/blob/e719d017ea4625397a97c2c14352ebdee66febe9/app/models/runtime/droplet_model.rb#L47-L48

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,148 @@
...
+Support for a generic map MUST be added to the manifest. The updated routes section in a manifest
+MAY look like this:
+
+```yml
+version: 1
+applications:
+- name: test
+  routes:
+  - route: example.com
+    options:
```
tcdowney, line 58: If the route already exists without these options would you expect it to be updated?

There hasn't really been much updating of routes before so you might want to mention that this endpoint would be enhanced:
https://v3-apidocs.cloudfoundry.org/version/3.159.0/index.html#update-a-route

And probably worth mentioning https://v3-apidocs.cloudfoundry.org/version/3.159.0/index.html#create-a-route as well. `cf create-route` and `cf map-route` (if the route doesn't already exist) both call that one.

maxmoehl: Good point! Yes, we expect those options to be modifiable via updates.

maxmoehl: I'm not that familiar with the CC API, are such `PATCH` calls usually additive or do they replace the object?

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: I've included the reference in the RFC.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: Do you think this should be included in the RFC? I'd probably leave such details up to the implementers.

I've also added the relevant API endpoints to the RFC.

tcdowney: I don't think it needs to get into a great amount of detail, but maybe a line saying that this config in the manifest will either create a new route with these options or update an existing route to have these options. Main reason I think that's important is because the current manifest logic is somewhat inconsistent (e.g. removing env vars from a manifest wont remove them from an app).

Speaking of that. If person A created a route ahead of time (`cf create-route`) with `options`, but someone pushed an app with a manifest that just mentioned the route, but without options, would you expect the options to be removed from the route? Or should that require a more intentional action?

toc/rfc/rfc-draft-generic-per-route-features.md
tcdowney: Can you think of ways that devs could use these options in bad ways that would impact other apps outside of their space? That would be the main reason I could think of for adding a feature flag for it.

toc/rfc/rfc-draft-generic-per-route-features.md
stephanme: I guess CF API responds with 4xx and an error message. The error msg can only be as good as the validation rules implemented in CC.

stephanme: Do you have a rough idea about the max size of the map (lets say serialized as json file)? 0.1k, 1k, 10k, 100k, ...?

toc/rfc/rfc-draft-generic-per-route-features.md
```diff
@@ -0,0 +1,157 @@
...
+* The map MUST be representable as an object.
+* The map MUST only use strings as keys.
+* The map MUST only use numbers, strings and the literal values `true` and `false` as values.
+
+### Required Changes
+
+#### CLI
+
+There MAY be either individual command line flags or a single flag which supports generic key-value
+pairs for the `map-route` sub-command.
```
stephanme, line 39: I would go for a single flag which supports generic key-value pairs in cli. This allows to keep it generic. CC and gorouter would be the only participants that need to have dedicated knowledge about the properties.

Similar to the -c option for specifying service instance parameters in `cf create-service`.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: That's an excellent question, I was hoping you can tell me that 😄 

The places where the size matters is probably in the databases which have to store it and the nats message (since they sent frequently). Since there are already [fields](https://github.com/cloudfoundry/cloud_controller_ng/blob/e719d017ea4625397a97c2c14352ebdee66febe9/app/models/runtime/droplet_model.rb#L47-L48) similar to this one: what is the limit on those?

maxmoehl: > I guess CF API responds with 4xx and an error message.

Yes, when the user submits a new route / updates an existing one the CC would do some validation. But it could still be the case that invalid data makes it to gorouter, how do we handle that?

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: I've been thinking about this for the past week and can't really decide what feels "right". But I'd be in favour of having to explicitly un-set the options. E.g. if you don't provide options they remain untouched and if you submit an empty map `{}` they will be removed.

Given:

```yaml
version: 1
applications:
- name: test
  routes:
  - route: a.example.com
    options:
      loadbalancing-algorithm: round-robin
      trim-path: true
  - route: b.example.com
    protocol: http2
  - route: b.example.com
    protocol: http2
    options: {}
```

Assuming all of the routes exist with some options already set:

* `a` would replace any existing options map with the one provided (potentially dropping what was set for, e.g. `session-cookie`).
* `b` would not change the options, if they exist, they stay.
* `c` would remove any previously attached options.

beyhan: I would be in favour of a "more" declarative approach here. Talking to users they see it as an CF issue that env vars aren't removed when they are not present in the manifest any more or service bindings are not touched, when they are removed from the manifest. In general we should strive to have the "desired state truth" defined in the app manifest.

maxmoehl: Agreed, but I wouldn't want to introduce a new feature that behaves differently to all the existing ones, that would probably cause even more confusion.

toc/rfc/rfc-draft-generic-per-route-features.md
ameowlia: > invalid data makes it to gorouter, how do we handle that?

Maybe log in gorouter and then just go with the foundation default? I wish we could force an app log here to warn the app dev. But maybe we are over-engineering this.

toc/rfc/rfc-draft-generic-per-route-features.md
ameowlia: > Can you think of ways that devs could use these options in bad ways that would impact other apps outside of their space?

It would impact if the route was shared between spaces, but that is the only way I can think of.

> Question to reviewers: Is there anyone who would make use of such toggles?

I haven't been able to think of any since we posed this question at the WG meeting. I think we can make this feature without a toggle and add it in later if it is requested.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: I've adjusted the wording accordingly.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: Okay so for now we will only log this on gorouter as an error log which is something the operator will have to investigate. If this becomes too much of an issue we might re-visit the topic and implement a way for app devs to see those errors themselves.

@stephanme regarding the limits: can you provide some insights on what an acceptable size would be for CC?

toc/rfc/rfc-draft-generic-per-route-features.md
domdom82: I'd be interested to know how it is implemented today?
If I do first
```
cf map-route my-app example.com --hostname myhost --app-protocol http2
```
then do
```
cf push
```
with a manifest saying
```
applications:
- name: my-app
  routes:
  - route: myhost.example.com
```
so no `protocol` key was defined in the manifest.
What does it do? Does it keep `http2` or overwrite with the default of `http1`?

Whatever we do today, it should be the same with the `options` field IMO.

toc/rfc/rfc-draft-generic-per-route-features.md
domdom82: > Do you have a rough idea about the max size of the map (lets say serialized as json file)? 0.1k, 1k, 10k, 100k, ...?

I think 10K is a reasonable size that should fit all current needs.

maxmoehl:
I've been resolving conversations where I either adjusted the RFC or think that we have consensus on not addressing the comment in this RFC, the only discussion left right now is about [API semantics in CC](https://github.com/cloudfoundry/community/pull/743#discussion_r1529225249). If you had comments please double check that they are addressed appropriately and unresolve discussions which still need clarification!

I would like to move into the final comment period soon-ish.

toc/rfc/rfc-draft-generic-per-route-features.md
beyhan: I agree with the suggestion of @domdom82 to have this consistent with the current implementation of the `http2` option. My intention was to provide feedback from CF users and motivate for a more declarative approach if it is possible and still keep things consistent.

beyhan: I tried this out. First I defined the route for my app with:
```
cf map-route cf-nodejs-bv-test <domain> --hostname cf-nodejs-bv-test-zany-chimpanzee-ra --app-protocol http2
```
Afterwards, I added that rout to my manifest with:
```
---
applications:
- name: cf-nodejs-bv-test
  memory: 1GB
  instances: 1
  random-route: true
  routes:
  - route: cf-nodejs-bv-test-zany-chimpanzee-ra.<domain>
```

Then I pushed the app:

```
...
Updating with these attributes...
  ---
  applications:
  - name: cf-nodejs-bv-test
    instances: 1
    memory: 1GB
+   random-route: true
    routes:
+   - route: cf-nodejs-bv-test-zany-chimpanzee-ra.<domain>
-   - protocol: http2
-     route: cf-nodejs-bv-test-zany-chimpanzee-ra.<domain>
...
```

The push diff states that the `http2` protocal will be removed but checking with the CF APIs the protocol is still `http2`
```
cf curl "/v3/routes?app_guids=3f3bbd62-d46a-4e55-871e-f17a551a6aac" | jq .
...
{
      "guid": "322a5fe3-3f93-4ecb-9e9c-b1103c327643",
      "created_at": "2024-04-08T08:40:22Z",
      "updated_at": "2024-04-08T08:40:22Z",
      "protocol": "http",
      "host": "cf-nodejs-bv-test-zany-chimpanzee-ra",
      "path": "",
      "port": null,
      "url": "cf-nodejs-bv-test-zany-chimpanzee-ra.<domain>",
      "destinations": [
        {
          "guid": "f6482e06-736b-4555-87a5-22cfa68b7848",
          "app": {
            "guid": "3f3bbd62-d46a-4e55-871e-f17a551a6aac",
            "process": {
              "type": "web"
            }
          },
          "weight": null,
          "port": 8080,
          "protocol": "http2"
        }
      ],
```
I'm not sure whether this is a bug in CAPI or the expected behaviour. @tcdowney or @stephanme could you please comment on that?

stephanme: I think this (indeed surprising) behaviour is correct according to the API spec: https://v3-apidocs.cloudfoundry.org/#apply-a-manifest-to-a-space

> Apply changes specified in a manifest to the named apps and their underlying processes. The apps must reside in the space. These changes are additive and will not modify any unspecified properties or remove any existing environment variables, routes, or services.

Since the protocol property was not specified, it remains unchanged which results in http2. The diff output could be considered as buggy.

Disclaimer: I don't understand the coding in every detail and I could not find a matching test case that deals with this protocol case. I will raise this in the CAPI Open Office Hour.

beyhan: We discussed this during the TOC meeting today and we should go for consistent behaviour as documented in the APIs.

beyhan:
We decided to start the FCP for this RFC with the goal to accept it next Tuesday 16th of April.

toc/rfc/rfc-draft-generic-per-route-features.md
maxmoehl: I've added a paragraph to capture that:

> The overall goal MUST be to implement the new fields in a way that fits in with the current behavior of the cloud controller API. It is up to the maintainers / implementers to decide on the exact behavior.

peanball (approved):

toc/rfc/rfc-draft-generic-per-route-features.md
tcdowney: I don't have a terribly strong opinion here, but I think we should go with what's consistent for now and provide a way for users to explicitly unset the properties in the manifest.

I know that @Gerg has been thinking about what a truly declarative "version 2" of the CF manifest my look/behave like and I think that would be an opportunity to incorporate some of that feedback. He has a doc somewhere, but I think it was made private due to some Google doc migrations. I'll let him chime in when he's available / let him know about this.

tcdowney (approved):

plowin (approved):

toc/rfc/rfc-draft-generic-per-route-features.md
ameowlia: > we should go for consistent behaviour as documented in the APIs.

I agree. I also remember being confused about this behavior regarding readiness health check manifest properties. It's odd, but that is how the app manifest currently works. I think changing it is out of scope for this RFC.

ameowlia (approved):

ChrisMcGowan (approved):

rkoster (approved):

beyhan (approved):

stephanme (approved):

PR #750, opened 2024-01-11
Archive garden-windows-ci repo

ebroberson:
Recently we had some issues with the windows-rootfs pipeline failing because of its references to both [garden-windows-ci](https://github.com/cloudfoundry/garden-windows-ci) and the [WG CI repo](https://github.com/cloudfoundry/wg-app-platform-runtime-ci). The conflict arose because we have some opsfiles that have the same name, but different functionality. We were able to fix this by removing the bad opsfile, but it would be better if we had one source of truth for our CI pipelines.

Since `garden-windows-ci` is no longer referenced by our pipelines and the relevant files are now in the WG CI repo, we should archive `garden-windows-ci`.

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -1266,6 +1266,7 @@ orgs:
         private: true
       garden-windows-ci:
         has_projects: true
+        archived: true
       garden-windows-tools-release:
         has_projects: true
       git-release-notes:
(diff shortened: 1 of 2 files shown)


ameowlia (approved):

ameowlia:
This is for my working group and I approve. I can't remember if I need to wait for full TOC approval.

beyhan (approved):

beyhan:
Usually, I take this into the TOC meeting but if we get another approval by @rkoster, @stephanme or @ChrisMcGowan we don't need to wait.

stephanme:
You should remove the project from https://github.com/cloudfoundry/community/blob/main/toc/working-groups/app-runtime-platform.md (repositories of area Garden Containers) as well.

rkoster (approved):

beyhan:
@stephanme thanks for the good catch. The request has been implemented and we have enough approvals I will merge to sped up things.

PR #719, opened 2023-12-04
Update app-runtime-interfaces buildpacks approvers

brayanhenao:
# Changes

## Removal of Ryan Moran
I removed Ryan Moran from all relevant groups, he is no longer part of the Buildpacks Team.

## Addition of Brayan Henao, Replacing Ryan Moran in Approvers' Roles
In response to approvers responsibilities left by Ryan Moran's departure, I have taken the initiative to introduce me into some specific teams. This decision is based on the need for at least three approvers in teams where Ryan's absence created a gap. The following exceptions apply:

- **Buildpacks Tooling:** Given my active involvement in the v2b release process, particularly within the `buildpacks-ci` repository and its related counterparts, I am confident in fulfilling the approver role.

- **Buildpacks Docs:** I haven't taken up the approver role there yet because I haven't contributed enough to the repositories

For your reference, please review my contributions in the following areas:

- **Go:** [Go Contributions](https://gist.github.com/brayanhenao/be8e704f63bf04145caa87f6b7480d47)
- **Nodejs:** [Nodejs Contributions](https://gist.github.com/brayanhenao/9e7e587a9877016493eaa0c0594cc28c)
- **Ruby:** [Ruby Contributions](https://gist.github.com/brayanhenao/243102c57fd4ae275e7b10bd128a14d7)
- **Tooling:** [Tooling Contributions](https://gist.github.com/brayanhenao/891758d7c81edd52e831568c2f9dc009)

PR #752, opened 2024-01-12
Clean roles of removed users

beyhan:
ryanmoran has been removed from the cloudfoundry org with this [pr](https://github.com/cloudfoundry/community/pull/719). The is still a role assigned to him which fails the org automation


stephanme (approved):

beyhan:
This is blocking our github org automation, that is why I will merge.

PR #746, opened 2024-01-10
Add roles yaml section to Vulnerability Management WG

beyhan:
This is need for the inactive user automation to not suggest users for deletion who has roles. cc @paulcwarren

PR #753, opened 2024-01-12
Add Vulnerability Management WG contributors

beyhan:
With [pr](https://github.com/cloudfoundry/community/pull/746) we introduced roles yaml for the VM WG. We missed to add the members as contributors with that pr.

### DIFF
--- a/org/contributors.yml
+++ b/org/contributors.yml
@@ -265,6 +265,7 @@ contributors:
 - ifindlay-cci
 - IgorBelyi
 - ihcsim
+- ijzerman
 - ikasarov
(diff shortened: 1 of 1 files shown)


stephanme:
Users with a WG role become a member of cloudfoundry org even if they are not part of contributors.yml. But of course, it doesn't harm to add them as contributors. Our promotion rules even require it.

The org automation is currently red and the new teams that should have been created by #746 don't exist yet.

stephanme (approved):

ChrisMcGowan (approved):

PR #762, opened 2024-01-25
Add draft RFC for establishing CFF as a CVE Numbering Authority

paulcwarren:
As discussed at TOC 1/23

🚀 Link for easy viewing: https://github.com/cloudfoundry/community/blob/rfc/cff-cve-numbering-authority/toc/rfc/rfc-draft-cff-cve-numbering-authority.md

### DIFF
--- a/toc/rfc/rfc-draft-cff-cve-numbering-authority.md
+++ b/toc/rfc/rfc-draft-cff-cve-numbering-authority.md
@@ -1,0 +1,22 @@
+# Meta
+[meta]: #meta
+- Name: CFF CVE Numbering Authority
+- Start Date: 2024-01-25
+- Author(s): paulcwarren
+- Status: Draft 
+- RFC Pull Request: (fill in with PR link after you submit it)
+
+
+## Summary
+
+For the last several years the CFF has relied on sponsoring member VMWare as the CVE numbering authority.  When we have needed to publish a CVE, we claim a CVE number from their reservation block.  
+
+This responsibility should lie with the foundation itself.
+
+## Problem
+
+CVE must be published in a timely fashion.  Any interruptions in the allocation of a CVE number, pre-disclosure or disclosure puts installations and their Users are at risk.
+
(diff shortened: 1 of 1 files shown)


paulcwarren:
It's a whole thing from what I can tell.  The steps as far as I can tell are:
 
Eligibility assessment
Application
Evaluation
Training, Signing the MOU.  

At which point we could start Issuing CVEs.

The interesting thing is that I am pretty sure that we used to be an authority, or we already are one (and we just flipped to using VMWare for a while), and the Linux Foundation is obviously already is one so it's possible it might not be a big deal.

As for timing, I cant make any determination there.  I think we would just need to get started.

christopherclark:
@paulcwarren I don't have context into whether the CFF was a CNA in the past, but if it was that would have been when it was a separate legal entity from the LF. So that's likely moot at this point. The LF in not a CNA, however some LF projects are, i.e. https://www.cve.org/PartnerInformation/ListofPartners/partner/zephyr

So, this is definitely possible, I just need to dig into this a bit more. OpenSSF recently published a blog post on this topic: https://openssf.org/blog/2023/11/27/openssf-introduces-guide-to-becoming-a-cve-numbering-authority-as-an-open-source-project/  

I'll reach out soon.

paulcwarren:
@christopherclark any thoughts on how to progress this.  I have not heard back from MITRE at all.  I can ping them again, or jsut start applying to become a CNA.  WDYT?

christopherclark:
@paulcwarren Sure, let's go ahead and apply. I'll be around Mon-Wed next week, happy to assist as needed.

paulcwarren:
After careful consideration, we're closing this issue, as the Cloud Foundry Foundation cannot independently publish CVEs under the current setup. Ownership of the root CVE allocation with MITRE remains with VMware (now Broadcom), and without a transfer of this authority, it isn't feasible for the Foundation to take on this role directly.

PR #765, opened 2024-01-31
Allow org automation to delete up to 80% of users

beyhan:
With (Inactive users to be deleted)[https://github.com/cloudfoundry/community/pull/736] we want to clean inactive users. This is the first run and it calculated over 400 users. The org automation doesn't allow to delete so many users at onance. This will re-configure the limit temporarily and I will revert it after the inactive users are deleted


rkoster (approved):

ChrisMcGowan (approved):

stephanme (approved):

beyhan:
Because this is blocking the automation I'm going to merge this.

PR #691, opened 2023-09-20
[ARI] Add temp MultiApps team for managing GitHub secrets

boyan-velinov:

PR #769, opened 2024-02-02
[ARI] Add temp autoscaler team for managing GitHub secrets

silvestre:
Without admin access we can no longer maintain GitHub Actions secrets.

Prior art for admin override teams: #441 + #691

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -2470,6 +2470,19 @@ orgs:
           multiapps: admin
           multiapps-controller: admin
           multiapps-cli-plugin: admin
+      temp-autoscaler-admins:
+        description: "Temporary team for managing autoscaler secrets until we have a permanent plan."
+        maintainers:
+        - asalan316
+        - silvestre
+        - joergdw
+        - olivermautschke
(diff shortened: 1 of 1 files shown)


org/cloudfoundry.yml
```diff
@@ -2470,6 +2470,19 @@ orgs:
           multiapps: admin
           multiapps-controller: admin
           multiapps-cli-plugin: admin
+      temp-autoscaler-admins:
```
stephanme, line 2473: Better: temp-ari-autoscaler-admins
Points directly to the WG and matches the naming pattern of most of the other temp admin teams

stephanme (changes requested):

org/cloudfoundry.yml
silvestre: Thanks, good catch. Copy/pasted from the wrong example. :)

stephanme (approved):

ChrisMcGowan (approved):

beyhan (approved):

ameowlia (approved):

PR #775, opened 2024-02-07
Add svcboteos to contributors

jbooherl:
This CI bot is used to bump [metric-store-release](https://github.com/cloudfoundry/metric-store-release) when we are about to do a new release.

It is already a part of the App Runtime Platform bots group


ameowlia (approved):

beyhan:
@jbooherl usually, it is enough when bots are listed in the WG. No need to add them as contributors. Was something not working for your? Just curious.

PR #777, opened 2024-02-09
Add cf-cli-eng bot to App Runtime Interfaces

a-b:
cf-cli-eng is used in CF CLI automations

cc @moleske

### DIFF
--- a/toc/working-groups/app-runtime-interfaces.md
+++ b/toc/working-groups/app-runtime-interfaces.md
@@ -49,6 +49,8 @@ bots:
   github: capi-bot
 - name: ARI WG Git Bot
   github: ari-wg-gitbot
+- name: CF CLI Eng
+  github: cf-cli-eng
 areas:
 - name: Autoscaler
   approvers:
--- a/toc/working-groups/app-runtime-interfaces.md
+++ b/toc/working-groups/app-runtime-interfaces.md
@@ -49,8 +49,6 @@ bots:
   github: capi-bot
 - name: ARI WG Git Bot
   github: ari-wg-gitbot
-- name: CF CLI Eng
-  github: cf-cli-eng
 areas:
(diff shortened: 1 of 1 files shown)


a-b:
@stephanme, could you please review this pull request since our working group's tech lead is on leave?

beyhan:
@moleske could you please put a thumbs up or approval on this. 

@stephanme could review or I will add this to be reviewed by the TOC.

toc/working-groups/app-runtime-interfaces.md
```diff
@@ -49,6 +49,8 @@ bots:
   github: capi-bot
 - name: ARI WG Git Bot
   github: ari-wg-gitbot
+- name: CF CLI Eng
+  github: cf-cli-eng
```
stephanme, line 53: Is there a real need to add cf-cli-eng as WG bot and not as WG area bot just for the CLI area?
See Notification WG area for an example.

There are more bot users that should be moved from WG into WG area. I will put this topic on the agenda for the next WG meeting.

stephanme (changes requested):

toc/working-groups/app-runtime-interfaces.md
moleske: I didn't realize you could scope bots to more specific areas.  Here are the example lines https://github.com/cloudfoundry/community/blob/522836cdbcba75494e25ebfce10e2ef42777ae46/toc/working-groups/app-runtime-interfaces.md?plain=1#L432-L435

moleske (approved):
This seems reasonable, given we make the change @stephanme is making about scoping it to the cli area rather than all of app runtime interfaces

a-b:
Addressed @stephanme suggestion and created follow up #781 for CAPI

PR #782, opened 2024-02-15
Add the bosh-windows-ci bot to the FIWG

danielfor:
Moving the bosh-windows-ci bot to the Foundational Infrastructure Working Group. This bot creates the releases for stembuild.


jpalermo (approved):
👍

beyhan:
We should try to add bots on area level, which is also supported and gives more restricted access to them instead been global for the WG.

PR #796, opened 2024-03-19
RFC to introduce Cloud Native Buildpacks lifecycle

modulo11:
This RFC tries to gently introduce Cloud Native Buildpacks to Cloud Foundry. It builds upon [RFC 0017](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0017-add-cnbs.md) but introduces CNBs directly in the platform instead of shimming them.

[Preview](https://github.com/sap-contributions/cf-community/blob/cnb-lifecycle/toc/rfc/rfc-draft-cnb-lifecycle.md)

@cloudfoundry/toc 
CC @loewenstein @c0d1ngm0nk3y, @pbusko, @nicolasbender, @beyhan, @stephanme, @stefanlay, @PlamenDoychev

### DIFF
--- a/toc/rfc/rfc-draft-cnb-lifecycle.md
+++ b/toc/rfc/rfc-draft-cnb-lifecycle.md
@@ -1,0 +1,135 @@
+# Meta
+
+[meta]: #meta
+- Name: Cloud Native Buildpacks Lifecycle
+- Start Date: 2024-03-19
+- Author(s): @c0d1ngm0nk3y, @pbusko, @nicolasbender, @modulo11
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: https://github.com/cloudfoundry/community/pull/796
+- Updates: [RFC 0017](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0017-add-cnbs.md)
+
+## Summary
+
+[Cloud Native Buildpacks (CNBs)](https://buildpacks.io/), also known as v3 buildpacks, are the current generation of buildpacks and offer some improvements over the v2 buildpacks that CF Deployment currently uses. The Cloud Foundry Foundation already has an implementation of Cloud Native Buildpacks via the [Paketo](https://paketo.io/) project, however these CNBs can't currently be used in CF.
+
+This RFC introduces a new optional lifecycle to Cloud Foundry which enables users to build their applications using Cloud Native Buildpacks.
+
+## Problem
+
+The v2 buildpacks are effectively in maintenance mode and do not receive substantial new features. By not integrating with v3 buildpacks, Cloud Foundry is missing out on new buildpacks (e.g. Java Native and Web Servers CNBs) as well as any new features that are added to the still-actively-developed v3 buildpacks.
+
+## Proposal
+
+- Introduce a new [lifecycle type](https://v3-apidocs.cloudfoundry.org/index.html#lifecycles) `cnb` and its lifecycle data
+- Introduce a new [app lifecycle](https://github.com/cloudfoundry/diego-design-notes#app-lifecycles) called `cnbapplifecycle` which interacts with the [CNB Lifecycle](https://github.com/buildpacks/lifecycle)
+- Reuse [cflinuxfs4](https://github.com/cloudfoundry/cflinuxfs4) as the base for staging and running apps
+- Introduce a new flag to CLI and the [app manifest](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html) to be able to use Cloud Native Buildpacks instead of v2 buildpacks
+
+### Architecture
+
+This will require changes in the following releases:
+
+- CF CLI
+- Cloud Controller
+- Diego
+
+No changes to how Diego runs workloads are necessary to implement this RFC.
+
+- The cli will forward a new app lifecycle type to the cloud controller.
+- The cloud controller will use a new `cnbapplifecycle` for the application.
+
+Affected cloud controller APIs (all that interact with [lifecycles](https://v3-apidocs.cloudfoundry.org/index.html#lifecycles)):
+
+- [apps](https://v3-apidocs.cloudfoundry.org/index.html#apps)
+- [builds](https://v3-apidocs.cloudfoundry.org/index.html#builds)
+- [droplets](https://v3-apidocs.cloudfoundry.org/index.html#droplets)
+- [manifests](https://v3-apidocs.cloudfoundry.org/index.html#manifests)
+
+### Goals
+
+- Establish the latest generation of buildpacks in CF as first-class citizen
+- Increase cohesion and app portability between CF Deployment and [Korifi](https://www.cloudfoundry.org/technology/korifi/), via mutual Paketo integration
+- Increase adoption of Cloud Native Buildpacks
+- Open the door for eventual deprecation of the v2 buildpacks, reducing maintenance costs (v2 buildpack deprecation is NOT included in this RFC)
+- No fundamental changes to the architecture of CF
+  - Result of the staging process will be a droplet
+  - No OCI registry necessary
+  - Reuse cflinuxfs4 as rootfs during build and run
+
+### High Level Implementation
+
+#### CNB App Lifecycle
+
+Introduce a new `lifecycle type` that enables the cloud controller to differentiate between the classical buildpacks and Cloud Native Buildpacks. On a high level, it will be very similar to the existing buildpackapplifecycle (v2 buildpacks). The new app lifecycle acts as a CNB [platform](https://buildpacks.io/docs/for-app-developers/concepts/platform/) and will:
+
+1. Download the app source code from the blobstore
+1. Download the CNB app lifecycle from the blobstore
+1. Download the configured buildpacks
+1. Write an [order.toml](https://github.com/buildpacks/spec/blob/main/platform.md#ordertoml-toml) with configured buildpacks
+1. Execute [detect](https://github.com/buildpacks/spec/blob/main/platform.md#detector) and [build](https://github.com/buildpacks/spec/blob/main/platform.md#builder) phases using the [CNB lifecycle](https://github.com/buildpacks/lifecycle)
+1. Package the result into a droplet and upload it to the blobstore
+1. Write a result.json file with the [Staging Result](https://github.com/cloudfoundry/buildpackapplifecycle/blob/f4b2bc9ff6cc6229402d7c27e887763154cf0378/models.go#L73-L80)
+
+#### CNB Lifecycle Type
+
+Introduce a new type of lifecycle type which indicates that Cloud Native Buildpacks should be used. In future, this can be enhanced with additional CNB inputs. For this RFC we’d start with:
+
+```json
+{
+  "type": "cnb",
+  "data": {
+    "buildpacks": ["gcr.io/paketo-buildpacks/java"],
+    "stack": "cflinuxfs4"
+  }
+}
+```
+
+Both, building and running an app will be based on the configured stack. If no stack is provided, the platform default is used. An empty (or not provided) list of buildpacks will lead to an error. This essentially means, that no auto-detection is supported at the moment. Once [system CNBs](#system-buildpacks) are supported, this behavior will change.
+
+#### CF CLI
+
+New flag `–[no-]cnb` will be introduced to the `cf push` command. The flag can not be combined with any ` --docker-*` commands. When the flag is set, CF CLI must set the lifecycle type to `cnb`.
+
+#### App Manifest
+
+New property `cnb: true|false` will be added to the App manifest.
+
+```yaml
+---
+applications:
+  - name: test-app
+    instances: 1
+    cnb: true
+    buildpacks:
+      - gcr.io/paketo-buildpacks/java
+```
+
+The `cnb` property cannot be present together with any of the `docker-*` properties. When the flag is set to `true`, the CF CLI must set the lifecycle type to `cnb`.
+
+Both changes (CLI and manifest) were chosen because they are simple (from a user perspective), easy to implement and remove, if CNBs will become the standard lifecycle in future.
+
+### Alternative APIs
+
+- Instead of a binary switch, introduce a `buildpack-type` (`v2`/`v3` or `cf`/`cnb` or `classic`/`cnb`) to distinguish between different lifecycles
+
+#### Diego Release
+
+The new lifecycle package `cnb_app_lifecycle` will be added to the Diego BOSH release, next to the existing `buildpack_app_lifecycle` and `docker_app_lifecycle`. This lifecycle package should be served by the File Server in the same way as existing lifecycle packages.
+
+### Possible Future Work
+
+#### Add Paketo Stack(s)
+
+This RFC does not include the addition of a new stack to Cloud Foundry, rather the resulting droplet would run on top of the existing `cflinuxfs4`. This should work for most, if not all apps, as the stack is much bigger (in terms of native packages installed) than the stacks used by Paketo.
+
+Paketo provides multiple stacks that are compatible with the Paketo buildpacks. There is an opportunity to adopt some or all of the Paketo stacks into CF Deployment. The Paketo buildpacks have greater cohesion with the Paketo stacks than with `cflinuxfs4`, and the Paketo "base" and "tiny" stacks could offer security-conscious CF Deployment users stacks with far fewer native packages than are currently included with `cflinuxfs4`.
+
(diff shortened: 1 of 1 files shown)


toc/rfc/rfc-draft-cnb-lifecycle.md
```diff
@@ -0,0 +1,135 @@
...
+
+#### Diego Release
+
+The new lifecycle package `cnb_app_lifecycle` will be added to the Diego BOSH release, next to the existing `buildpack_app_lifecycle` and `docker_app_lifecycle`. This lifecycle package should be served by the File Server in the same way as existing lifecycle packages.
+
+### Possible Future Work
+
+#### Add Paketo Stack(s)
+
+This RFC does not include the addition of a new stack to Cloud Foundry, rather the resulting droplet would run on top of the existing `cflinuxfs4`. This should work for most, if not all apps, as the stack is much bigger (in terms of native packages installed) than the stacks used by Paketo.
```
FloThinksPi, line 123: The mechanism that currently exists rebases the base layer(stack) on every application restart onto the droplet.
The behaviour CNBs have is to add the base image (stack equivalent) at build time to a resulting OCI Image.
The way this mechanism works is different when building internally or externally with the same CNB.

The way CNBs add the base layer while building can be favorable for the stack depreciation problem where currently its not possible to remove a stack with application still depending on it. When the stack gets added at build time this changes as then missing stacks just prevent restaging. Already build droplets include the stack in the droplet and keep running.
However this also comes with a some downsides:
- Stack update means restaging all apps(resource costly, complex)
- The stack is saved multiple times in the blobstore(storage and traffic)

Although using the build workflow that is common for CNBs in CF would help plus make the systems behave similar it comes with some costs im not sure if they are worth it yet but maybe something to consider.

toc/rfc/rfc-draft-cnb-lifecycle.md
```diff
@@ -0,0 +1,135 @@
...
+      - gcr.io/paketo-buildpacks/java
+```
+
+The `cnb` property cannot be present together with any of the `docker-*` properties. When the flag is set to `true`, the CF CLI must set the lifecycle type to `cnb`.
+
+Both changes (CLI and manifest) were chosen because they are simple (from a user perspective), easy to implement and remove, if CNBs will become the standard lifecycle in future.
+
+### Alternative APIs
+
+- Instead of a binary switch, introduce a `buildpack-type` (`v2`/`v3` or `cf`/`cnb` or `classic`/`cnb`) to distinguish between different lifecycles
```
stephanme, line 113: +1 for `buildpack-type` in app manifest. It allows smooth extension in the future while a `cnb: true` means that every new lifecycle type needs additional properties. Future extensions might be a CNB lifecycle that builds OCI images (including stack) instead of traditional droplets (as @FloThinksPi suggested below).

`buildpack-type` matches to a certain extent the `lifecycle.type` attribute in the [CF API](https://v3-apidocs.cloudfoundry.org/index.html#lifecycles). Full consistency between CF API, app manifest and cf cli is probably out of reach due to history.

loewenstein: As you explicitly mention a potential CNB based lifecycle that introduces full OCI images, we should then probably not use `cnb`, but rather something like `cnb-droplet`. WDYT?

toc/rfc/rfc-draft-cnb-lifecycle.md
loewenstein: I'd say that the thought of using CNBs to produce full OCI images is independent from this "potential future work". Also for a droplet based approach, the smaller Paketo base images in *Static*, *Tiny* and *Base* are an improvement for application security imho.

toc/rfc/rfc-draft-cnb-lifecycle.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Summary
+
+[Cloud Native Buildpacks (CNBs)](https://buildpacks.io/), also known as v3 buildpacks, are the current generation of buildpacks and offer some improvements over the v2 buildpacks that CF Deployment currently uses. The Cloud Foundry Foundation already has an implementation of Cloud Native Buildpacks via the [Paketo](https://paketo.io/) project, however these CNBs can't currently be used in CF.
+
+This RFC introduces a new optional lifecycle to Cloud Foundry which enables users to build their applications using Cloud Native Buildpacks.
+
+## Problem
+
+The v2 buildpacks are effectively in maintenance mode and do not receive substantial new features. By not integrating with v3 buildpacks, Cloud Foundry is missing out on new buildpacks (e.g. Java Native and Web Servers CNBs) as well as any new features that are added to the still-actively-developed v3 buildpacks.
```
dsboulder, line 19: The folks on our side are doing development on both V2 and CNBs in parallel, I'm seeing new capabilities come online for V2 frequently as well. What's giving you folks this impression? This will be significant work in quite a few areas to support CNBs in CF and I want to make it's worth it.

toc/rfc/rfc-draft-cnb-lifecycle.md
```diff
@@ -0,0 +1,135 @@
...
+- Establish the latest generation of buildpacks in CF as first-class citizen
+- Increase cohesion and app portability between CF Deployment and [Korifi](https://www.cloudfoundry.org/technology/korifi/), via mutual Paketo integration
+- Increase adoption of Cloud Native Buildpacks
+- Open the door for eventual deprecation of the v2 buildpacks, reducing maintenance costs (v2 buildpack deprecation is NOT included in this RFC)
+- No fundamental changes to the architecture of CF
+  - Result of the staging process will be a droplet
+  - No OCI registry necessary
+  - Reuse cflinuxfs4 as rootfs during build and run
+
+### High Level Implementation
```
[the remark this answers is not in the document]
dsboulder: I never saw a _great_ solution to the VCAP services => CN bindings problem? Is there a way to make this work that isn't a ton of brittle translation tables?

toc/rfc/rfc-draft-cnb-lifecycle.md
PlamenDoychev: I would also prefer to have `buildpack-type` instead of cnb: true, which more or less looks like a switch.

toc/rfc/rfc-draft-cnb-lifecycle.md
loewenstein: We've taken this pretty literally from [RFC 0017](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0017-add-cnbs.md#problem).

Regarding the amount of work...
> This will be significant work in quite a few areas to support CNBs in CF and I want to make it's worth it.

... our plan was not only to propose this RFC, but also to work on contributing the implementation.

toc/rfc/rfc-draft-cnb-lifecycle.md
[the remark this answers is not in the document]
beyhan: The https://github.com/cloudfoundry/community/pull/804 tries to address this point. I would be interested in your feedback. 

FYI @dmikusa as we discussed this during the TOC meeting and feedback is welcome.

toc/rfc/rfc-draft-cnb-lifecycle.md
pbusko: We've updated the RFC and changed the `cnb: true|false` to the `lifecycle: buildpack|docker|cnb`. However, the introduction of `cnb-oci` or `cnb-droplet` seems to be too early. What would be the benefit for the end user to be able to choose between droplet-style and raw OCI? But since the whole feature will be behind a feature flag, the behaviour could be also controller by administrators on the CCNG config level.

stephanme: Potential end user benefit:
- cnb-droplet: user gets stack updates automatically, similar to buildpack lifecycle
- cnb-oci: user gets a standard OCI image that could be used outside outside of CF as well (ok, a rather artificial benefit as there are other options to build such an image), but no automatic stack updates (need a restage) 

So yes, one cnb lifecycle should be sufficient for today.

toc/rfc/rfc-draft-cnb-lifecycle.md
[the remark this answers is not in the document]
loewenstein: @rkoster @dsboulder @beyhan Should we add a little subsection referring to https://github.com/buildpacks/libcnb/pull/228, the possibility to do the same for `packit` based Paketo buildpacks and referring to #804 for solving it once and for all?

In short, I would probably write this as
> Buildpacks that need credentials from services should be adapted to read from `VCAP_SERVICES` until #804 removes this requirement.
> **Note**: Buildpacks based on `libpak` or `libcnb` already understand credentials from `VCAP_SERVICES` since [libcnb#228](https://github.com/buildpacks/libcnb/pull/228) has been merged and released.

modulo11:
@beyhan we added a section regarding consuming buildpacks from private registries. We did not yet decide on an implementation. Does this require additional reviewers?

beyhan:
> @beyhan we added a section regarding consuming buildpacks from private registries. We did not yet decide on an implementation. Does this require additional reviewers?

When the discussions are resolved I will require review from the involved people any way. Regarding the change could you please commend why the current support for private registries described in https://docs.cloudfoundry.org/devguide/deploy-apps/push-docker.html#registry isn't helpful. I think this comment/explanation could be added also to the RFC

toc/rfc/rfc-draft-cnb-lifecycle.md
loewenstein: @dsboulder would you want to propose a different wording or are we good to resolve this question?

modulo11:
> > @beyhan we added a section regarding consuming buildpacks from private registries. We did not yet decide on an implementation. Does this require additional reviewers?
> 
> @modulo11 When the discussions are resolved I will require review from the involved people anyway. Regarding the change could you please comment why the current support for private registries described in https://docs.cloudfoundry.org/devguide/deploy-apps/push-docker.html#registry isn't helpful or can't be reused. I think this comment/explanation could be added also to the RFC.

I updated the section. It's basically the lack of support for multiple registries.

toc/rfc/rfc-draft-cnb-lifecycle.md
loewenstein: @FloThinksPi I would say you are calling for a separate RFC that's largely independent from the Cloud Native Buildpack topic. Are we good to resolve this question?

toc/rfc/rfc-draft-cnb-lifecycle.md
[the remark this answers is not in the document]
c0d1ngm0nk3y: > Ideally a heading would be added discussion what to do with service bindings.

Should it really? It feels a bit out of place because...

a) This is not necessarily `cf` related, but rather buildpack related
b) This feels like an implementation detail as already established in the predecessor rfc ([here](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0017-add-cnbs.md#service-bindings))

What should we mention here? That `cf` does not have to change?

loewenstein:
@cloudfoundry/toc if @rkoster and @beyhan are ok with https://github.com/cloudfoundry/community/pull/796#discussion_r1570898862, i.e. relying on the [Service Binding section in the older CNB RFC](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0017-add-cnbs.md#service-bindings), there are no pending discussions left I believe.

How do we bring this forward for a vote? I guess it would be good to have explicit buy-in to review the necessary changes in @cloudfoundry/wg-app-runtime-platform (new app lifecycle in the Diego Bosh release) and @cloudfoundry/wg-app-runtime-interfaces (CAPI and CLI changes). Once the vote is done and positive and this RFC merged, we'd need to request the creation of one new repository (for the new app lifecycle) - is there a process for this?

beyhan:
@loewenstein I added your last comment as topic for the next TOC meeting.

toc/rfc/rfc-draft-cnb-lifecycle.md
```diff
@@ -0,0 +1,176 @@
...
+### Goals
+
+- Establish the latest generation of buildpacks in CF as first-class citizen
+- Increase cohesion and app portability between CF Deployment and [Korifi](https://www.cloudfoundry.org/technology/korifi/), via mutual Paketo integration
+- Increase adoption of Cloud Native Buildpacks
+- Open the door for eventual deprecation of the v2 buildpacks, reducing maintenance costs (v2 buildpack deprecation is NOT included in this RFC)
+- No fundamental changes to the architecture of CF
+  - Result of the staging process will be a droplet
+  - No OCI registry necessary
+  - Reuse cflinuxfs4 as rootfs during build and run
```
loewenstein, line 57:
```suggestion
  - Reuse cflinuxfs4 as rootfs during build and run
  - No change to how the Cloud Foundry platform provides service binding information
```

toc/rfc/rfc-draft-cnb-lifecycle.md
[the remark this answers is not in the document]
loewenstein: @beyhan like https://github.com/cloudfoundry/community/pull/796/files#r1574769369?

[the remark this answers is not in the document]
beyhan: yes, thanks!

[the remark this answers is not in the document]
rkoster: > Should it really? It feels a bit out of place because...
Sure now that we have an RFC addressing the problem of servicebinding spec compatibility there is no need to discuss it in this RFC. A link to the service-binding RFC would be nice so that future readers understand how they relate.

beyhan:
We discussed the last commend during the TOC meeting today 23.04.24 and we decided to start the FCP with the goal to approve this RFC next Tuesday 30.04.24.

toc/rfc/rfc-draft-cnb-lifecycle.md
```diff
@@ -0,0 +1,177 @@
...
+- [manifests](https://v3-apidocs.cloudfoundry.org/index.html#manifests)
+
+### Goals
+
+- Establish the latest generation of buildpacks in CF as first-class citizen
+- Increase cohesion and app portability between CF Deployment and [Korifi](https://www.cloudfoundry.org/technology/korifi/), via mutual Paketo integration
+- Increase adoption of Cloud Native Buildpacks
+- Open the door for eventual deprecation of the v2 buildpacks, reducing maintenance costs (v2 buildpack deprecation is NOT included in this RFC)
+- No fundamental changes to the architecture of CF
+  - Result of the staging process will be a droplet
```
natalieparellano, line 55: The CNB lifecycle supports exporting to OCI layout on disk (the resulting image can even be "sparse" in that run image layers will be missing). This should facilitate exporting to a CF droplet without having to re-implement large parts of the exporter.

beyhan:
@loewenstein  Additionally, we discussed that @ameowlia, @rkoster and @stephanme will support the contributions to get approved. You can also open a pr like https://github.com/cloudfoundry/community/pull/730 to add the new lifecycle repository and @ameowlia will help with finding the right place for it.

toc/rfc/rfc-draft-cnb-lifecycle.md
```diff
@@ -0,0 +1,177 @@
...
+  - No OCI registry necessary
+  - Reuse cflinuxfs4 as rootfs during build and run
+  - No change to how the Cloud Foundry platform provides service binding information
```
loewenstein, line 58:
```suggestion
  - No change to how the Cloud Foundry platform provides service binding information (see [#804](https://github.com/cloudfoundry/community/pull/804) for how this will be covered in a separate RFC)
```

loewenstein: @beyhan @rkoster ^^

ameowlia (approved):

stephanme (approved):

ChrisMcGowan (approved):

beyhan (approved):

rkoster (approved):

toc/rfc/rfc-draft-cnb-lifecycle.md
alexbarbato: Not to de-rail - but is this a point worth considering further:

> The stack is saved multiple times in the blobstore(storage and traffic)
This is a core benefit of an OCI registry that it should see the layers and thus save a tremendous amount of space. Re-using the blobstore certainly is simplest, but there are cons on the other end to my eye.

Maybe that concern belongs in the follow on RFC?

loewenstein-sap: I am not sure if moving CF integrated CNBs from creating droplets to creating container images is something we want. If we don't, then there is no problem with a stack saved multiple times, because the roots is saved separately from the droplets.

PR #799, opened 2024-03-26
Enable branch protection automation for ARI WG

stephanme:
- refer to [rfc-0015-branch-protection.md](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0015-branch-protection.md)

### DIFF
--- a/toc/working-groups/app-runtime-interfaces.md
+++ b/toc/working-groups/app-runtime-interfaces.md
@@ -458,4 +458,6 @@ areas:
   - cloudfoundry/notifications
   - cloudfoundry/app-runtime-interfaces-infrastructure
 
+config:
+  generate_rfc0015_branch_protection_rules: true
 ```


silvestre:
@stephanme : Does that mean that our custom main branch protection rule, that also requires checks (GitHub workflows) would be overwritten?

image

stephanme:
@silvestre : Yes, they would be overwritten. But you can specify project specific branch protection rules in [branchprotection.yml](https://github.com/cloudfoundry/community/blob/main/org/branchprotection.yml) to overrule the standard. See also [the docs](https://github.com/cloudfoundry/community/blob/main/org/readme.md#branch-protection-rules).
I suggest to setup those project specific rules before this PR gets merged.

silvestre:
@stephanme : Thanks! 

I assume that only the requirements from RFC 0015:
> only the bots for the Working Group owning the repository and any areas in the Working Group contributing to the repository may push directly to the branch
human contributors must make contributions through a regular pull request workflow.

are binding, and settings like
```
            dismiss_stale_reviews: true
```
from the docs are just a suggestion but not a requirement?

stephanme:
@silvestre: dismiss_stale_reviews=true is not a requirement. For the automation, we had to decide for true or false and thought that true is the more sensible value.

#805 LGTM. If you remove the draft flag, I could merge it once the org automation github action is green again.

silvestre:
#805 is ready for review and merge.

stephanme:
CAPI specific configuration: #819

beyhan (approved):

PR #804, opened 2024-04-04
[RFC] Add Support for File based Service Binding Information

beyhan:
This RFC suggests to introduce an alternative option for the CF platform to provide Service Binding information to applications.

[Preview](https://github.com/cloudfoundry/community/blob/add-file-based-service-binding-information/toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md)


toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
```diff
@@ -0,0 +1,112 @@
...
+  LogSource: "some-log-source",
+  SuppressLogOutput: false,
+}
+```
+
+## Workstreams
+
+### App Runtime Interfaces WG
+
+Cloud Controller should be extended with an App Feature to activate the new file base Service Binding option. If the file-based service binding feature is active the Cloud Controller should generate a Run action, which configures the `VCAP_SERVICES` content to be stored as tmpfs file in the application container instead of environment variable.
```
loewenstein, line 84: This sounds like an implementation of the proposal 1), while the proposal section seems to favour the proposal 2). Is that intentional?

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
```diff
@@ -0,0 +1,112 @@
...
+      * Less implementation effort for the Cloud Controller
+   * Disadvantages:
+      * Can’t make use of tools and libraries from the Cloud Native community because K8s specifies a different file structure and format for the service binding information
+2. Implement the K8s service binding specification. The environment variable `SERVICE_BINDING_ROOT` defines the location for the service bindings. There is a file per service binding. The name of the file and the format follow the [K8s specification](https://servicebinding.io/):
+   * Advantages:
+      * CF community could re-use service binding libraries from the Cloud Native community
+      * Moving application between CF or K8s deployments will be easier
+   * Disadvantages
+      * Higher implementation efforts for the Cloud Controller
+      * Applications with binding information >`130KB` have to go with the file option and adopt it
```
silvestre, line 42: Isn't that a disadvantage of the whole proposal? I don't think it only applies to alternative 2)

beyhan: Thanks! You are right and I tried to address this with https://github.com/cloudfoundry/community/commit/40681233f3aaad7ff80519e114ff60372ce742b4

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
```diff
@@ -0,0 +1,110 @@
...
+2. Implement the K8s service binding specification. The environment variable `SERVICE_BINDING_ROOT` defines the location for the service bindings. There is a file per service binding. The name of the file and the format follow the [K8s specification](https://servicebinding.io/):
+   * Advantages:
+      * CF community could re-use service binding libraries from the Cloud Native community
+      * Moving application between CF or K8s deployments will be easier
+   * Disadvantages
+      * Higher implementation efforts for the Cloud Controller
+
+
+The 2) alternatives offers more than just addressing the issue of this RFC. It suggests an option to evolve the CF platform towards a different service binding specification defined by the Cloud Native community. This means higher implementation efforts for the CF platform and application developers, but possible benefits from the Cloud Native community. This RFC has a light preference for the 2) alternative because of the listed advantages but the feedback of the CF community is wanted here.
```
mogul, line 43: Some app developers are going to want to make the minimum changes possible, and 1 definitely works there. Other app developers are going to porting an app from k8s, and 2 makes their path onto CF easier. Is it possible to offer **both** 1 and 2, and have developers pick which environment variable they'll use to find the file?

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+## Proposal
+
+The CFF community should implement an alternative approach for service binding information based on `tmpfs` file(s). Using a file or files to provide service binding information to applications will address the challenges listed in the problem section because:
+- The file size limit can be controlled by the CF platform
+- Already today CF uses `tmpfs` for [Instance Identity Credentials](https://docs.cloudfoundry.org/devguide/deploy-apps/instance-identity.html) which are rotated without restarting the application every `24h` by default
+
+The two approaches should be supported in parallel. Users should be able to select which approach Cloud Controller should use to deliver the binding information. Applications with binding information >`130KB` have to go with the file option and adopt it. There are two alternatives regarding service binding file organization:
+1. The VCAP_SERVICES content is stored in a file which location is specified via the VCAP_SERVICES_FILE_PATH env var in the same format as the VCAP_SERVICES environment variable
```
mogul, line 29: For consistency with how env vars are referenced below:
```suggestion
1. The `VCAP_SERVICES` content is stored in a file which location is specified via the `VCAP_SERVICES_FILE_PATH` env var in the same format as the `VCAP_SERVICES` environment variable
```

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+## Workstreams
+
+### App Runtime Interfaces WG
+
+Cloud Controller should be extended with an App Feature to activate the new file base Service Binding option. If the file-based service binding feature is active the Cloud Controller should generate a Run action, which configures the Service Bindings to be stored as tmpfs file(s) in the application container instead of `VCAP_SERVICES` environment variable.
+
+Additionally, the suggest limit for the size should be implemented.
```
rkoster, line 85: It would be good to add a section describing how the `name` and `tags` [OSBAPI fields](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#service-offering-object) could be mapped to `type` and `provider` [service binding fields](https://servicebinding.io/spec/core/1.1.0/#example-directory-structure) (if these are not specified by the broker). This could potentially help with initial broker compatibility. This could for example be done based on well known tags like `postgres` and `mysql`. This would be in-line with the[ service-specific code that is currently used to populate the DATABASE_URL variable](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/lib/cloud_controller/database_uri_generator.rb#L3-L7).

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
dmikusa: My read is that users will be able to pick between using the existing env variable approach, which means zero change for their apps, or they can use the file-based approach which would mean some level of change for their app with either approach. So there is control for the user that doesn't want change.

Both of the implementations above sound like there would be work involved. Minimally, you're going to need to know to read from a file instead. It's small, but if you've got apps that are not actively being developed, even that may be more work than you're willing to do.

Not sure if it needs to go into this proposal, but I would suggest that we target the common libraries that folks are using to read `VCAP_SERVICES`, like `java-cfenv` (I believe there are Python & Node.js libraries also) and get support for both formats, env variable and file, and then it's a lot less relevant how it's implemented. The libraries will smooth things over and possibly even make it so that no code change is required at the app level.

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
dmikusa: In libcnb for Cloud-Native Buildpacks, we implemented a translation from `VCAP_SERVICES` to CNB bindings. What we did was this:

1. Credentials map was turned into key/value files (i.e. key is the file name, value is the file contents). We did not go into nested maps, so if a key in Credentials had a nested JSON value, then the JSON is written as the contents of the file.
2. We used the Label from `VCAP_SERVICES` as the `type`.
3. We used the top-level key from the `VCAP_SERVICES` map as the `provider`.

I haven't used it myself, but I know people are using it. @loewenstein - How has this mapping worked out in practice?

loewenstein: We contributed that to `libcnb` mainly because of the Paketo Dynatrace Buildpack and for that it works - as far as I can tell.

beyhan: @loewenstein is https://github.com/buildpacks/libcnb/pull/228 you are referring to? The mentioned mapping above isn't directly visible in the change but most probably I have to dive deeper into it.

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
beyhan: >My read is that users will be able to pick between using the existing env variable approach, which means zero change for their apps, or they can use the file-based approach which would mean some level of change for their app with either approach. So there is control for the user that doesn't want change.

This is correct. @mogul given this clarification do you still see benefits in supporting both options in parallel? 

>Not sure if it needs to go into this proposal, but I would suggest that we target the common libraries that folks are using to read VCAP_SERVICES, like java-cfenv (I believe there are Python & Node.js libraries also) and get support for both formats, env variable and file, and then it's a lot less relevant how it's implemented. The libraries will smooth things over and possibly even make it so that no code change is required at the app level.

@dmikusa Do you mean here option 1) or both? Most probably both but I wanted to check?

dmikusa: Both. I think regardless of what we do, we can lean on the libraries to help with smoothing things over for hopefully a lot of users.

UlfFildebrandt: Just to comment on the reason for both options: There is a problem with the length of VCAP and running with existing VCAP + K8s service binding would not simplify to use the libraries to get the variables. So, to run both alternatives in parallel, I would propose to store the variables in both formats in two different files, not only keep the current VCAP in memory. This would solve the existing problem and also prepare for future service bindings.

mogul: >  @mogul given this clarification do you still see benefits in supporting both options in parallel?

Nope, I hadn't taken into consideration that `VCAP_SERVICES` wasn't going away. I think it's OK to say "if you want better credential rotation or a larger set of bindings, then you have to use this new method which will require understanding `SERVICE_BINDING_ROOT`".

dmikusa: If a user is hitting the length limit of `VCAP_SERVICES`, I would make the case that the user should then switch to the CNB bindings option which wouldn't suffer from that problem.

If you're using a library to parse `VCAP_SERVICES`, the idea would be that you can toggle between the two implementations (VCAP ENV or CNB) and your app wouldn't notice the difference. That should be achievable, at least with the Java/Spring libraries I've worked on.

Even if a user is not using a library and can't for some reason, both proposed implementations will require changes to the user's app, 1.) to read the file name from env variable, then read VCAP contents from a file, or 2.) read binding info from CNB binding files. Again, why not just read in the CNB bindings?

For those unfamiliar with the bindings spec, implementing it in a client only requires reading a couple of files from a directory. It's straightforward to find and read dependency data. I think even apps that have profile scripts that might `jq` some info out of `VCAP_SERVICES` should be able to switch and do something similar with a bit of bash script (it might not be a one-liner, but maybe two or three lines).

mogul: > There is a problem with the length of VCAP and running with existing VCAP + K8s service binding would not simplify to use the libraries to get the variables. 

If their logic was to look for `SERVICE_BINDING_ROOT` first, and only fallback to reading from `VCAP_SERVICES` if it's not present, I think libraries would simplify this.

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
beyhan: Looking into the implementation of the CC the top-level key from the `VCAP_SERVICES` [seems to be also the label](https://github.com/cloudfoundry/cloud_controller_ng/blob/b8c6dbef8bc469d0f0dca6de77a482b2038b3dac/app/presenters/system_environment/system_env_presenter.rb#L15-L22) from the `VCAP_SERVICES`. The value is taken from the `name` attribute of the service offering as defined in the OSBAPI [spec](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#catalog-management). I think it make sense to be consistent and have the same mapping as implemented in `libcnb` and will update the RFC with this if we agree to go with the 2) approach and don't here anything against the suggested mapping.

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
beyhan: I fully agree with the last commend provided by @dmikusa:
> If a user is hitting the length...

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
```diff
@@ -0,0 +1,111 @@
...
+- Updates of the Service Binding Information require restage. This is not optimal for an eventual support of [Binding rotation](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#binding-rotation) specification from the OSBI API spec.
+
+
+## Proposal
+
+The CFF community should implement an alternative approach for service binding information based on `tmpfs` file(s). Using a file or files to provide service binding information to applications will address the challenges listed in the problem section because:
```
blgm, line 25: While being generally very supportive of this idea, it's important to consider the permissions of this file. Currently, when integrated with CredHub, Diego looks up service credentials just before starting an App and injects them into the App's environment, so that only the App (and not anyone connecting via `cf ssh`) can see the credentials in plaintext. If plaintext credentials are moved to a file, then it's important to think about who can see this file. Is it just the App, or should `cf ssh`, `cf run-task`, etc... see them? On my system both Apps and `cf ssh` run with `uid=2000(vcap)`, so any files that an App can read are also readable by `cf ssh` or `cf run-task` users.

beyhan: Thanks @blgm for this important point and I also agree that these file should bring the same security level as we have today with `VCAP_SERVICES`. I wonder about:
>Currently, when integrated with CredHub, Diego looks up service credentials just before starting an App and injects them into the App's environment, so that only the App (and not anyone connecting via cf ssh) can see the credentials in plaintext.

When I do `cf ssh` into an app and execute the command `env` I can see the content of the `VCAP_SERVICES` env var or any env var Diego configured for the app container. With that I can see the credentials which are in plain text in `VCAP_SERVICES` services with `cf ssh` already today or do I miss something here?

rkoster: https://github.com/cloudfoundry/buildpackapplifecycle/blob/main/credhub/credhub.go#L26-L45
At least for buildpack based apps this seems to be controlled by the `CREDHUB_SKIP_INTERPOLATION` environment variable.

dmikusa: I haven't tried in a while, but I'm guessing you can still `cat /proc/<pid>/environ [see](https://www.baeldung.com/linux/process-environment-variables) to view the values, even as careful as CredHub/Diego are, I'm pretty sure you can still see those values.

+1 for secure file permissions, probably `600` owned by the uid running the app.

I don't think you can block someone sufficiently motivated from seeing them if they have `cf ssh` access though.

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
beyhan: @loewenstein can we resolve this?

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Possible Future Work
+
+The App Features API aren’t supported currently in the CF CLI and [app manifest](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html). To make the use of this proposal for CF operators easier this should be addressed.
+
+### App Manifest Attributes Proposal
```
stephanme, line 95: Adding app features to the cf app manifest has been requested before for enabling ssh: https://github.com/cloudfoundry/cloud_controller_ng/issues/2996

The proposal should allow to explicitly enable and disable a feature flag (i.e. not just by listing the flag). Otherwise it would be impossible to disable a feature flag because the CF app manifest is not declarative but additive (see https://v3-apidocs.cloudfoundry.org/#apply-a-manifest-to-a-space).

E.g.
```
applications:
- name: testapp
  features:
  - file-based-service-bindings: true
```
or, closer to the [app-features v3 API](https://v3-apidocs.cloudfoundry.org/#app-features), 
```
applications:
- name: testapp
  features:
  - name: file-based-service-bindings
    enabled: true
```

beyhan: I agree and tried to address this with https://github.com/cloudfoundry/community/pull/804/commits/0ae14081c9f2febb8506f8cdbaf99bbb6ab8ec01. Is that OK?

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
beyhan: I would like to resolve this so that the RFC could be updated with the option(s) we want to implement. That is why please vote:

| Option 1    | Option 2 | Both | Abstain |
| -------- | ------- |------- |------- |
| :+1:  | :thumbsdown:  | :rocket: | :eyes: |

cc TOC: @stephanme, @rkoster, @ChrisMcGowan, @ameowlia

loewenstein: Should we consider :one:, :two:, :three: and :four: as kinda neutral alternatives to the vote indicators you proposed. No strong opinion, but e.g. 👎 seems to have a negative connotation and 🚀 seems to signal a very clear preference...

;)

beyhan: Ok, let's do this. I thought about this but also wanted to make it easier without a need of commenting :-)

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
blgm: Yes, it's definitely resolved from my perspective.

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
beyhan: Just to clarify that the voting is open also for others not only for the TOC.

loewenstein: :two:

blgm: 2️⃣

rkoster: 2️⃣

pivotal-marcela-campo: 2️⃣

mogul: 2️⃣

PlamenDoychev: 2️⃣

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
```diff
@@ -0,0 +1,123 @@
...
+  -	file-based-service-bindings: true
+```
+or as alternative proposal:
+
+```
+---
+applications:
+- name: test-app
+  features:
+  -	file-based-service-bindings
```
stephanme, line 113: "name" is missing before file-based-service-bindings in the alternative proposal:

```
features:
  - name: file-based-service-bindings
    enabled: true
```

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
stephanme: From concept point of view, it looks ok. But there is one yaml error in the 2nd proposal (see my comment below).

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
stephanme: :two:

ameowlia: 2️⃣

ChrisMcGowan: 2️⃣

beyhan: Thank you for the votes! I didn't set an end date for the voting but most of the RFC reviewers and the TOC already voted and the winner is clear. That is why, I updated the RFC (https://github.com/cloudfoundry/community/pull/804/commits/88f5fb6e54b3c292396d5b56157821d5b7133e04) with the winning option. Please review the latest state.

toc/rfc/rfc-draft-add-support-for-file-based-service-binding.md
beyhan: I added details about the `VCAP_SERVICES` -> K8s bindings mapping with https://github.com/cloudfoundry/community/pull/804/commits/88f5fb6e54b3c292396d5b56157821d5b7133e04. Please review and provide feedback.

dmikusa (approved):

PlamenDoychev (approved):

loewenstein (approved):

ChrisMcGowan (approved):

beyhan:
We are starting the FCP with the goal to approve this RFC by next Tuesday 7th of May.

stephanme (approved):

PR #807, opened 2024-04-09
[RFC] Option to tunnel CF SSH through WebSockets

domdom82:
This RFC proposes an option in Cloud Foundry that would allow the CF SSH feature to be tunnelled through a WebSocket connection. The benefit would be that the regular HTTP port 443 could be reused and the (often complained about) additional port 2222 could be closed on the network edge.

[Easy viewing](https://github.com/sap-contributions/cf-community/blob/rfc-ssh-ws/toc/rfc/rfc-draft-ssh-over-ws.md)


domdom82:
@cloudfoundry/toc @ameowlia @geofffranks @PlamenDoychev @beyhan @stephanme @maxmoehl @plowin @peanball I've opened a new RFC for piping CF SSH through WebSockets instead of port 2222. Please comment / review.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,269 @@
...
+
+This RFC aims at providing an additional way of using CF SSH in an effort to reduce the potential attack surface against SSH.
+
+By tunneling SSH through a WebSocket connection, the SSH client may reuse the same port as regular HTTP traffic, removing the need for port 2222.
+
+## Problem
+
+Cloud Foundry is a shared platform.
+
+In production deployments of Cloud Foundry, it is commonplace for customers to run security scans on their workloads to ensure compliance and reduce attack surfaces.
```
PlamenDoychev, line 20: Perhaps commonplace => common place

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,269 @@
...
+
+These scans usually complain that besides port 443 for HTTP traffic, also port 2222 is open for SSH. Many customers demand closure of the port as they
```
PlamenDoychev, line 22: I am not sure if it is worth mentioning that the ssh is not enabled by default. Manual action is required in order to allow ssh, sadly most of the customers later forget to close it and execute scans.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,269 @@
...
+		wg.Done()
+	}()
+
+	wg.Wait()
+	_ = wsConn.Close()
+}
+```
+Source of `ws-proxy` used below
+
+It MAY be used with regular ssh like so:
```
maxmoehl, line 266:
```suggestion
Which can be used with regular ssh like so:
```

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+
+### Changes needed for regular SSH, SCP and SFTP
+
+- Getting the ssh token via `cf ssh-code` remains unchanged as it connects to UAA.
+- Regular ssh, scp and sftp commands SHOULD support a `-o ProxyCommand` option.
+- The program run as `ProxyCommand` MUST read from `stdin` and write to `stdout` for ssh to work.
+- The CF community MAY offer a simple WebSocket proxy for regular ssh as a [homebrew formular](https://github.com/cloudfoundry/homebrew-tap) or similar option.
+- A simple WebSocket proxy MAY look like this:
```
maxmoehl, line 230:
```suggestion
- A simple WebSocket proxy could look like this:
```

This is not an optional specification as part of this RFC but rather a supplement / example to guide implementers, as such it shouldn't use `MAY`.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,269 @@
...
+          server_cert_domain_san: ssh.((system_domain))
+          uris:
+          - ssh.((system_domain))       # e.g. ssh.cf-app.com, announced by CAPI as wss://ssh.cf-app.com
+    release: routing
+```
+
+
+### Changes needed for regular SSH, SCP and SFTP
+
+- Getting the ssh token via `cf ssh-code` remains unchanged as it connects to UAA.
```
maxmoehl, line 226:
```suggestion
- Getting the ssh token via `cf ssh-code` MUST remain unchanged as it connects to UAA.
```

toc/rfc/rfc-draft-ssh-over-ws.md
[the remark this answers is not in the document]
peanball: do you mean exposing a "local SSH port" that you could use to connect to via stock `ssh`?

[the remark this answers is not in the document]
maxmoehl: Using the `ProxyCommand` option is fine, but I'd prefer it if you can setup the tunnel with a command that most systems already have installed. The middle ground would probably be to add the tunnel functionality to the cf-cli as most (if not all) cf users probably have that installed.

toc/rfc/rfc-draft-ssh-over-ws.md
domdom82: I've checked it. Looks correct now.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,269 @@
...
+- Regular ssh, scp and sftp commands SHOULD support a `-o ProxyCommand` option.
```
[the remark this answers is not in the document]
domdom82: > I would even go a bit further and say that the CF CLI must add support for cf proxy-ssh or whatever.

I had the same thought initially, but I figured people who use regular ssh instead of cf ssh usually don't have the CF CLI installed, so a `cf proxy-ssh` command wouldn't make much sense. But I am open for putting it in two places: 
- CF CLI
- separate binary

toc/rfc/rfc-draft-ssh-over-ws.md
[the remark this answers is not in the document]
domdom82: curl can be massaged into opening a WebSocket connection. We could try using it as a proxy, too. Fair point.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,269 @@
...
+          tls_port: 443 # or 9443 if preferred
+          registration_interval: 20s
+          server_cert_domain_san: ssh.((system_domain))
+          uris:
+          - ssh.((system_domain))       # e.g. ssh.cf-app.com, announced by CAPI as wss://ssh.cf-app.com
+    release: routing
+```
+
+
+### Changes needed for regular SSH, SCP and SFTP
```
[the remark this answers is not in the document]
domdom82: Agreed.

toc/rfc/rfc-draft-ssh-over-ws.md
domdom82: Agreed.

toc/rfc/rfc-draft-ssh-over-ws.md
domdom82: Agreed.

toc/rfc/rfc-draft-ssh-over-ws.md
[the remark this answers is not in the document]
maxmoehl: For SCP you can't use the CF CLI. In such cases most users probably use the cf cli to generate the token which is then (manually) passed on to scp. Having the proxy built-in to the CLI would help in that case as it avoids the need to install a separate binary.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### Scenario C: WS only
+This scenario MAY be enabled by operators. It turns of the legacy scenario A and allows operators to close the public port 2222 on the load balancer. The CF CLI MUST understand the new flow using a WebSocket tunnel. Other components MUST support the new flow and MAY support both scenarios.
+
+
+### Changes per Component (Scenario B and C)
+
+#### CF CLI
+- The CLI MUST query the "v2/info" endpoint of Cloud Controller.
+- The endpoint MUST return the legacy SSH info:
```
peanball, line 85: If you were to configure that you don't want the legacy SSH in your CF environment, would it not make sense to not have it anymore? By default I would say that this should still be there, but if you opt out of 'classic CF SSH', that info should not be there.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,270 @@
...
+- The rest of CF CLI SHOULD be unchanged and work as before.
+
+#### CAPI
+
+- CAPI MAY provide an additional `app_ssh_ws_endpoint` field in the "v2/info" endpoint:
+```
+  "app_ssh_ws_endpoint": "wss://ssh.cf-app.com"
+```
+
+- The URL presented in the field MUST be `wss://` + the route as announced by route-registrar for SSH proxy (see [CF Deployment](#cf-deployment))
```
peanball, line 168: would it be useful to have an addressing scheme that is directly related to the app's route?

e.g. `cf ssh wss://myapp.cf-app.com`? The fact that this is an SSH tunnel request could be indicated via separate header. Ultimately the websocket communication will end up in gorouter, so gorouter can decide that this is actually a web service request. Finally, we also need to communicate the instance ID that we want to connect to, which would also be in a header.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,270 @@
...
+
+(...)
+```
+Changes in [proxy.go](https://github.com/cloudfoundry/diego-ssh/blob/main/proxy/proxy.go#L67)
+
+**Remark:**
+(these changes are not meant as verbatim code; they only illustrate areas that would need to change)
+
+
+- SSH Proxy MUST retain the legacy SSH server on port 2222 for backwards compatibiliy
```
peanball, line 196: As noted, I would see an opt-in to disable this port.

toc/rfc/rfc-draft-ssh-over-ws.md
maxmoehl: The web socket tunnel breaks with some of the well-known concepts. Previously there was just one SSH server, the ssh-scheduler (or whatever it's called) and it decides to which app you will be connected based on the user (see also the [docs](https://docs.cloudfoundry.org/devguide/deploy-apps/ssh-apps.html#other-ssh-access)). Now we have this web socket tunnel and in theory you still have to say where you want to go once you exit the tunnel, but you don't want to go anywhere. So the real target host for ssh is probably going to be `localhost`?

`ssh -o ProxyCommand=cf-ws-tunnel cf:${process_id}/0@$host`

* `cf-ws-tunnel` would talk to the CC and get `wss://ssh.cf.example.com` as the WebSocket endpoint which points to the SSH server (registered via route-registrar like any other route is).
* `cf:${process_id}/0` is the same as it is right now, tells the SSH server which process to access.
* `$host`: it's unclear to me what that value would be, I'm suspecting it has to be `localhost` or it doesn't matter at all since the web socket tunnel will be piped to the SSH server no matter what.

It becomes this weird hybrid of a tunnel and an SSH connection which both are taking on responsibilities of the other.

toc/rfc/rfc-draft-ssh-over-ws.md
domdom82: Yes. I'm weakening the requirement to SHOULD. The CLI must be able to cope with both cases.

toc/rfc/rfc-draft-ssh-over-ws.md
domdom82: I disagree. The point of this feature request is to access SSH proxy via a WebSocket tunnel instead of a direct SSH connection. Changing the way CF SSH works in general is out of scope of this RFC. For gorouter, this will be just another WebSocket request without any special handling needed.

toc/rfc/rfc-draft-ssh-over-ws.md
domdom82: Agreed as above changing to SHOULD. I will also add a statement that the code for the legacy server shall not be removed.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Summary
+
+This RFC aims at providing an additional way of using CF SSH in an effort to reduce the potential attack surface against SSH.
+
+By tunneling SSH through a WebSocket connection, the SSH client may reuse the same port as regular HTTP traffic, removing the need for port 2222.
+
+## Problem
+
+Cloud Foundry is a shared platform.
```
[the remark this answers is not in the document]
domdom82: Agreed.

toc/rfc/rfc-draft-ssh-over-ws.md
domdom82: Added more clarity in the document.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,274 @@
...
+This scenario MAY be enabled by operators. It allows legacy SSH as well as pure WebSocket-based SSH in one environment. The CF CLI MUST understand the legacy flow from scenario A. The CF CLI SHOULD understand the new flow using a WebSocket tunnel. Other components MUST support both scenarios.
+
+#### Scenario C: WS only
+This scenario MAY be enabled by operators. It turns of the legacy scenario A and allows operators to close the public port 2222 on the load balancer. The CF CLI MUST understand the new flow using a WebSocket tunnel. Other components MUST support the new flow and MAY support both scenarios.
+
+
+### Changes per Component (Scenario B and C)
+
+#### CF CLI
+- The CLI MUST query the "v2/info" endpoint of Cloud Controller.
```
stephanme, line 86: CLI v7 and v8 use the `/` endpoint of CC to learn about the foundation.
The ssh info looks like this:
```
{
  "links": {
    "app_ssh": {
      "href": "ssh.cf-app.com:2222",
      "meta": {
        "host_key_fingerprint": "...",
        "oauth_client": "ssh-proxy"
      }
    },
```

The `v2/info` endpoint belongs to the deprecated CF API v2. It shall not be used anymore and new features won't get implemented in v2.
CLI v6 is based on CF API v2 (and uses the `v2/info` endpoint) but CLI v6 is out of support: https://github.com/cloudfoundry/cli/wiki/Versioning-and-Support-Policy

I suggest to base this RFC on CF API v3 and CLI v8 only.

domdom82: hi @stephanme thanks for your input. I think I've misread the code looking for where the CLI gets the SSH endpoint from. There is a [getSSHEndpointInfo](https://github.com/cloudfoundry/cli/blob/main/cf/commands/application/ssh.go#L178) function that's still in the main branch of the CLI. However, I've tested the v8 CLI with `-v` and it does indeed call the `/` endpoint of CC.

I will update the RFC accordingly, thanks!

domdom82: Pushed a commit to change the compatibility level to v3 only.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,329 @@
...
+### Architecture Changes
+
+![architecture-changes](rfc-draft-ssh-over-ws/cf-ssh-over-websocket.drawio.png)
+
+There are three scenarios:
+
+#### Scenario A: Legacy SSH
+This is the default scenario as of today. All changed components MUST ensure this scenario still works. It SHOULD be tested with legacy CLI versions such as v6 which are no longer updated. This scenario includes regular SSH clients like OpenSSH.
+
+#### Scenario B: Dual SSH / WS
```
plowin, line 76: Which scenarios do you have in mind to also support the mixed mode?
I currently do not see the benefit for an operator; if the port cannot be closed, are the additional switches worth the gain of partial websocket usage?

domdom82: buying time for legacy clients basically. you want to migrate customers off the 2222 port while also allowing frontrunners to use the new feature. the mixed mode allows you to do a gradual shift instead of a big bang.
if you can't run mixed you won't have any time to collect feedback ahead of time and cannot easily roll back in case of any show stoppers that might turn up.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,329 @@
...
+	c, chans, reqs, err := ssh.NewClientConn(wsConn, url, config)
+	if err != nil {
+		return nil, err
+	}
+	sshClient := ssh.NewClient(c, chans, reqs)
+	return &secureClient{client: sshClient}, nil
+}
+```
+
+- The rest of CF CLI SHOULD be unchanged and work as before.
```
stephanme, line 209: Question: Local port forwarding (e.g. to a DB service or to JMX port of my Java app) would be supported as well, correct?
Example: `cf ssh -N -L 5000:localhost:5000 myapp`

toc/rfc/rfc-draft-ssh-over-ws.md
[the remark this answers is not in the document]
domdom82: Update: It seems that curl can only be used for the initial websocket handshake but does not support WebSocket frame sending while [libCurl](https://curl.se/docs/websocket.html) does support WS/WSS protocols.
There are other tools like [websocat](https://github.com/vi/websocat) could probably work but they are non-standard on most Linux distros (there is no package on Ubuntu / Debian for example, though there is a brew formula).

It begs the question if it won't be overkill to plug such a tool instead of providing just 10 lines of go code or a CLI plugin binary for the same purpose. It's a fair point, though and we should not dismiss it.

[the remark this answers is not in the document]
rkoster: I have no idea how they are doing it but github supports ssh over port 443: https://docs.github.com/en/authentication/troubleshooting-ssh/using-ssh-over-the-https-port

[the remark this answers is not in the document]
rkoster: Nevermind they are just using a different host (ssh.github.com). Which CF already does via `app_ssh_endpoint`. Not sure if it currently is configurable but it should be trivial to change from port 2222 to 443. Scanners would however probably still detect the protocol and complain about that. So this RFC is not so much about ports, but more about wrapping SSH in a WebSocket so scanners won't be able to detect the use of the SSH protocol.

toc/rfc/rfc-draft-ssh-over-ws.md
maxmoehl: This is a good additional example of why interoperability with OpenSSH is crucial (see the [parallel discussion](https://github.com/cloudfoundry/community/pull/807#discussion_r1559013471)).

I think such cases should be supported via a tunnel command, e.g.:

```
ssh -o ProxyCommand='cf tunnel-ssh my-app' ... ???@???
```

Otherwise we'd have to re-implement major parts of the OpenSSH client which I don't think is reasonable.

Users could configure their ssh client to automatically use the proxy command when connecting to an app to avoid having to type out the options each time.

toc/rfc/rfc-draft-ssh-over-ws.md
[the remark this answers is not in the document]
maxmoehl: Technically, it would probably be possible to distinguish the two protocols based on the first few bytes, for SSH the client will send an identification string according to [RFC4253](https://datatracker.ietf.org/doc/html/rfc4253#section-4.2):

```
SSH-protoversion-softwareversion SP comments CR LF
```

whereas a HTTP client will send a TLS handshake which is not as stable but the first bytes will look something like [this](https://tls12.xargs.org/#client-hello/annotated) (in hex):

```
16 03 01 00 a5
```

But from an operator perspective this is probably an absolute nightmare as it adds complexity 😄

It would be interesting to see how scanners deal with this. The SSH protocol states that both sides have to send the identification and when the server distinguish the protocol based on the clients first bytes it would be relatively hard to detect. The scanner would have to deliberately try to set up a SSH connection on port 443 to trigger this, otherwise the server would not announce it as it is waiting for the client to make the first move.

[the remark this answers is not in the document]
domdom82: @rkoster the point of this RFC is to *not* change `app_ssh_endpoint` because that would be an incompatible change. I'm proposing to leave `app_ssh_endpoint` as it is, untouched. Additionally, the CC will offer a second endpoint that uses a `wss://` scheme to indicate the WebSocket protocol. But you are right in that the RFC is not so much about ports, though the implementation would allow you to ultimately close the 2222 port completely.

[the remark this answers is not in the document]
domdom82: @maxmoehl it *could* probably be done using a lib like [cmux](https://github.com/soheilhy/cmux) which allows you to peek into a connection and multiplex to different protocols based on a matcher. This logic would have to be implemented in Gorouter and bypass any of the usual "lookup-route" etc. logic for the SSH case. 

Gorouter would have to peek and see if the incoming TCP connection tries a TLS handshake, a regular HTTP request or an SSH handshake. It wouldn't work with ssh clients that wait for the server to announce the SSH id string, though.

I would classify this as more of an academic point, not something we would really want for CF.

toc/rfc/rfc-draft-ssh-over-ws.md
domdom82: Yes, the features of SSH, including port forwarding should be unaffected as the WebSocket is handled transparently in the `ProxyCommand` option. Similar to what you would get with any other proxy like SOCKS5 etc.
The SSH protocol does not "know" it's being run over a proxy.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,329 @@
...
+
+the SSH proxy greets with the SSH-2.0 version, allowing the attacker to try all kinds of SSH-2.0 related probes.
+
+Another facet of the problem is that many customers require explicit allow-listing of external web sites from their on-prem environments.
+Port 443 is usually not a problem as many firewalls open it by default, but port 2222 likely is frowned upon and cause for concern.
+
+## Proposal
+
+### Note on Security
+The proposal does not aim to solve security issues of (CF) SSH per se. Tunneling a protocol through another does not make it inherently more secure. Any weakness of SSH within the CF SSH proxy will still be present even if the RFC is accepted. The main goal is to reduce the number of open ports required. Although, there is one minor security improvement as the SSH traffic will then be also encrypted by TLS within the WebSocket tunnel.
```
winkingturtle-vmw, line 64: If I understand this correctly, this proposal doesn't solve the problem mentioned above. Should the problem statement be something else to reflect what the proposal is solving?

domdom82: @winkingturtle-vmw It does improve on the SSH port sticking out and waiting to be exploited. I've added this section to manage the expectation that somehow removing the port will make SSH more secure, which it does not. If there is an attack possible on SSH the same attack will work on SSH over WebSockets. 
But I do get your point, any proposal for a change in the problem statement?

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,329 @@
...
+
+In production deployments of Cloud Foundry, it is commonplace for customers to run security scans on their workloads to ensure compliance and reduce attack surfaces.
+
+These scans usually complain that besides port 443 for HTTP traffic, also port 2222 is open for SSH on the internet-facing load balancer. Many customers demand closure of the port as they
+view it as a potential attack surface for brute-force attacks against the SSH protocol.
+
+The issue is exacerbated by features like "bring-your-own-domain" where customers provide a CNAME entry alongside a matching x509 certificate that points their own (sub-)domain to an app hosted on the shared Cloud Foundry environment.
+
+This can cause potential reputation damage to the customer because to an outsider, it looks like the port 2222 is open on the customer's domain, e.g. `www.big-corp.com:2222` open looks much worse than `my-app.cfapps.eu10.hana.ondemand.com:2222` even though both are technically the same.
+
```
rkoster, line 30: Ssh ingress has been hard-coded to be handled by a different host: https://github.com/cloudfoundry/capi-release/blob/main/jobs/cloud_controller_ng/templates/cloud_controller_ng.yml.erb#L126

Typically handled by a different load balancer. So I don't understand how custom domains are affected.

domdom82: That may be a SAP-thing. We usually use one LB for all of the platform ingress and let customers point their DNS to that LB, so implicitly all open ports for that LB will show up on the custom domain. It could be done differently, true, I will change the section to reflect that.

domdom82: I've added a section to reflect different options for ingresses. Both have advantages and costs. I've also removed SAP-specific references.

toc/rfc/rfc-draft-ssh-over-ws.md
winkingturtle-vmw: I think the problem you are trying to solve has to do with SSH protocol being discoverable. With this proposal, we are hiding away that port from being discoverable.

domdom82: thank you. I've added a statement on both the problem as well as the solution sections to emphasize that.

beyhan:
We discussed this RFC during the TOC meeting on 23.04.24 and decided to start FCP with the goal to accept it next Tuesday 30.04.24.

toc/rfc/rfc-draft-ssh-over-ws.md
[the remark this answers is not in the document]
domdom82: Closing this discussion as resolved. Feel free to reopen if needed.

toc/rfc/rfc-draft-ssh-over-ws.md
```diff
@@ -0,0 +1,269 @@
...
+## Proposal
+
+### Note on Security
+The proposal does not aim to solve security issues of (CF) SSH per se. Tunneling a protocol through another does not make it inherently more secure. Any weakness of SSH within the CF SSH proxy will still be present even if the RFC is accepted. The main goal is to reduce the number of open ports required.
+
+
+### Architecture Changes
+
+![architecture-changes](rfc-draft-ssh-over-ws/cf-ssh-over-websocket.drawio.png)
```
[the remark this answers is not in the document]
domdom82: Resolving after discussing both options. We will leave it as one image unless more reviewers are in favor of multiple images.

ChrisMcGowan (approved):

ameowlia (approved):
Thank you for putting together this RFC. I am fine with this as long as we keep both configuration options with the default being the old way. 

Once this is rolled out and we've seen it in action, let's chat again before removing the old way.

stephanme (approved):

rkoster (approved):

beyhan (approved):

PR #815, opened 2024-04-18
RFC: Distribute official CF container images using Github Packages instead of DockerHub

tcdowney:
### Update
After some discussion we feel the best way forward is for teams to continue using DockerHub, but via working-group associated accounts that make it easier for collaboration across organizations.

See this comment:
https://github.com/cloudfoundry/community/pull/815#issuecomment-2085723716

---

Teams have recently been dealing with some toil around the accounts that we use to publish images to DockerHub and this has got us rethinking our usage of it a bit. In conversations on [Slack](https://cloudfoundry.slack.com/archives/C02LR0XCV3M/p1713421244243789?thread_ts=1706014252.635429&cid=C02LR0XCV3M) as well as synchronously with members of the community (@ameowlia and @rkoster 😁) it has been suggested that we switch over to Github Container Registry or Google Artifact Registry.

This is simple enough for images that are used internally by the teams themselves, but a bit more impactful for images that are meant to be consumed by the CF community. Images like [cloudfoundry/cli](https://hub.docker.com/repository/docker/cloudfoundry/cli/general) or [cloudfoundry/cflinuxfs4](https://hub.docker.com/repository/docker/cloudfoundry/cflinuxfs4/general). Some of these were [explicitly requested](https://github.com/cloudfoundry/cli/issues/1578) by the community, after all.

The purpose of this draft RFC is to provide a place to discuss this issue further and consider moving off of DockerHub in a way that causes minimal disruption for the community.

[Readable Preview](https://github.com/cloudfoundry/community/blob/f07c1d71a26d6ca92511fb31c0f443d315d7b7c1/toc/rfc/rfc-draft-use-github-packages-container-registry-instead-of-dockerhub.md)

### DIFF
--- a/toc/rfc/rfc-draft-use-github-packages-container-registry-instead-of-dockerhub.md
+++ b/toc/rfc/rfc-draft-use-github-packages-container-registry-instead-of-dockerhub.md
@@ -1,0 +1,37 @@
+# Meta
+[meta]: #meta
+- Name: Distribute Public Community Images using Github Packages Container Registry instead of DockerHub
+- Start Date: 2024-04-18
+- Author(s): @tcdowney
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (fill in with PR link after you submit it)
+
+
+## Summary
+
+Move away from DockerHub and distribute official Cloud Foundry container images using the [Github Packages](https://github.com/features/packages) [Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry). This RFC is primarily concerned with public container images that are published for the Cloud Foundry community.
+
+## Terminology
+
+* **Community Image -** A public container image that is intended to be consumed by members of the Cloud Foundry community.  Some examples are the `cloudfoundry/cli` and `cloudfoundry/cflinuxfs4` images that are published by the CLI and Buildpacks teams.
+* **Team Image -** A container image that a team publishes for its own internal use (e.g. for use in CI). These images may or may not be public.
+
+## Problem
+
+Currently some working groups distribute container images for consumption by the Cloud Foundry community through the [CloudFoundry DockerHub org](https://hub.docker.com/repositories/cloudfoundry).
+
+Managing DockerHub accounts adds additional overhead to the release processes for these teams. By switching to using Github Packages, teams can use their existing Github accounts to publish images.
+
+It is also unclear what the long-term costs of using DockerHub will be. I believe that the `cloudfoundry` org is currently covered by the [Docker-sponsored Open Source program](https://docs.docker.com/trusted-content/dsos-program/), but [Github Package usage is free for public packages](https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages#about-billing-for-github-packages).
+
+## Proposal
+
+### Deprecation Period
+A deprecation period of 3 months *MUST* be communicated through the cf-dev mailing list and via the READMEs of the images on DockerHub. References to the DockerHub images *MUST* be updated in documentation and code repository READMEs. Community images will continue to be published to DockerHub during this period. Old Community Images *MUST* not be deleted, but the floating tags (i.e. `latest`) *MUST* be deleted to alert consumers that they are no longer using updated versions of the images.
+
+### Community Images
+Teams *MUST* begin publishing new Community Images to the Cloud Foundry Organizations [Github Packages repository](https://github.com/orgs/cloudfoundry/packages) and during the deprecation window teams *MUST* continue to publish images to DockerHub. Once the deprecation period has ended teams *MUST* no longer publish new Community Images to DockerHub. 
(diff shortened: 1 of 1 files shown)


toc/rfc/rfc-draft-use-github-packages-container-registry-instead-of-dockerhub.md
```diff
@@ -0,0 +1,37 @@
...
+* **Team Image -** A container image that a team publishes for its own internal use (e.g. for use in CI). These images may or may not be public.
+
+## Problem
+
+Currently some working groups distribute container images for consumption by the Cloud Foundry community through the [CloudFoundry DockerHub org](https://hub.docker.com/repositories/cloudfoundry).
+
+Managing DockerHub accounts adds additional overhead to the release processes for these teams. By switching to using Github Packages, teams can use their existing Github accounts to publish images.
+
+It is also unclear what the long-term costs of using DockerHub will be. I believe that the `cloudfoundry` org is currently covered by the [Docker-sponsored Open Source program](https://docs.docker.com/trusted-content/dsos-program/), but [Github Package usage is free for public packages](https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages#about-billing-for-github-packages).
```
tcdowney, line 25: @christopherclark my knowledge on this could be stale, so please let me know if we're still covered by their DSOS program.

beyhan:
Thanks for this RFC @tcdowney. Could you please provide an easy to read link in the description like [here](https://github.com/cloudfoundry/community/pull/796) e.g..

beyhan:
@jpalermo you mentioned some points about the Github Packages during the last FI WG meeting. FYI In case you have feedback for this RFC.

jpalermo:
The first time publishing of an image was a bit weird because the image in the Github container registry wasn't linked to any Github repo. There are ways of getting it to link directly using Tags, but we manually associated it after it was pushed the first time and everything seems to be fine now.

So maybe some odd initial configuration, but being able to leverage team permissions is very nice.

toc/rfc/rfc-draft-use-github-packages-container-registry-instead-of-dockerhub.md
```diff
@@ -0,0 +1,37 @@
...
+- RFC Pull Request: (fill in with PR link after you submit it)
+
+
+## Summary
+
+Move away from DockerHub and distribute official Cloud Foundry container images using the [Github Packages](https://github.com/features/packages) [Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry). This RFC is primarily concerned with public container images that are published for the Cloud Foundry community.
+
+## Terminology
+
+* **Community Image -** A public container image that is intended to be consumed by members of the Cloud Foundry community.  Some examples are the `cloudfoundry/cli` and `cloudfoundry/cflinuxfs4` images that are published by the CLI and Buildpacks teams.
```
stephanme, line 16: Do we have a list of community images that require a more formal migration process? E.g. in the CF docs?

ameowlia:
I've been debating this for a bit. It would be nice to having everything in one spot and it would be great if that spot was github given how the community permissions work. But ultimately I don't want to force teams to do unnecessary work migrating images for marginal benefit, especially given the impact on outside users.

I really want to make sure that a move off Dockerhub is _required_ before asking teams to move their images.

geofffranks:
> Teams have recently been dealing with some toil around the accounts that we use to publish images to DockerHub and this has got us rethinking our usage of it a bit.

Can you elaborate on this toil? Changing all of our community docker image locations, references, and making sure people consuming those images outside of core CF contributors update and have access to the new images seems like an immense amount of work.

Our team doesn't experience much toil at all when publishing to dockerhub. Our main painpoints have been being rate limited in CI while pulling from dockerhub. As a solution, we implemented a GCP Artifact repository to act as a pull-through cache of dockerhub, and we haven't had any issues since.

toc/rfc/rfc-draft-use-github-packages-container-registry-instead-of-dockerhub.md
christopherclark: cloudfoundry is NOT covered by the DSOS program, but we pay for that org (about $500/month). Paketo buildpacks _is_ covered by the DSOS program. You have to re-apply annually, for each DH organization, and TBH the criteria for this program is opaque, and they aren't very communicative about how it works.

tcdowney:
@ameowlia @rkoster @ChrisMcGowan 

I think we should close this RFC. I think the suggestion for WGs to have their own WG-aligned DockerHub accounts is still a good idea, but it is not really something to have an RFC about. ARI is already doing it and I think other WGs should too as the need arises.

The only loose thread here is whether or not we can get some form of Google Groups set up for the CFF that would make this easier. cc @christopherclark

beyhan:
This was discuss during the TOC meeting on 21 of May. @christopherclark will try this out together with @rkoster for the FI WG and provide a path forward. So, as suggested we close this.

PR #821, opened 2024-04-24
add cnbapplifecycle repo

modulo11:
Following @beyhan's [comment](https://github.com/cloudfoundry/community/pull/796#issuecomment-2074171579), we would like to request a cloudfoundry/cnbapplifecycle repository for the lifecycle created within this [RFC](https://github.com/cloudfoundry/community/pull/796).

Is that the right spot?

CC @loewenstein @stephanme @ameowlia

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -866,6 +866,10 @@ orgs:
       cnb2cf:
         archived: true
         has_projects: true
+      cnbapplifecycle:
+        description: Lifecycle to create apps using Cloud Native Buildpacks
+        has_projects: false
+        has_wiki: false
       commandrunner:
         has_projects: true
       community:
--- a/toc/working-groups/app-runtime-platform.md
+++ b/toc/working-groups/app-runtime-platform.md
@@ -103,6 +103,7 @@ areas:
   - cloudfoundry/cfdot
   - cloudfoundry/cfhttp
   - cloudfoundry/clock
+  - cloudfoundry/cnbapplifecycle
   - cloudfoundry/consuladapter
   - cloudfoundry/debugserver
   - cloudfoundry/diego-acceptance


loewenstein (approved):
LGTM

toc/working-groups/foundational-infrastructure.md
```diff
@@ -325,6 +325,7 @@ areas:
   - cloudfoundry/bosh-workstation
   - cloudfoundry/bpm-release
   - cloudfoundry/bsdtar
+  - cloudfoundry/cnbapplifecycle
```
stephanme, line 328: I think the cnbapplifecycle repo should be added to the [App Runtime Platform WG](https://github.com/cloudfoundry/community/blob/main/toc/working-groups/app-runtime-platform.md), Diego area and not to Foundational Infrastructure.
At least, the other lifecycle implementation live in ARP/Diego.

beyhan: Sorry, if my comment on the RFC with the example wasn't fully clear but definitely the FI WG isn't the right place for it.

ameowlia (changes requested):
Hi @modulo11, Thanks for opening this PR.

In order to add a new repo, I think you will also have to open a PR to add it [here](https://github.com/cloudfoundry/community/blob/a863f61074ceeb244dbca70d5624a58ab7da6949/org/cloudfoundry.yml#L19-L22) first. That way the automation will create a blank repo. Then @christopherclark can move it over.
Are any approvers moving over with this repo? If, so can you add them in the same PR?
Also, since those approvers probably don't want to maintain other diego components, I think you should create a new area for this.

rkoster (approved):

ChrisMcGowan (approved):

loewenstein:
> * Are any approvers moving over with this repo? If, so can you add them in the same PR?
> * Also, since those approvers probably don't want to maintain other diego components, I think you should create a new area for this.

Should we start with the repo just being in the Diego area with no additional approvers. That way we'll have the repo for opening PRs.

I guess we'll anyway need to figure out whom to invite for review for some initial work. We can then later figure out how to treat the new lifecycle with regard to approvers. WDYT?

ameowlia (approved):

PR #832, opened 2024-04-30
branchprot: Exclude buildpack repos managed by CI

arjun024:
see e.g. failure https://buildpacks.ci.cf-app.com/teams/core-deps/pipelines/cf-release/jobs/create-ruby-buildpack-dev-release/builds/1161

PR #837, opened 2024-05-06
Fix branch protection config for buildpack repos

stephanme:
Removed org prefix from repo names. Org is already specified in the yaml hierarchy.


stephanme:
Branch protection config for buildpacks was introduced with #832

beyhan:
Merging this to fix the branch protection.

PR #872, opened 2024-05-24
Remove branchprotection from capi-env-pool

Gerg:
[capi-env-pool](https://github.com/cloudfoundry/capi-env-pool) is used for a concourse pool resource (https://github.com/concourse/pool-resource) for development environments and is not packaged with capi-release or cf-deployment, so branch protection is not needed.


Gerg:
cc @tcdowney @moleske @sethboyles

stephanme (approved):

PR #873, opened 2024-05-30
A temporaray team for uaa pipeline access at https://bosh.ci.cloudfou…

hsinn0:
…ndry.org/

- For new UAA team members
- As suggested by rkoster

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -2548,3 +2548,18 @@ orgs:
           terraform-provider-csbpg: admin
           terraform-provider-csbmysql: admin
           terraform-provider-csbsqlserver: admin
+      temp-uaa-ci-admins:
+        description: "Temporary team for managing UAA CI"
+        maintainers:
+          - coolgang123
+          - duanemay
+          - iprotsiuk
+        members: [ ]
+        privacy: closed
+        repos:
+          cf-uaac: admin
+          uaa: admin
+          uaa-ci: admin
+          uaa-cli: admin
+          uaa-release: admin
+          uaa-singular: admin
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -2557,9 +2557,4 @@ orgs:
         members: [ ]
         privacy: closed
         repos:
-          cf-uaac: admin
-          uaa: admin
           uaa-ci: admin
-          uaa-cli: admin
-          uaa-release: admin
-          uaa-singular: admin


org/cloudfoundry.yml
```diff
@@ -2548,3 +2548,18 @@ orgs:
...
+      temp-uaa-ci-admins:
+        description: "Temporary team for managing UAA CI"
+        maintainers:
+          - coolgang123
+          - duanemay
+          - iprotsiuk
+        members: [ ]
+        privacy: closed
+        repos:
+          cf-uaac: admin
```
rkoster, line 2560: Why are these team members also getting `admin` rights on these repo's? I'm assuming read would be enough? Or no repo's needed at all, since it would just be used to grant access in concourse.

hsinn0: Pushed another commit.

org/cloudfoundry.yml
```diff
@@ -2548,3 +2548,13 @@ orgs:
           terraform-provider-csbpg: admin
           terraform-provider-csbmysql: admin
           terraform-provider-csbsqlserver: admin
+      temp-uaa-ci-admins:
+        description: "Temporary team for managing UAA CI"
+        maintainers:
+          - coolgang123
+          - duanemay
+          - iprotsiuk
```
beyhan, line 2556: During the TOC meeting we noticed that some of the users getting admin access through this pr don't have any role in the UAA area. We would suggest to give admin access to approvers from the UAA area.

hsinn0: > During the TOC meeting we noticed that some of the users getting admin access through this pr don't have any role in the UAA area. We would suggest to give admin access to approvers from the UAA area.

I do not understand. Do you mean that I should bring back the repo list with admin rights in the 1st commit, which I removed in the 2nd commit in this PR?

beyhan: Hi @hsinn0,

I'm sorry if my comment is not understandable. The people [listed in the commit ](https://github.com/cloudfoundry/community/pull/873/files#diff-81cd47b9171f8de6dd9563548aa8dd398148ce2d4c63de03942db47a355eacc4R2554-R2556) do not have the [approver role](https://github.com/cloudfoundry/community/blob/main/toc/ROLES.md#approver) for the UAA area. For us `maintainers` is equal to approvers and that is why our suggestion is to add only [UAA approvers](https://github.com/cloudfoundry/community/blob/main/toc/working-groups/foundational-infrastructure.md?plain=1#L132-L152) to the maintainers list.

hsinn0: @beyhan , the purpose of this PR was to give rw access to UAA CI pipelines at https://bosh.ci.cloudfoundry.org/ to our new team members that do not have approver role to UAA repos yet (because they do not meet the requirement yet). Current UAA repo approvers already has the access to the UAA CI pipelines, and we would not need this PR if these people are approvers.
This PR was [suggested](https://cloudfoundry.slack.com/archives/C0267SM551U/p1717079329206169?thread_ts=1699563132.761139&cid=C0267SM551U) by @rkoster in #uaa-maintainers-private Slack channel.

rkoster (approved):

strehle (approved):

org/cloudfoundry.yml
beyhan: Ok, I understand now. This is required only for CI access and should be fine. Also, doesn't need to go over the TOC, that is why I will remove the TOC reviewers and label.

PR #890, opened 2024-06-12
Use a virtual environment for pip in org automation readme

beyhan:
Pip will refuse to install system-wide packages and suggests the use of a virtual environment. More details in https://peps.python.org/pep-0668/

### DIFF
--- a/org/readme.md
+++ b/org/readme.md
@@ -85,6 +85,8 @@ Requires Python 3.9.
 How to run locally:
 ```
 cd ./org
+python -m venv <path/to/venv>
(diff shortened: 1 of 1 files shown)


stephanme (approved):
There are many options for setting up python and venvs (e.g. I use pyenv which takes care of the python version and installation).

But venv is the most basic one and ships with python out of the box.

ChrisMcGowan (approved):

rkoster (approved):

ameowlia (approved):

PR #896, opened 2024-06-13
Fix remove not existing uaa-ci repo listin

beyhan:
`repos` expect a map and not an array. Let's try this fix. I'm not sure whether this will work with the `peribolos` tooling but I think so based on code here https://github.com/kubernetes/test-infra/blob/0c1e01159b26a39d2a37de4d83cb18ab94747ed8/prow/cmd/peribolos/main.go#L1146-L1164

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -2584,4 +2584,4 @@ orgs:
           - iprotsiuk
         members: [ ]
         privacy: closed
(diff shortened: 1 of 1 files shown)


ChrisMcGowan (approved):

beyhan:
This is required for the automation. That is why I will just merge with one approval.

PR #929, opened 2024-07-18
Update rfc-0030-add-support-for-file-based-service-binding.md

PlamenDoychev:
Updating the RFC document as it contradicts with the servicebinding.io spec. Most probably leftover from the discussion phase.


stephanme (approved):
Plamen and team implemented a first POC. This is the feedback from a review/discussion with the CAPI team.

beyhan (approved):

PR #940, opened 2024-08-06
Move volume services to App-runtime-platform

pivotal-marcela-campo:
Approvers for volume services are now all members of app-runtime-platform.  Whilst the separation between platform (volman) and services (volume services in this case) makes sense theoretically, in practice means that the same engineers need to be part of two working groups and liase with different leaders in the CF org, for a project that is not particularly active in the community at this point.


ameowlia (approved):

ChrisMcGowan (approved):

ameowlia:
⚠️ This PR removes the following approvers from the area: @dlresende, @totherme, @blgm, @ifindlay-cci, @nouseforaname, @syslxg, @jhvhs.

If you would like to remain an approver, please comment here and I will add you back. 

If you are okay with having your approver status revoked, please emoji react on this comment with a 👍 .

You have two weeks to refute this removal. If you do not comment by Aug 20th your approver status will be revoked.

rkoster (approved):

PR #946, opened 2024-08-13
[RFC] User Creation by Org Managers

stephanme:
Allow Org Managers to create users in UAA in order to improve the onboarding procedure for larger developer groups into multi-tenant Cloud Foundry foundations.

[Preview](https://github.com/cloudfoundry/community/blob/rfc-user-creation-by-org-managers/toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md)

### DIFF
--- a/toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
+++ b/toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
@@ -9,7 +9,7 @@
 
 ## Summary
 
-Allow Org Managers to create users in UAA in order to improve the onboarding procedure for larger developer groups into multi-tenant Cloud Foundry foundations.
+Allow Org Managers (and CF Admins) to create users in UAA in order to improve the onboarding procedure for larger developer groups into multi-tenant Cloud Foundry foundations. The users are created by the Cloud Controller on behalf of the Org Manager.
 
 ## Problem
 
@@ -21,15 +21,15 @@ This is an issue for customers that want to onboard larger developer groups into
 
 The proposal is to grant Org Managers the right to assign CF roles to users which are not yet known in UAA. The users are created in UAA either explicitly via an enhanced `POST /v3/users` call or implicitly when an Org Manager assigns an organization role using `POST /v3/roles`.
 
-The functionality SHALL be gated by a feature flag and SHALL be disabled by default.
+The functionality SHALL be gated by a deployment manifest configuration flag and SHALL be disabled by default.
 
 ## Required Changes by Component
 
 ### Cloud Controller
 
-#### Feature Flags
+#### CAPI Release
 
-Introduce a new [feature flag](https://v3-apidocs.cloudfoundry.org/version/3.172.0/index.html#list-of-feature-flags) `user_creation_by_org_manager` which is disabled by default .
+Introduce a configuration flag `cc.allow_user_creation_by_org_manager` which is disabled by default .
 
 #### POST /v3/users
 
@@ -40,23 +40,23 @@ Required Parameters for Create User by Name
 |Name|Type|Description|
 |---|---|---|
 |username|string|The username as (to be) registered in UAA.|
-|origin|string|The origin / identity provider for the user.|
+|origin|string|The origin / identity provider for the user. `origin=uaa` is not allowed as it requires a password. |
 
 
 Permitted roles for Create User by Name
 - Admin
-- OrgManager (if feature flag `user_creation_by_org_manager` is enabled)
+- OrgManager (if configuration flag `cc.allow_user_creation_by_org_manager` is enabled)
 
-CAPI will call the [POST /Users](https://docs.cloudfoundry.org/api/uaa/version/77.14.0/index.html#create-2) endpoint of the UAA using a CAPI owned UAA client for creating the user if it doesn't exist yet in UAA. The user is then registered in the Cloud Controller database under the user guid returned by UAA.
+CAPI will call the [POST /Users](https://docs.cloudfoundry.org/api/uaa/version/77.14.0/index.html#create-2) endpoint of the UAA using a CAPI owned UAA client with scope `scim.write` or `uaa.admin` for creating the user if it doesn't exist yet in UAA. The user is then registered in the Cloud Controller database under the user guid returned by UAA.
 
 #### POST /v3/roles
 
-If a role is created by user name and origin (requires feature flag `set_roles_by_username` to be enabled) and the feature flag `user_creation_by_org_manager` is enabled, the user with the provided name and origin SHALL be created in CF UAA if the users doesn't exist yet in UAA.
+If a role is created by user name and origin (requires feature flag `set_roles_by_username` to be enabled) and the configuration flag `cc.allow_user_creation_by_org_manager` is enabled, the user with the provided name and origin SHALL be created in CF UAA if the users doesn't exist yet in UAA.
 In other words: Role assignment shall succeed even if the user is still unknown to CF UAA. The user is created implicitly similar to `POST /v3/users` by username/origin described above.
 
 ### CLI
 
-`cf create-user USERNAME [--origin ORIGIN]` (user creation without specifying a password) SHOULD use the CF API `POST /v3/users` instead of calling CF UAA directly so that Org Managers can create users if feature flag `user_creation_by_org_manager` is enabled.
+`cf create-user USERNAME [--origin ORIGIN]` (user creation without specifying a password) SHOULD use the CF API `POST /v3/users` instead of calling CF UAA directly so that Org Managers can create users if configuration flag `cc.allow_user_creation_by_org_manager` is enabled.
 
 ### UAA
 
--- a/toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
+++ b/toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
@@ -1,10 +1,10 @@
 # Meta
 [meta]: #meta
 - Name: User Creation by Org Managers
-- Start Date: (fill in today's date: YYYY-MM-DD)
+- Start Date: 2024-08-13
 - Author(s): stephanme
 - Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
-- RFC Pull Request: (fill in with PR link after you submit it)
+- RFC Pull Request: [#946](https://github.com/cloudfoundry/community/pull/946)
 
 
 ## Summary
--- a/toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
+++ b/toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
@@ -29,7 +29,9 @@ The functionality SHALL be gated by a deployment manifest configuration flag and
 
 #### CAPI Release
 
-Introduce a configuration flag `cc.allow_user_creation_by_org_manager` which is disabled by default .
+Introduce a configuration flag `cc.allow_user_creation_by_org_manager` which is disabled by default.
+
+Add configuration for an UAA client that is used for creating the users in UAA on behalf of CAPI, e.g. `uaa.clients.cloud_controller_shadow_user_creation`. The UAA client is required when `cc.allow_user_creation_by_org_manager` is enabled, otherwise optional.
 
 #### POST /v3/users
 
@@ -57,6 +59,10 @@ In other words: Role assignment shall succeed even if the user is still unknown 
 ### UAA
 
 No changes required.
+
+### cf-deployment
+
+Add an [ops file](https://github.com/cloudfoundry/cf-deployment/blob/main/README.md#ops-files) to enable user creation as Org Manager, i.e. the ops file enables `cc.allow_user_creation_by_org_manager` and provisions the UAA client `cloud_controller_shadow_user_creation` with the required scopes.
 
 ## Possible Future Work
 
--- a/toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
+++ b/toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
@@ -1,0 +1,63 @@
+# Meta
+[meta]: #meta
+- Name: User Creation by Org Managers
+- Start Date: (fill in today's date: YYYY-MM-DD)
+- Author(s): stephanme
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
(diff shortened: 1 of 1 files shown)


stephanme:
FYI: @cloudfoundry/toc , @cloudfoundry/wg-app-runtime-interfaces-capi-approvers

Gerg:
In this workflow, who or what is granting the user the Org Manager role? Could the Org Manager also be granted an appropriate scope to create users in UAA, instead of CC creating the user on their behalf?

Gerg:
cc @cloudfoundry/wg-foundational-infrastructure-identity-and-auth-uaa-approvers

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
```diff
@@ -0,0 +1,63 @@
...
+- Name: User Creation by Org Managers
+- Start Date: 2024-08-13
+- Author(s): stephanme
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: [#946](https://github.com/cloudfoundry/community/pull/946)
+
+
+## Summary
+
+Allow Org Managers to create users in UAA in order to improve the onboarding procedure for larger developer groups into multi-tenant Cloud Foundry foundations.
```
Gerg, line 12: Two things it might be nice to include in the summary (assuming they're true):
1. CC is creating the user on behalf of the Org Manager
2. CC admin users can also use this

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
```diff
@@ -0,0 +1,63 @@
...
+
+The functionality SHALL be gated by a feature flag and SHALL be disabled by default.
+
+## Required Changes by Component
+
+### Cloud Controller
+
+#### Feature Flags
+
+Introduce a new [feature flag](https://v3-apidocs.cloudfoundry.org/version/3.172.0/index.html#list-of-feature-flags) `user_creation_by_org_manager` which is disabled by default .
```
Gerg, line 32: Should this be an API feature flag, or a configuration property in the deployment manifest? Would this be something that CC admins would want to toggle at run time?

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
```diff
@@ -0,0 +1,63 @@
...
+
+#### POST /v3/users
+
+Enhance the [POST /v3/users](https://v3-apidocs.cloudfoundry.org/version/3.172.0/index.html#create-a-user) endpoint by the parameters `username` and `origin`. `guid` and `username` parameters are exclusive.
```
Gerg, line 36: These are the current UAA User fields that CC shadows. There are additional fields that can be set directly via the UAA API: https://docs.cloudfoundry.org/api/uaa/version/77.14.0/index.html#create-2

Gerg (commented):
Overall, I'm a bit wary of adding a POST request to UAA nested inside of a synchronous CC request. We'd have to contend with issues of timeouts, error handling, etc.

Ideally, we could get the Org Manager users sufficient permissions to create the user in UAA themselves, maybe with the facilitation of the CLI?

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
```diff
@@ -0,0 +1,63 @@
...
+|---|---|---|
+|username|string|The username as (to be) registered in UAA.|
+|origin|string|The origin / identity provider for the user.|
+
+
+Permitted roles for Create User by Name
+- Admin
+- OrgManager (if feature flag `user_creation_by_org_manager` is enabled)
+
+CAPI will call the [POST /Users](https://docs.cloudfoundry.org/api/uaa/version/77.14.0/index.html#create-2) endpoint of the UAA using a CAPI owned UAA client for creating the user if it doesn't exist yet in UAA. The user is then registered in the Cloud Controller database under the user guid returned by UAA.
```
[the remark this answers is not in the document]
stephanme: Scopes added: `scim.write` or `uaa.admin` is required (according to UAA API doc).

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
stephanme: Done

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
stephanme: I chose a feature flag at first because of the related feature flag `set_roles_by_username`.

But indeed, a deploy time config property should be sufficient. I can't imagine of an use case where CC admins have to change this at run time.

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
stephanme: Yes. For origin!=uaa, the additional fields are usually set by the identity provider when the user authenticates.

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
```diff
@@ -0,0 +1,63 @@
...
+#### POST /v3/users
+
+Enhance the [POST /v3/users](https://v3-apidocs.cloudfoundry.org/version/3.172.0/index.html#create-a-user) endpoint by the parameters `username` and `origin`. `guid` and `username` parameters are exclusive.
+The existing user creation by `guid` SHALL NOT be changed.
+
+Required Parameters for Create User by Name
+|Name|Type|Description|
+|---|---|---|
+|username|string|The username as (to be) registered in UAA.|
+|origin|string|The origin / identity provider for the user.|
```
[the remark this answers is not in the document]
stephanme: Both params are required for the 'Create User by Name' scenario.
Added that origin=uaa is not allowed.

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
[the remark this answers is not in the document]
stephanme: Yes. Eventually, those fields are set/updated by the identity provider when the user authenticates.

stephanme:
> In this workflow, who or what is granting the user the Org Manager role? Could the Org Manager also be granted an appropriate scope to create users in UAA, instead of CC creating users on their behalf?

The initial Org Manager role is assigned by an external onboarding process that uses a technical CF admin user. From then on, the initial Org Manager can grant Org Manager role to additional users using CF CLI or any other CF API client (e.g. terraform).
Assigning a special UAA scope could work for the initial Org Manager (the external onboarding process could assign it) but not anymore for the additional ones - unless it is implemented in the `POST /v3/roles` endpoint which is even more complex compared to the proposed solution.

> Overall, I'm a bit wary of adding a POST request to UAA nested inside of a synchronous CC request. We'd have to contend with issues of timeouts, error handling, audit-ability, etc.

That's a valid point. But on the other hand, `POST /v3/roles` calls already the UAA to get the user guid and as mitigation, the UAA POSTs are behind a config flag so that foundations that don't need this feature are not impacted.

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
```diff
@@ -0,0 +1,63 @@
...
+CAPI will call the [POST /Users](https://docs.cloudfoundry.org/api/uaa/version/77.14.0/index.html#create-2) endpoint of the UAA using a CAPI owned UAA client for creating the user if it doesn't exist yet in UAA. The user is then registered in the Cloud Controller database under the user guid returned by UAA.
+
+#### POST /v3/roles
+
+If a role is created by user name and origin (requires feature flag `set_roles_by_username` to be enabled) and the feature flag `user_creation_by_org_manager` is enabled, the user with the provided name and origin SHALL be created in CF UAA if the users doesn't exist yet in UAA.
+In other words: Role assignment shall succeed even if the user is still unknown to CF UAA. The user is created implicitly similar to `POST /v3/users` by username/origin described above.
+
+### CLI
+
+`cf create-user USERNAME [--origin ORIGIN]` (user creation without specifying a password) SHOULD use the CF API `POST /v3/users` instead of calling CF UAA directly so that Org Managers can create users if feature flag `user_creation_by_org_manager` is enabled.
```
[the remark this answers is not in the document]
stephanme: Let's hear the feedback by @wg-app-runtime-interfaces-cli-approvers

beyhan:
We discussed this during the TOC meeting on 01.10.2024 and decided to start the FCP with the goal to accept it during the next TOC meeting.

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
```diff
@@ -0,0 +1,63 @@
...
+CAPI will call the [POST /Users](https://docs.cloudfoundry.org/api/uaa/version/77.14.0/index.html#create-2) endpoint of the UAA using a CAPI owned UAA client with scope `scim.write` or `uaa.admin` for creating the user if it doesn't exist yet in UAA. The user is then registered in the Cloud Controller database under the user guid returned by UAA.
+
+#### POST /v3/roles
+
+If a role is created by user name and origin (requires feature flag `set_roles_by_username` to be enabled) and the configuration flag `cc.allow_user_creation_by_org_manager` is enabled, the user with the provided name and origin SHALL be created in CF UAA if the users doesn't exist yet in UAA.
+In other words: Role assignment shall succeed even if the user is still unknown to CF UAA. The user is created implicitly similar to `POST /v3/users` by username/origin described above.
+
+### CLI
+
+`cf create-user USERNAME [--origin ORIGIN]` (user creation without specifying a password) SHOULD use the CF API `POST /v3/users` instead of calling CF UAA directly so that Org Managers can create users if configuration flag `cc.allow_user_creation_by_org_manager` is enabled.
```
Gerg, line 59: Won't this be a breaking change for CLI users who:
1. Do have permissions to create users in UAA
2. Do NOT have the `cc.allow_user_creation_by_org_manager` enabled

In that case neither the CLI nor CC will attempt to create the user in UAA.

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
```diff
@@ -0,0 +1,63 @@
...
+|---|---|---|
+|username|string|The username as (to be) registered in UAA.|
+|origin|string|The origin / identity provider for the user. `origin=uaa` is not allowed as it requires a password. |
+
+
+Permitted roles for Create User by Name
+- Admin
+- OrgManager (if configuration flag `cc.allow_user_creation_by_org_manager` is enabled)
+
+CAPI will call the [POST /Users](https://docs.cloudfoundry.org/api/uaa/version/77.14.0/index.html#create-2) endpoint of the UAA using a CAPI owned UAA client with scope `scim.write` or `uaa.admin` for creating the user if it doesn't exist yet in UAA. The user is then registered in the Cloud Controller database under the user guid returned by UAA.
```
Gerg, line 50: To confirm, 
1. This will be a new client, not adding scopes to the [existing usernames lookup client](https://github.com/cloudfoundry/cf-deployment/blob/2bb7d43f751714c8049b8d612a5c9e6836b9f345/cf-deployment.yml#L689)
2. This client will NOT be included by default in cf-deployment

Are those correct?

Gerg (changes requested):
I'm still nervous about nesting a POST inside a CC request, but I understand the use case it is trying to solve. I'm fine with this as long as it is off by default in cf-d.

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
stephanme: That's correct. We would add a new client e.g. cloud_controller_shadow_user_creation (shadow user because the use case works only for external IDPs).

A new ops file in cf-deployment could set the configuration flag and configure the new UAA client so that the client doesn't exist by default. CC would respond with a 5xx if `cc.allow_user_creation_by_org_manager` is enabled but the UAA client is missing.

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
stephanme: Good point. I shifted the CLI enhancements into a 'possible future work' section with two options that address the case "user has scim.write/uaa.admin AND `cc.allow_user_creation_by_org_manager` disabled" (more or less elegantly).
There were concerns before regarding the CLI changes so taking them out of scope for now seems the right step.

@beyhan, @torsten-sap: Any concerns from your side? Developer onboarding doesn't need `cf create-user` but `cf set-org/space-role` is sufficient which will implicitly create shadow users if flag `cc.allow_user_creation_by_org_manager` is enabled.

torsten-sap: @stephanme From my point of view it is fine to start with the implicit shadow user creation during the role assignment. Nevertheless, I believe that from product perspective the "create-user" command should also follow. Otherwise, it seems like a gap.

@lechnerc77 Could you please also add your point of view? Thanks!

beyhan: Good catch regarding the breaking change. I think it is fine to focus in this RFC on the API side and leave the CF clients (only CF CLI in this case) as possible future improvement. The nice side-effect here is that CF CLI `cf set-org/space-role` will just work as mentioned.

lechnerc77: @torsten-sap I fully agree with your statement: starting with an implicit shadow user creation is okay (and customers will already be quite happy about it), as long a the full functionality will be shipped later.

toc/rfc/app-runtime-interfaces/rfc-draft-user-creation-by-org-managers.md
stephanme: I mentioned the extra UAA client in the RFC and added a section for a cf-deployment ops file that enables the feature and provisions the additional UAA client.

philippthun (approved):

Gerg (approved):

ameowlia (approved):

ChrisMcGowan (approved):

rkoster (approved):

beyhan (approved):

PR #950, opened 2024-08-19
Fix default branch for bosh-linux-stemcell-builder

stephanme:
Was renamed from ubuntu-jammy/master into ubuntu-jammy bypassing org automation.


stephanme:
Wrong branch.

PR #970, opened 2024-09-05
Add new repo `uaa-ci`

hsinn0:
- To migrate from `pivotal` org.
- Giving approver roles only to the current team members with access to secrets that are required to update the pipelines.

### DIFF
--- a/org/cloudfoundry.yml
+++ b/org/cloudfoundry.yml
@@ -2368,6 +2368,10 @@ orgs:
         description: CloudFoundry User Account and Authentication (UAA) Server
         has_projects: true
         has_wiki: false
+      uaa-ci:
+        default_branch: main
+        has_projects: true
+        has_wiki: false
       uaa-cli:
         description: CLI for UAA written in Go
         has_projects: true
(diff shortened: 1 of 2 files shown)


hsinn0:
Following the instructions given in the [Slack thread]( https://cloudfoundry.slack.com/archives/C0267SM551U/p1725564442199159?thread_ts=1725558475.984609&cid=C0267SM551U).

jpalermo (approved):

beyhan (approved):

ameowlia:
> Giving approver roles only to the current team members with access to secrets that are required to update the pipelines.

Approver is not enough to allow acces to github secrets. Github secrets can only be accessed by admin users. If people need admin access, you can create an admin override team [like this](https://github.com/cloudfoundry/community/blob/71fa9c498ac308eb1dc73db19b775451b52d1f04/org/cloudfoundry.yml#L2532-L2547).

hsinn0:
> @hsinn0 - Approver level is not enough to access to github secrets. Github secrets can only be accessed by admin users.

@ameowlia, the secrets here are not Github secrets. They are ones stored inside an internal vault.

rkoster (approved):

ameowlia (approved):

PR #977, opened 2024-09-17
Add Stratos to App Runtime Interfaces Working Group

krutten:


rkoster (approved):

ChrisMcGowan (approved):

beyhan (approved):

toc/working-groups/app-runtime-interfaces.md
```diff
@@ -510,6 +510,28 @@ areas:
   - cloudfoundry/notifications
   - cloudfoundry/app-runtime-interfaces-infrastructure
 
+- name: Stratos Console for Cloud Foundry
+   approvers:
```
stephanme, line 514: PR validation failed. Looks to me as if yaml indentation is not correct (approvers etc should be on same level as name)

stephanme (changes requested):
FYI: There is an RFC  #941 to define an EOL for CF API v2 and to remove it from CAPI eventually. This is still far ahead in the future but you should be aware of this RFC and plan the migration of any CF API v2 usage in Stratos to v3.

toc/working-groups/app-runtime-interfaces.md
krutten: Fixed indentation and tested with spruce.

Gerg (approved):

ameowlia (approved):

PR #991, opened 2024-10-02
Add new approver for Prometheus

benjaminguttmann-avtq:
Adding an additional approver from anynines


benjaminguttmann-avtq:
@frodenas @gmllt @mdimiceli can you please show that you are okay with that

beyhan:
We have thumps-up from two approvers. That is why I'm going to merge this now.

beyhan (approved):

PR #994, opened 2024-10-10
Bindable Secrets Resource

tcdowney:
This RFC proposes adding a `secret` resource to the V3 Cloud Foundry API, similar in design and purpose to the Kubernetes [Secret resource](https://kubernetes.io/docs/concepts/configuration/secret/). Cloud Foundry Secrets will store arbitrary data in [CredHub](https://docs.cloudfoundry.org/credhub/) inject it into app containers via `tmpfs`-mounted files at a given path, using the same mechanism as [RFC 0030 - Add Support for File based Service Binding Information](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0030-add-support-for-file-based-service-binding.md).

---

**[Readable Version](https://github.com/tcdowney/community/blob/first-class-secrets-in-cf-rfc/toc/rfc/rfc-draft-bindable-secrets.md)**

### DIFF
--- a/toc/rfc/rfc-draft-bindable-secrets.md
+++ b/toc/rfc/rfc-draft-bindable-secrets.md
@@ -8,31 +8,33 @@
 
 ## Summary
 
-**tl;dr** Kubernetes-style Secrets in Cloud Foundry
+**tl;dr** Kubernetes-style Secrets for Cloud Foundry
 
-This RFC proposes a new resource to the V3 Cloud Foundry APIs called the `Secret` that is similar in design and purpose to the Kubernetes [Secret resource](https://kubernetes.io/docs/concepts/configuration/secret/). Cloud Foundry Secrets will support arbitrary data that is stored in [CredHub](https://docs.cloudfoundry.org/credhub/) and made available to apps via `tmpfs` mounted files using the same mechanism as [RFC 0030 - Add Support for File based Service Binding Information](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0030-add-support-for-file-based-service-binding.md).
+This RFC proposes adding a `secret` resource to the V3 Cloud Foundry API, similar in design and purpose to the Kubernetes [Secret resource](https://kubernetes.io/docs/concepts/configuration/secret/). Cloud Foundry Secrets will store arbitrary data in [CredHub](https://docs.cloudfoundry.org/credhub/) inject it into app containers via `tmpfs`-mounted files at a given path, using the same mechanism as [RFC 0030 - Add Support for File based Service Binding Information](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0030-add-support-for-file-based-service-binding.md).
 
 ## Problem
 
-The Cloud Foundry platform currently prefers for applications to accept configuration through environment variables. Other application platforms, such as Kubernetes, support other mechanisms for providing configuration at runtime [as volume-mounted files](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod). Many modern applications are adopting this file-based configuration approach since these files are not subject to the size limits of environment variables and can be updated without having to recreate the container. This RFC proposes introducing a similar mechanism in order to support these sorts of applications on Cloud Foundry -- without requiring code changes.
+Cloud Foundry currently supports app configuration via environment variables. Other application platforms, such as Kubernetes, support app configuration at runtime [via volume-mounted files](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod). Many apps are adopting this file-based configuration approach, since environment variables have length limits and files can be updated without having to recreate the container. This RFC proposes introducing file-based configuration to support these sorts of apps on Cloud Foundry — without requiring app code changes.
 
-### Why can't apps just use user-provided service instances (UPSIs) and service binding files?
-Some apps may have config they want as files provided at runtime that don't naturally fit into the service binding model. For example, if an app relies on a config file that needs to live at a certain path, it may not be able to be updated to find it at another path such as `$SERVICE_BINDING_ROOT`. This is especially the case for commercial off-the-shelf software (COTS).
+### What About User-Provided Services?
+
+Why can't developers use user-provided service instances (UPSIs) and service bindings for file-based app configuration? Some apps may have file-based config that don't naturally fit into the service binding model. For example, if an app relies on a config file at a certain path, that plath may not be configurable to `$SERVICE_BINDING_ROOT`. This is especially the case for commercial off-the-shelf software (COTS).
 
 ## Proposal
 
-We introduce two new resources to the V3 Cloud Foundry APIs: the `Secret` and the `Secret Binding` along with their associated CRUD API endpoints.
+Add two new resources to the V3 Cloud Foundry API: `secret` and the `secret_binding`, along with associated CRUD API endpoints.
 
-### Why a separate `Secret` resource? Why not just store them under apps like environment variables?
-Several reasons here.
+### Why Top-Level Resources?
+
+Why not represent secrets as sub-resources of apps like environment variables? Two reasons:
 
 1. Secrets may be larger than environment variables. Modeling them separately lets us naturally keep them in a separate database table and even use separate services, like CredHub, to store their contents.
-2. `Secrets` may be shared across multiple apps. For example, if an "app" may actually be several actual Cloud Foundry apps that need to share certain configuration, certificates, or other credentials. By scoping the `Secret` at the `Space` level we can easily support this. Or an app may simply be using the blue-green deployment pattern and be several CF apps for that reason.
+2. Secrets may be shared across multiple apps. For example, a "logical app" may actually be several Cloud Foundry apps resources that need to share certain configuration, certificates, or other credentials. By scoping the `secret` at the `space` level, we can easily support this. Similarly, an app using the blue-green deployment pattern can easily share configuration between versions.
 
-### `Secret` and `Secret Binding` Resources
-The `Secret` will be a `Space`-scoped resource (similar to a `Route` or a `Service Instance`) that can be bound to one or more `Apps` using a `Secret Binding`.
+### `secret` and `secret_binding` Resources
+The `secret` will be a `space`-scoped resource (similar to a `route` or a `service_instance`) that can be bound to one or more `app`s in the `space` using a `secret_binding`.
 
-#### Example `Secret` Object
+#### Example `secret` Object
 
 ```json
 {
@@ -70,9 +72,9 @@ The `Secret` will be a `Space`-scoped resource (similar to a `Route` or a `Servi
 }
 ```
 
-The `type` parameter corresponds with a [CredHub credential type](https://docs.cloudfoundry.org/credhub/credential-types.html). To start we will focus on supporting the `value` and `json` credential types, but can expand to others like the `certificate` type as these may pave the way to future enhancements and use cases.
+The `type` field corresponds with a [CredHub credential type](https://docs.cloudfoundry.org/credhub/credential-types.html). To start, `secret`s will support the `value` and `json` credential types. In the future, this can be expanded to other CredHub-supported types like `certificate`, to support additional use cases.
 
-#### Example `Secret Binding` Object
+#### Example `secret_binding` Object
 
 ```json
 {
@@ -111,9 +113,9 @@ The `type` parameter corresponds with a [CredHub credential type](https://docs.c
 }
 ```
 
-The `mount_path` parameter defines the `tmpfs` directory that will contain files for the `Secret`'s content. If a credential has subkeys (such as a [`json` type credential](https://docs.cloudfoundry.org/api/credhub/version/main/#_find_a_credential_by_id_type_json) or [`certificate` type](https://docs.cloudfoundry.org/api/credhub/version/main/#_find_a_credential_by_id_type_certificate)) then files will be made for each key.
+The `mount_path` field defines the `tmpfs` directory that will contain files for the `secret`'s content. If a credential has subkeys (such as a [`json` type credential](https://docs.cloudfoundry.org/api/credhub/version/main/#_find_a_credential_by_id_type_json) or [`certificate` type](https://docs.cloudfoundry.org/api/credhub/version/main/#_find_a_credential_by_id_type_certificate)), then files will be made for each key.
 
-So for a `json` type credential whose value is:
+For example, consider a `json`-type `secret` with the following `value`:
 
 ```json
 "value": {
@@ -122,19 +124,19 @@ So for a `json` type credential whose value is:
 }
 ```
 
-We would have a file called `application.yml` created under `/etc/my-conf`.
+The above `secret` would mount files named `application.yml` and `secret.txt` at the path configured in the binding.
 
-For a `value` type credential the file would simply take the same name as the `Secret Binding`.
+For a `value`-type `secret`, the file name be the `secret_binding`'s name.
 
-### Creating a `Secret`
+### Creating a `secret`
 
-To create a `Secret` we will introduce a new CLI command: `cf create-secret`.
+To create a `secret`, users will use a new CLI command: `cf create-secret`.
 
 ```console
 cf create-secret SECRET_NAME SECRET_TYPE -p <CREDENTIALS_INLINE_OR_FILE>
 ```
 
-If `SECRET_TYPE` is `value` then contents of `-p` will not be parsed. If `SECRET_TYPE` is `json`, `certificate`, or some other CredHub credential type then the contents of `-p` MUST be valid JSON and may be parsed.
+If `SECRET_TYPE` is `value` then contents of `-p` will not be parsed. If `SECRET_TYPE` is `json`, `certificate`, or some other CredHub credential type, then the contents of `-p` MUST be valid JSON and may be parsed.
 
 This command will ultimately create a `POST` request to `/v3/secrets` that looks like this:
 
@@ -164,8 +166,8 @@ curl "https://api.example.org/v3/secrets" \
   }'
 ```
 
-### Reading a `Secret` and its value
-Users that have permission to view Environment Variables for an App will be able to view `Secret` values within a `Space`. We will introduce a separate endpoint for retrieving these (similar to how app environment variables and service instance credentials are retrieved): `GET /v3/secrets/:guid/value`
+### Reading a `secret` and Its Value
+Users that have permission to view environment variables for an app will be able to view `secret` values within a `space`. We will introduce a separate endpoint for retrieving these (similar to how app environment variables and service instance credentials are retrieved): `GET /v3/secrets/:guid/value`.
 
 ```console
 curl "https://api.example.org/v3/secrets/[guid]/value" \
@@ -177,23 +179,31 @@ curl "https://api.example.org/v3/secrets/[guid]/value" \
 }
 ```
 
-To support this Cloud Controller will need to have both read and write permissions for the `Secret`'s corresponding credential in CredHub.
+To support this, Cloud Controller will need to have both read and write permissions for the `secret`'s corresponding credential in CredHub.
 
-### Creating a `Secret Binding`
-We will introduce a new CLI command to allow users to create `Secret Bindings`: `cf bind-secret`. This will create a `POST` request to `/v3/secret_bindings`.
+### Creating a `secret_binding`
+To create `secret_bindings`, users will use a new CLI command: `cf bind-secret`. This will create a `POST` request to `/v3/secret_bindings`.
 
 ```console
 cf bind-secret APP_NAME SECRET_NAME BINDING_NAME
 ```
 
-### CLI changes
-Work will need to be done to implement `cf create-secret`, `cf bind-secret`, and the associated update/delete commands in the CLI.
+### CLI Changes
+Work will need to be done to implement the following commands:
+- `cf create-secret`
+- `cf bind-secret`
+- `cf secrets`
+- `cf secret`
+- `cf delete-secret`
+- `cf unbind-secret`
 
-### Cloud Controller changes
-New database tables will need to be created to store `Secrets` and `Secret Bindings` and CRUD APIs will need to be created for these resources. We will also need an API to support viewing a `Secret`'s value.
+In addition, modifications may be made to existing CLI commands to surface `secrets` and `secret_bindings`. For example, adding bound `secret`s to `cf app` output.
+
+### Cloud Controller Changes
+New database tables will need to be created to store `secret`s and `secret_binding`s. CRUD APIs will need to be created for these resources. We will also need an API to support viewing a `secret`'s value.
 
 ### BBS changes
-BBS was recently updated to support `ServiceBindingFiles` as part of RFC 0030. A `Secret`'s value may result in one or more files being created in the `tmpfs` mounted directory defined by the `MountPath` parameter.
+BBS was recently updated to support `ServiceBindingFiles` as part of RFC 0030. A `secret`'s value may result in one or more files being created in the `tmpfs` mounted directory defined by the `MountPath` parameter.
 
 ```
 action := &models.RunAction{
@@ -228,18 +238,18 @@ action := &models.RunAction{
 }
 ```
 
-### Launcher changes?
+### Launcher Changes?
 The launcher (or something else) will need to be updated to retrieve credentials from CredHub for `Secret Bindings`. It currently already does this for CredHub references contained within `VCAP_SERVICES` using the app container's instance identity credentials. Is there something other than the launcher that should have this responsibility?
 
 ### Other Considerations
 
-#### Not using CredHub
-Although Cloud Foundry comes with CredHub by default, it is not a required component. We could support `value` and `json` type `Secrets` directly in Cloud Controller as a fallback.
+#### Not Using CredHub
+Although Cloud Foundry comes with CredHub by default, it is not a required component. We could support `value` and `json` type `secret`s directly in Cloud Controller as a fallback.
 
-#### CredHub Maximum Secret size
-Kubernetes Secrets and ConfigMaps have a [maximum size of 1 MiB](https://kubernetes.io/docs/concepts/configuration/configmap/#motivation). The [maximum size of a CredHub credential is 64Kb](https://docs.cloudfoundry.org/credhub/credential-types.html). If we plan on using CredHub we will need to limit our Secrets to 64KB or get CredHub to increase the limit.
+#### CredHub Maximum Secret Size
+Kubernetes Secrets and ConfigMaps have a [maximum size of 1 MiB](https://kubernetes.io/docs/concepts/configuration/configmap/#motivation). The [maximum size of a CredHub credential is 64Kb](https://docs.cloudfoundry.org/credhub/credential-types.html). If we plan on using CredHub, we will need to limit our Secrets to 64KB or increase the CredHub limit.
 
 #### (Future) Rotatable Secrets
-If we make the `*Files` parameters on the `DesiredLRP` mutable fields in BBS then theoretically Cloud Controller could update the `DesiredLRP` when `Secret` contents change and this could be propagated to the container without requiring container recreation.
+If we make the `*Files` parameters on the `DesiredLRP` mutable in BBS, then Cloud Controller could update the `secret`s in the `DesiredLRP`, without requiring container recreation.
 
-For certain CredHub credential types such as `certificate` or `password` we may even be able to expose CredHub credential rotation/[regeneration](https://docs.cloudfoundry.org/api/credhub/version/main/#_regenerate_credentials_endpoint) functionality.
+For certain CredHub credential types such as `certificate` or `password`, we may be able to expose CredHub credential rotation/[regeneration](https://docs.cloudfoundry.org/api/credhub/version/main/#_regenerate_credentials_endpoint) functionality.
--- a/toc/rfc/rfc-draft-bindable-secrets.md
+++ b/toc/rfc/rfc-draft-bindable-secrets.md
@@ -26,10 +26,7 @@ Add two new resources to the V3 Cloud Foundry API: `secret` and the `secret_bind
 
 ### Why Top-Level Resources?
 
-Why not represent secrets as sub-resources of apps like environment variables? Two reasons:
-
-1. Secrets may be larger than environment variables. Modeling them separately lets us naturally keep them in a separate database table and even use separate services, like CredHub, to store their contents.
-2. Secrets may be shared across multiple apps. For example, a "logical app" may actually be several Cloud Foundry apps resources that need to share certain configuration, certificates, or other credentials. By scoping the `secret` at the `space` level, we can easily support this. Similarly, an app using the blue-green deployment pattern can easily share configuration between versions.
+Why not represent secrets as sub-resources of apps like environment variables? Secrets may be shared across multiple apps. For example, a "logical app" may actually be several Cloud Foundry apps resources that need to share certain configuration, certificates, or other credentials. By scoping the `secret` at the `space` level, we can easily support this. Similarly, an app using the blue-green deployment pattern can easily share configuration between versions.
 
 ### `secret` and `secret_binding` Resources
 The `secret` will be a `space`-scoped resource (similar to a `route` or a `service_instance`) that can be bound to one or more `app`s in the `space` using a `secret_binding`.
@@ -138,7 +135,7 @@ cf create-secret SECRET_NAME SECRET_TYPE -p <CREDENTIALS_INLINE_OR_FILE>
 
 If `SECRET_TYPE` is `value` then contents of `-p` will not be parsed. If `SECRET_TYPE` is `json`, `certificate`, or some other CredHub credential type, then the contents of `-p` MUST be valid JSON and may be parsed.
 
-This command will ultimately create a `POST` request to `/v3/secrets` that looks like this:
+This command will ultimately create a `POST` request to `/v3/secrets`. Example `curl` invocation:
 
 ```console
 curl "https://api.example.org/v3/secrets" \
@@ -240,6 +237,8 @@ action := &models.RunAction{
 
 ### Launcher Changes?
 The launcher (or something else) will need to be updated to retrieve credentials from CredHub for `Secret Bindings`. It currently already does this for CredHub references contained within `VCAP_SERVICES` using the app container's instance identity credentials. Is there something other than the launcher that should have this responsibility?
+
+Alternatively, credentials could be fetched by Cloud Controller (it does this already for Service Keys) and passed through directly to Diego. This has the disadvantage of increasing the size of the DesiredLRP and storing storing the secret contents in Diego's database.
 
 ### Other Considerations
 
--- a/toc/rfc/rfc-draft-bindable-secrets.md
+++ b/toc/rfc/rfc-draft-bindable-secrets.md
@@ -231,7 +231,10 @@ The launcher (or something else) will need to be updated to retrieve credentials
 
 ### Other Considerations
 
-#### Maximum Secret size
+#### Not using CredHub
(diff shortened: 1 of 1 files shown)


toc/rfc/rfc-draft-bindable-secrets.md
Gerg, line 1: I'm uncertain about directly exposing the different CredHub credential types. 

Is it worth supporting _both_ `value` and `json`? As proposed, they essentially serve the same purpose (arbitrary secrets), just one file vs many. This adds come conceptual complexity around how the file names are determined, either inferred from the `secret_binding`'s name, or explicitly set in the `secret`'s value.

One option could be to only support json, and then always use the json keys for filenames.

---

Also, I think we eventually run into issues when expanding to support other CredHub types, like the `certificate` type. Per the [CredHub Docs](https://docs.cloudfoundry.org/credhub/credential-types.html), a `certificate` credential looks like:
```json
{
  "ca": "-----BEGIN CERTIFICATE-----...-----END CERTIFICATE-----",
  "certificate": "-----BEGIN CERTIFICATE-----...-----END CERTIFICATE-----",
  "private_key": "-----BEGIN EXAMPLE RSA PRIVATE KEY-----...-----END EXAMPLE RSA PRIVATE KEY-----"
}
```

Since the keys there are hard-coded, the user wouldn't be able to specify the file names using one of the proposed mechanisms. Since one of the objectives is supporting [COTS apps](https://en.wikipedia.org/wiki/Commercial_off-the-shelf),  they may be picky about what their cert files are called, or even want them at different paths from each other.

As I understand it, the value of `certificate`, `password`, etc types in CredHub is they support auto-generation. Outside of that, I think they could essentially be replaced by either `value` or `json`.

We could potentially get around these issues by adding more abstraction between the CC and CredHub resources. So, make it so that CC allows for filename and value configuration and uses the `json` CredHub type under the hood (for instance). In the future, we could add other CredHub types + secret generation, but keep the interface for defining file paths consistent (somehow).

tcdowney: > One option could be to only support json, and then always use the json keys for filenames.
We discussed some of this offline, I'm fully on board with getting rid of the `value` type and always using `json` for the arbitrary secret usecase.

> Since the keys there are hard-coded, the user wouldn't be able to specify the file names using one of the proposed mechanisms. Since one of the objectives is supporting [COTS apps](https://en.wikipedia.org/wiki/Commercial_off-the-shelf), they may be picky about what their cert files are called, or even want them at different paths from each other.

These apps could just use one or more `json` secrets then. I think the question here is whether or not there is value for platform-assisted `credential` secret types. I think regeneration and CA rotation of these could be a very powerful feature. We could cut the scope of this and deal with that later, but I think this was something others were excited about (like @mkocher 🙂). 

> In the future, we could add other CredHub types + secret generation, but keep the interface for defining file paths consistent (somehow).

I think this is a big _somehow_ -- I can't see how it would work unless CCNG handles the CredHub dereferencing or passes down to Diego some sort of key mapping.

Gerg:
cc @cloudfoundry/wg-app-runtime-interfaces-capi-approvers @cloudfoundry/wg-app-runtime-interfaces-cli-approvers @cloudfoundry/wg-app-runtime-platform-diego-approvers

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,254 @@
...
+  SuppressLogOutput: false,
+}
+```
+
+### Launcher Changes?
+The launcher (or something else) will need to be updated to retrieve credentials from CredHub for `Secret Bindings`. It currently already does this for CredHub references contained within `VCAP_SERVICES` using the app container's instance identity credentials. Is there something other than the launcher that should have this responsibility?
+
+Alternatively, credentials could be fetched by Cloud Controller (it does this already for Service Keys) and passed through directly to Diego. This has the disadvantage of increasing the size of the DesiredLRP and storing storing the secret contents in Diego's database.
+
+### Other Considerations
```
sethboyles, line 243: Any thoughts about how secrets would be managed in app/space manifests?

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,254 @@
...
+## Proposal
+
+Add two new resources to the V3 Cloud Foundry API: `secret` and the `secret_binding`, along with associated CRUD API endpoints.
+
+### Why Top-Level Resources?
+
+Why not represent secrets as sub-resources of apps like environment variables? Secrets may be shared across multiple apps. For example, a "logical app" may actually be several Cloud Foundry apps resources that need to share certain configuration, certificates, or other credentials. By scoping the `secret` at the `space` level, we can easily support this. Similarly, an app using the blue-green deployment pattern can easily share configuration between versions.
+
+### `secret` and `secret_binding` Resources
+The `secret` will be a `space`-scoped resource (similar to a `route` or a `service_instance`) that can be bound to one or more `app`s in the `space` using a `secret_binding`.
```
sethboyles, line 32: **question** Both `routes` and `service_instances` are sharable with other spaces.  Do you anticipate 'shared secrets' as well?

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,254 @@
...
+
+### Creating a `secret`
+
+To create a `secret`, users will use a new CLI command: `cf create-secret`.
+
+```console
+cf create-secret SECRET_NAME SECRET_TYPE -p <CREDENTIALS_INLINE_OR_FILE>
+```
+
+If `SECRET_TYPE` is `value` then contents of `-p` will not be parsed. If `SECRET_TYPE` is `json`, `certificate`, or some other CredHub credential type, then the contents of `-p` MUST be valid JSON and may be parsed.
```
sethboyles, line 136: **question:** I seems to recall some issues where users wanted to avoid any secrets being exposed in CLI commands or outputs in case they are using the CLI in CI/CD environments.  Not supporting a file for `value` would expose secrets in those cases. I suppose that would only affect bash scripts being run with `+x` and the like. Is that something worth worrying about?

tcdowney: Oh I didn't intend to mean that it won't support files for `value` secrets with type `-p`, but rereading this I can see how it implies that. I will rework this sentence (especially if we get rid of the `value` type).

toc/rfc/rfc-draft-bindable-secrets.md
tcdowney: Heh, I hadn't considered that. I don't think it would be in scope for an MVP, but that is probably something people would want eventually. That's one of the most used features of Carvel's secret-gen-controller on K8s for example. https://github.com/carvel-dev/secretgen-controller/blob/develop/docs/secret-export.md

What is missing from the design to support this? Does sharing require a separate table?

toc/rfc/rfc-draft-bindable-secrets.md
sethboyles: **suggestion:** The question above aside, I might be simpler to have two flags with two distinct uses:

`-p` reads inline/literal credentials, no matter the `SECRET_TYPE`
`-f` reads credentials from a file, no matter the `SECRET_TYPE`

This would be analogous the `kubectl` `--from-literal` and `--from-file` flags, and wouldn't require users to remember which `SECRET_TYPE` need to be in files and which don't

tcdowney: Cool, that makes sense to me. @Gerg suggested something similar here https://github.com/tcdowney/community/pull/1#discussion_r1794288536 and I initially convinced him with the argument that this mirrors the `cf cups` and `cf bind-service` flag usage, but I'm convinced now that discrete flags is simpler to reason about.

sethboyles: > Oh I didn't intend to mean that it won't support files for value secrets with type -p, but rereading this I can see how it implies that. I will rework this sentence (especially if we get rid of the value type).

Ah, got it.  One last thought then: if `-p`supports either a file or inline-literal, then mistyping a file name could result in a secret with an incorrect value (e.g. setting your password to `mysecrets.josn`)

Gerg: Or only accept files, and teach people how to use `<()` 😉

toc/rfc/rfc-draft-bindable-secrets.md
sethboyles: I'm not super familiar with the resource sharing code, but I think `routes` and `service_instances` are linked to other spaces via a separate table/model (`RouteShare` and `ServiceInstanceShare`).

Mostly asking as I'm wondering if we should start thinking about making the Sharable modeling or code more generic for if/when we want to make secrets (or future resources) sharable, as it seems to be becoming a pattern.

toc/rfc/rfc-draft-bindable-secrets.md
sethboyles: Oh, I didn't realize we already had that pattern in the CLI. That is a strong argument

Although I have to admit I first was thinking that we used `-f` for files when applying manifests, so we should use the same pattern here 😄

Gerg: Maybe we can avoid this pattern in CLI v9

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,254 @@
...
+ ServiceBindingFiles: []*models.Files{
+    {
+      Name: "/etc/cf-instance-binding",
+      Value: "VALUE",
+    },
+  },
+  Secrets: []*models.Secret{
+    {
+      Name: "SECRETNAME",
+      MountPath: "/etc/conf/",
```
tcdowney, line 226: @winkingturtle-vmw @ameowlia 

Are there any paths we'd need to prevent here that would interfere with Diego or otherwise be bad? Allow app devs to drop arbitrary files _anywhere_ could be bad.

toc/rfc/rfc-draft-bindable-secrets.md
tcdowney: Updated in ff25954

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,292 @@
...
+
+```console
+cf bind-secret APP_NAME SECRET_NAME BINDING_NAME
+```
+
+### CLI Changes
+Work will need to be done to implement the following commands:
+- `cf create-secret`
+- `cf update-secret`
+- `cf bind-secret`
```
Samze, line 229: **thought** For a user with no knowledge of the inner workings of this, they may create a secret, bind it and expect to see it somewhere in the apps ENV and be confused. Not sure we can do much here beyond making sure the commands behaviour are well documented.

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Problem
+
+Cloud Foundry currently supports app configuration via environment variables. Other application platforms, such as Kubernetes, support app configuration at runtime [via volume-mounted files](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod). Many apps are adopting this file-based configuration approach, since environment variables have length limits and files can be updated without having to recreate the container. This RFC proposes introducing file-based configuration to support these sorts of apps on Cloud Foundry — without requiring app code changes.
+
+### What About User-Provided Services?
+
+Why can't developers use user-provided service instances (UPSIs) and service bindings for file-based app configuration? Some apps may have file-based config that don't naturally fit into the service binding model. For example, if an app relies on a config file at a certain path, that plath may not be configurable to `$SERVICE_BINDING_ROOT`. This is especially the case for commercial off-the-shelf software (COTS).
```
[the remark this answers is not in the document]
stephanme: [RFC 0030 - Add Support for File based Service Binding Information](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0030-add-support-for-file-based-service-binding.md) requires to configure an app feature flag and then switches all service bindings from VCAP_SERVICES to files under `$SERVICE_BINDING_ROOT`.

I could imagine that secrets and secret bindings could allow a smoother migration path from VCAP_SERVICES to file based service bindings by allowing both mechanisms in parallel.

If secret bindings are allowed to mount under `$SERVICE_BINDING_ROOT`, a prioritisation is needed between service bindings and secret bindings.

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,292 @@
...
+- Start Date: 2024-10-08
+- Author(s): @tcdowney @gerg
+- Status: Draft
+- RFC Pull Request: https://github.com/cloudfoundry/community/pull/994
+
+## Summary
+
+**tl;dr** Kubernetes-style Secrets for Cloud Foundry
+
+This RFC proposes adding a `secret` resource to the V3 Cloud Foundry API, similar in design and purpose to the Kubernetes [Secret resource](https://kubernetes.io/docs/concepts/configuration/secret/). Cloud Foundry Secrets will store arbitrary data in [CredHub](https://docs.cloudfoundry.org/credhub/) inject it into app containers via `tmpfs`-mounted files at a given path, using the same mechanism as [RFC 0030 - Add Support for File based Service Binding Information](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0030-add-support-for-file-based-service-binding.md).
```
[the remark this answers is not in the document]
tcdowney: That's good to know. I was under the impression that CredHub was a required CF component at this point since it had been a default cf-deployment component for so long. Out of curiosity why don't y'all use it? I feel it has a much stronger security posture than Cloud Controller.

I can rework this to be CredHub optional of course.

toc/rfc/rfc-draft-bindable-secrets.md
tcdowney: As part of updating this to not require CredHub, I removed the `value` type and renamed `json` to `opaque`:
274b3b0

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,254 @@
...
+  Args: []string{"some", "args to", "pass in"},
+  Dir: "/path/to/working/directory",
+  User: "username",
+  EnvironmentVariables: []*models.EnvironmentVariable{
+    {
+      Name: "ENVNAME",
+      Value: "ENVVALUE",
+    },
+  },
+ ServiceBindingFiles: []*models.Files{
```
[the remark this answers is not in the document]
tcdowney: Yeah, I agree.

@beyhan do you have thoughts on how to make this cleaner? How far along is the Service Binding Files work for us to potentially generalize the BBS interface?

toc/rfc/rfc-draft-bindable-secrets.md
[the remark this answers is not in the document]
tcdowney: I took a pass at making CredHub optional: 274b3b0

toc/rfc/rfc-draft-bindable-secrets.md
tcdowney: I think that's fine. It's a new feature that I expect primarily to be used by users coming from Kubernetes. You did remind me though that it would be interesting to map secrets to env vars like in k8s. This could be useful if the secrets are stored in CredHub. I'm not sure what a clean UX for that would look like, though.

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,254 @@
...
+}
+```
+
+To support this, Cloud Controller will need to have both read and write permissions for the `secret`'s corresponding credential in CredHub.
+
+### Creating a `secret_binding`
+To create `secret_bindings`, users will use a new CLI command: `cf bind-secret`. This will create a `POST` request to `/v3/secret_bindings`.
+
+```console
+cf bind-secret APP_NAME SECRET_NAME BINDING_NAME
```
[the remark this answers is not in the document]
tcdowney: Eh, I don't think a default should exist. Omitting it from the `cf bind-secret` example is a miss on my part. Fixed: c767d5e

As for restricting it, I think we probably should. I'm not sure whether or not an allow-list vs deny-list approach would be best -- looking for opinions here as well https://github.com/cloudfoundry/community/pull/994#discussion_r1799767455.

toc/rfc/rfc-draft-bindable-secrets.md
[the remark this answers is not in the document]
tcdowney: Hmm, I can see arguments for both. My gut tells me that service bindings should take priority, but I also think it is most flexible for devs if secrets take precedence.

For example, imagine you have a MySQL instance hosted on one instance of CF that creates publicly reachable service instances (exposed via TCP routes) and you want to bind to these from apps on other Cloud Foundry installations. One workflow would be to make a Service Key on the "services CF" and then create UPSIs on the ones where your apps are running based on that key.

Problem is an UPSI does not have the correct `type` that helper library might expect (and not all libraries look at tags). If you could use a secret for this then you could explicitly set the keys to what you want to as a dev. But semantically the UPSI still makes more sense here -- so maybe that interface just needs to be more flexible.

Thoughts @Samze @stephanme @Gerg ?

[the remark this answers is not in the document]
Gerg: We have a kinda similar situation with environment variables. We intentionally[ restrict users from setting environment variables that conflict with VCAP_SERVICES and friends](https://github.com/cloudfoundry/cloud_controller_ng/blob/c6078aab8821a6fcdc9078f3e0fe988d3ca8a122/app/messages/validators.rb#L81-L103), but there can be [collisions between global environment variables and app-level ones](https://github.com/cloudfoundry/cloud_controller_ng/blob/c6078aab8821a6fcdc9078f3e0fe988d3ca8a122/lib/cloud_controller/diego/environment.rb#L39-L49).

toc/rfc/rfc-draft-bindable-secrets.md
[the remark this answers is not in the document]
beyhan: I think my comment https://github.com/cloudfoundry/community/pull/994#discussion_r1801149230 is also relevant here.

toc/rfc/rfc-draft-bindable-secrets.md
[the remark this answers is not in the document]
blgm: We acknowledge that there is some overlap between bindable secrets and user-provided service instances ("UPSIs"); and also to some extent "managed" (broker provided) services ("MSIs"), particularly when/if [RFC 0030 - Add Support for File based Service Binding Information](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0030-add-support-for-file-based-service-binding.md) is implemented too. I can foresee folks struggling to differentiate: should I use a bindable secret or an UPSI?

If we do this right, we could _simplify_ the platform - for example by making UPSIs and MSI subtypes of bindable secrets, and deprecating UPSIs. If we do this wrong then we could add a forever confusion about what's the right choice, and then have to document and train folks out of the mess.

Perhaps this is too radical, but I'd be interested on folk's thoughts about:
- deprecating UPSIs and changing the commands to recommended a bindable secret instead (but not breaking anything yet)
- teaching that MSIs are a special type of bindable secrets rather than a different concept

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,293 @@
...
+```json
+"value": {
+  "application.yml" : "---\napplication:\n ...",
+  "secret.txt": "my-secret"
+}
+```
+
+The above `secret` would mount files named `application.yml` and `secret.txt` at the path configured in the binding.
+
+### Creating a `secret`
```
blgm, line 123: There's not a lot in this doc yet about permissions. I infer that an app dev with access to an App will be able to create and bind secrets, and that makes sense, but what if we go further? In the past folks have talked about an Independent Service Manger (ISM) that's a bit like a service broker, but is independent of the platform and might not implement OSBAPI. What if there was a way to give an off-platform service the ability to create secrets, but have no access to Apps? For example if I'm connecting to a MySQL on AWS, I might use AWS Credential Manager to rotate my password every few days. What if a Lambda on AWS could be triggered by this change, and update a secret in CF? We would not want that Lambda to have access to Apps, or other secrets on CF. But it would be ok if could update its own secret. Equally we might allow LetsEncrypt to update a secret that's a certificate. Maybe this is out of scope for now, but it would be nice to make sure we don't block anything like that in the future.

toc/rfc/rfc-draft-bindable-secrets.md
joaopapereira: Secrets are not shared with other namespaces in secretgen. What it does is the reconciler will copy these secrets from the source to the other namespaces and ensure that whenever there is a change in the original secret it will get copied to the namespaces where the export is done to.
Is your suggestion to do something similar @tcdowney? 

Another thing that came to mind when I was reading the proposal was the ability to have Organization level secrets, pretty much like github does with their secrets where you can define secrets in the organization and they can be used in all the repositories. Is this something that yll thought about or want to make possible one day?

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,293 @@
...
+cf create-secret SECRET_NAME -t opaque -p <CREDENTIALS_INLINE>
+```
+
+```console
+cf create-secret SECRET_NAME -t opaque -f <FILE_PATH_TO_CREDENTIALS_FILE>
+```
+
+* `-t` the secret `type`, defaults to `opaque`
+* `-p` reads inline/literal credentials
+* `-f` reads credentials from a file
```
joaopapereira, line 137: Is this flag going to be repeatable? something like `cf create-secret bananas -f file1.yaml -f file2.yaml`? If so how would the secrets be represented?

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,293 @@
...
+
+```console
+cf create-secret SECRET_NAME -t opaque -p <CREDENTIALS_INLINE>
+```
+
+```console
+cf create-secret SECRET_NAME -t opaque -f <FILE_PATH_TO_CREDENTIALS_FILE>
+```
+
+* `-t` the secret `type`, defaults to `opaque`
```
joaopapereira, line 135: If we do not have a different type right now is it worth it to add this flag?

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,293 @@
...
+      "secret.txt": "updated-secret-value"
+    }
+    "metadata": {
+      "annotations": {},
+      "labels": {}
+    }
+  }'
+```
+
+### Reading a `secret` and Its Value
```
joaopapereira, line 199: Is there going to be an endpoint to get the full secret "object"?
Is there going to be a CLI command to retrieve the secret?

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,293 @@
...
+  
+"value" : {
+  "some-json-key" : "some-json-value"
+}
+```
+
+To support this, Cloud Controller will need to have both read and write permissions for the `secret`'s corresponding credential in CredHub.
+
+### Creating a `secret_binding`
+To create `secret_bindings`, users will use a new CLI command: `cf bind-secret`. This will create a `POST` request to `/v3/secret_bindings`.
```
joaopapereira, line 215: Is this going to force a reload of the application to ensure that secrets will be available or the user will have to do it afterward manually?

toc/rfc/rfc-draft-bindable-secrets.md
geofffranks: Can we have the secret fail to mount + fail to start the app, if the mount path is already an existing directory on the container filesystem? 

These mounts would replace the directory structure they're mounted on top of, rather than be additive, right? If so, this should prevent most problems related to masking system directories/files, or accidentally masking app code or other required resources in the app container.

ameowlia:
❓ Are we worried about different secrets writing to the same place?

1. I make a secret named "meow-secret" with value `{"meow.txt": "meow"}` and bind it to my app.
1. I make a secret named "also-meow-secret" with value`{"meow.txt": "woof"}` and bind it to my app.
1. What happens here? Is there an error? Is it non-determanistic?

toc/rfc/rfc-draft-bindable-secrets.md
ameowlia: ❓ Could we make a dir where the secrets always go to? Where users could make nested dirs if they want. But this way we could limit exactly where they are putting things and not need to worry about it?

tcdowney:
> ❓ Are we worried about different secrets writing to the same place?
> 
> 1. I make a secret named "meow-secret" with value `{"meow.txt": "meow"}` and bind it to my app.
> 2. I make a secret named "also-meow-secret" with value`{"meow.txt": "woof"}` and bind it to my app.
> 3. What happens here? Is there an error? Is it non-determanistic?

One option is to make `mount_path` unique per app in the `secret_binding`, but I feel there are valid reasons to allow multiple secrets to be bound to the same `mount_path` dir. In my opinion it's fine if this is non-deterministic since the space dev is in control of what happens here (they chose to mount them both to the same path).

toc/rfc/rfc-draft-bindable-secrets.md
tcdowney: I think for a new binding developers will need to `cf restart` their app the same way they do to make env vars and service binding changes take effect. I believe this is the same behavior as Kubernetes when you add new volume mounts to a Pod spec. For updates of existing bindings I think the same is true, at least to start, but we could explore secret changes not requiring a restart in the future.

geofffranks:
> > ❓ Are we worried about different secrets writing to the same place?
> > 
> > 1. I make a secret named "meow-secret" with value `{"meow.txt": "meow"}` and bind it to my app.
> > 2. I make a secret named "also-meow-secret" with value`{"meow.txt": "woof"}` and bind it to my app.
> > 3. What happens here? Is there an error? Is it non-determanistic?
> 
> One option is to make `mount_path` unique per app in the `secret_binding`, but I feel there are valid reasons to allow multiple secrets to be bound to the same `mount_path` dir. In my opinion it's fine if this is non-deterministic since the space dev is in control of what happens here (they chose to mount them both to the same path).

If we go the route of preventing mount path from mounting on top of an existing directory, this sorts itself out with no additional logic at the CAPI layer. Otherwise it could get expensive to check that the credentials inside a single secret do not conflict with any of the mount paths/credentials of any of the other secrets that are bound to any of the apps bound to the secret being updated.

toc/rfc/rfc-draft-bindable-secrets.md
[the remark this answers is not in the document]
tcdowney: I know this may be too late, but is the name of the field `ServiceBindingFiles` set in stone? I'm wondering if we could make it more generic -- even `VolumeMountedFiles` or something. I just don't know how far along the file-based service binding implementation is at this point.

cc @Gerg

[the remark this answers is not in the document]
Gerg: Ideally I would have noticed this in the file-based binding RFC, but this does feel like CF-level concepts leaking into BBS's domain model. Historically, BBS has tried to separate itself from CF concepts (e.g. `app` vs `LRP`).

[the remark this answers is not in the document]
beyhan: > Ideally I would have noticed this in the file-based binding RFC, but this does feel like CF-level concepts leaking into BBS's domain model. Historically, BBS has tried to separate itself from CF concepts (e.g. `app` vs `LRP`).

I think at the stage which we are with the implementation it should be possible to change `ServiceBindingFiles` to `VolumeMountedFiles`. @stephanme, @PlamenDoychev any thoughts on this?

philippthun:
> > ❓ Are we worried about different secrets writing to the same place?
> > 
> > 1. I make a secret named "meow-secret" with value `{"meow.txt": "meow"}` and bind it to my app.
> > 2. I make a secret named "also-meow-secret" with value`{"meow.txt": "woof"}` and bind it to my app.
> > 3. What happens here? Is there an error? Is it non-determanistic?
> 
> One option is to make `mount_path` unique per app in the `secret_binding`, but I feel there are valid reasons to allow multiple secrets to be bound to the same `mount_path` dir. In my opinion it's fine if this is non-deterministic since the space dev is in control of what happens here (they chose to mount them both to the same path).

At some point CAPI builds this "file structure" passed to Diego, i.e. it iterates over all the secrets and adds them to the `VolumeMountedFiles` array. Here conflicts will be easily detectable and should - from my point of view - result in an error.

toc/rfc/rfc-draft-bindable-secrets.md
```diff
@@ -0,0 +1,293 @@
...
+      "href": "https://api.example.org/v3/secret_bindings?secret_guids=4f00b75b-3455-48f1-ba4a-f3b82ebad943"
+    },
+    "value": {
+      "href": "https://api.example.org/v3/secrets/4f00b75b-3455-48f1-ba4a-f3b82ebad943/value"
+    }
+  }
+}
+```
+
+A `secret` has a type that identifies its contents. To start we will support the `opaque` credential type (corresponds to the `json` [CredHub credential type](https://docs.cloudfoundry.org/credhub/credential-types.html)) which is designed to hold arbitrary user-defined data. In the future, this can be expanded to support more semantic secret types such as those [supported by Kubernetes](https://kubernetes.io/docs/concepts/configuration/secret/#secret-types) or other CredHub-supported types like `certificate`. `opaque` secrets consist of a JSON object whose keys map to files and values map to file content.
```
philippthun, line 72: **question:** Can keys only be filenames or also relative paths?

For example:
- The secret binding's `mount_path` is `/etc/my-conf/`.
- The secret's `value` is `{"passwords/db_user.txt":"abc123","certificates/db_cert.pem":"..."}`.

Would result in:
```
etc
└── my-conf
    ├── certificates
    │   └── db_cert.pem
    └── password
        └── db_user.txt
```

tcdowney: I feel like supporting relative paths adds complexity that could be resolved by just using multiple secrets. For example, I'd expect many config files might contain JSON and it would be nice if we could avoid parsing the values themselves and just treat it as opaque file contents.

toc/rfc/rfc-draft-bindable-secrets.md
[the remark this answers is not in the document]
beyhan: FYI: https://github.com/cloudfoundry/community/pull/1008

toc/rfc/rfc-draft-bindable-secrets.md
beyhan: > Whether or not secrets themselves should be defined in a v2 space manifest is in another question.

@tcdowney what do you mean by this?

toc/rfc/rfc-draft-bindable-secrets.md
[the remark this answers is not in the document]
beyhan: > * deprecating UPSIs and changing the commands to recommended a bindable secret instead (but not breaking anything yet)

My thoughts on this. I think we can have a chance here if we could offer the migration of UPSis to bindable secrets without requiring any changes in the application code. That means, the libraries parsing service binding information need to be adjusted and do the heavy lifting. I'm not sure whether this means also, that bindable secrets need to support not only file based approach for the service binding but also environment based because the move from UPSis  to bindable secrets will mean also to opt-in into the file based service binding approach.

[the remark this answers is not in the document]
tcdowney: I don't think we should deprecate UPSI's. They are still semantically distinct from a generic Secret -- they provide a way to provision services off platform (either on a separate Cloud Foundry instance or via some other mechanisms) and provide _service_ information to apps through similar a similar UX as a fully managed service.

I think although Secrets could be used to accomplish something similar, we would be going against over a decade of ingrained UX for CF developers. I also think supporting file-based service bindings for UPSIs is important for service offerings (both older CF oriented ones and newer K8s ones) that may not conform exactly to what the various service binding loading application libraries expect. If a specific key is slightly different, for example, an UPSI can be used to massage its presentation to what the library would need.

I guess in short, there are actually some use-cases that file-based UPSIs would provide that I am interested in leveraging and I wouldn't want to gate them behind a non-implemented (and not yet approved) Secrets proposal. 😅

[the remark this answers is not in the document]
Gerg: I agree with Tim.

Similarly, there is "duplication" between `cf cups` and `cf set-env` letting you set environment variables in app containers. You _could_ theoretically achieve logical "service bindings" with `cf set-env`, and you could configure your app using UPSIs. However, we still keep both of those around because they serve semantically different purposes, are intended for different user workflows, and affect different portions of the app container's environment variables.

toc/rfc/rfc-draft-bindable-secrets.md
tcdowney: @Gerg is working on a draft proposal of a "v2" version of the CF App manifest that is more declarative and includes the ability to configure other space scoped resources. This is still WIP, but he has a version of it here:
https://github.com/Gerg/community/blob/rfc_v2-app-manifests/toc/rfc/rfc-draft-v2-app-manifests.md#add-secrets

toc/rfc/rfc-draft-bindable-secrets.md
tcdowney: I was just mentioning some prior art. I think this is an interesting and possible future enhancement for Secrets, but I want to keep scope limited for now. If we develop Secrets and this use case comes up we can revisit in a follow-up RFC.

toc/rfc/rfc-draft-bindable-secrets.md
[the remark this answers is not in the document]
tcdowney: Thanks @beyhan , I will update this RFC to reflect those changes.

toc/rfc/rfc-draft-bindable-secrets.md
tcdowney: Yeah, I think this is out of scope for now. I will add a section around permissions mentioning that Secrets can be created and managed by users with the `SpaceDeveloper` role since this will treat them the same as other sensitive resources like app environment variables and service binding credentials.

toc/rfc/rfc-draft-bindable-secrets.md
tcdowney: I think it should be a single flag and support a single file for now. If the CLI ever wants it to support an array of files then that could be a future enhancement (but I don't want to be too prescriptive of the CLI UX here).

toc/rfc/rfc-draft-bindable-secrets.md
tcdowney: Yeah, we don't need it for now. I'll update this and remove the flag.

toc/rfc/rfc-draft-bindable-secrets.md
tcdowney: There probably should be a CLI command, but as it stands having a separate endpoint for the non-sensitive bits of the Secret and the Secret value has parity with how env vars and service binding credentials work. This is mostly necessary because of how permissions work in Cloud Controller (we want some roles to be able to see the existence of a Secret / Binding to an app, but not its contents).

The CLI could call both endpoints for `SpaceDeveloper` users and show more details if that's desired.

tcdowney:
> **question**: Are we aware of any encoding constraints? For example, are we expecting binary secret data encoded to base64, or is this not required?

This is an interesting point. I hadn't considered those types of secrets, but that would be important to fully support COTS / k8s-deployed apps that have non-unicode secrets. Maybe I should update this to look more like the K8s implementation (stored and read as base64, but users can provide `data:` or `stringData:` when creating one). 🤔

beyhan:
@tcdowney, @Gerg I see that there are still open discussions and pending updates to this RFC. Pls let me know when you think that it is ready for final review and then staring the FCP.

beyhan:
ping @tcdowney, @Gerg ^

tcdowney:
Hey @beyhan , I am going to retract this for now. The use case we originally wanted it for can be solved well enough with file-based service bindings and I don't think there is interest in implementing this on our end at this time. I think the idea is still valid and interesting, though, so if anyone wants to take this idea and run with it I'm in full support. 🙂

PR #1004, opened 2024-10-21
Bosh Templating Language RFC

jpalermo:
Proposal for adding a new templating language for bosh releases


rkoster (changes requested):
Would it be possible to add some examples of what the new grammar would look like for at least the basic usecases (`if_p`, `p('something').each do |prop|`, multi line certs (with respecting indentation, some to_json/yaml, etc)?

toc/rfc/rfc-draft-add-new-bosh-release-templating-language.md
```diff
@@ -0,0 +1,47 @@
...
+The Bosh Director is written in Ruby, so it will be able to shell out to the CLI to render templates. While
+the Bosh CLI is written in golang so it will be able to import the library and use it directly.
+
+The grammar would be kept relatively small. We will not be providing a direct replacement for ERB. Many validations
+and data transformations are happening within ERB templates right now that should be happening at different layers
+of the deployment process.
+
+Data transformation should be happening within the release runtimes or pre-start scripts. Validation can also happen
+there, although it's possible we want to expand the Bosh Release `spec` files to provide some level of validation so
+errors can be caught early in the deploy process (that work is NOT part of this RFC).
```
ameowlia, line 47: ARP does _A LOT_ of validation in the template. I always thought this was best practice because then the user gets an error far earlier in the deploy process. 

Here are some examples of the validations we have: 
* provided IPs are valid IPs
* integer ranges (must be larger than X or smaller than X)
* do domains end with a "."? If not, add one.
* if you set X, Y, Z property, then fail if a cert is not provided. Else it is fine. 
* a provided string must be either "a" or "b"
* string length must be more than X

ameowlia:
ARP releases  do a lot of "check to see if this property is provided, if not then use the property from a bosh link". Is this type of logic someone that would be provided? 

But really, I second what @rkoster said above asking for examples of what you want to include in your templating language, that way I can be more specific in my worries 😅

ameowlia:
I come bearing examples of template logic!

* Most of our jobs use a function like this one to do the "check to see if this property is provided, if not then use the property from a bosh link" logic. [link](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L3-L17)
* [IP validation](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L20-L29). We do this in many jobs.
* basic int logic greater than/less than. [link](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L31-L36)
* string must be one of a set. [link](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L38-L44)
* type checking, it must be an array. [link](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L57-L59)
* if the property is provided, and once we strip whitespace, then it must be longer than 50 chars. [link](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L61)
* requiring certain properties, but only when other properties are set. [link](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L200-L202)
* turn bosh machine info into json. [ink](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L230)
* if a certain thing is a certain number, then do all these other things. [link](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L255-L262)
* cert validation [link](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L484)
* validating cert SANs [link](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L499-L521)
* at least one CA has a certain SAN [link](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/gorouter/templates/gorouter.yml.erb#L525-L532)
* and those examples are all from gorouter!
* here we use a property as a deprecation warning to alert users that nats (non-tls) is going away soon. [link](https://github.com/cloudfoundry/routing-release/blob/8c155e5d30e2f9a62449a35ab6759aaf0a10b949/jobs/route_registrar/templates/registrar_settings.json.erb#L16)
* pulling out db params from a string [link](https://github.com/cloudfoundry/diego-release/blob/f4e3d1c0c7a193f26182451ff85f0b8ff20056a1/jobs/locket/templates/locket.json.erb#L40-L58)
* turn array into a comma separated list [link](https://github.com/cloudfoundry/diego-release/blob/f4e3d1c0c7a193f26182451ff85f0b8ff20056a1/jobs/route_emitter/templates/route_emitter.json.erb#L72-L74)

After going through all these files I see that nearly every one of ours has this helper method defined. It would be great if you owned these types of helper functions :) 
```
  def parse_ip (ip, var_name)
    unless ip.empty?
      ip = ip.split(":")[0]
    begin
      parsed = IPAddr.new ip
    rescue  => e
      raise "Invalid #{var_name} '#{ip}': #{e}"
    end
    end
  end
  ```

toc/rfc/rfc-draft-add-new-bosh-release-templating-language.md
ameowlia: added more specific examples with links in this comment: https://github.com/cloudfoundry/community/pull/1004#issuecomment-2432585910

maxmoehl:
If we want to push logic out of the templates why not drop the central template rendering altogether and allow release others to consume the YAML directly? The provided YAML would be placed in the config directory of the job and the release decides how to proceed with it.

This would also work around the core issue here which I think is dependency management: the release authors are implementing against a templating system which is controlled by the user of the release (e.g. the operator) and not the release author. By explicitly making the template rendering a part of the release itself we solve that issue. Bumping the template renderer becomes the responsibility of the release author which can properly test and resolve any issues with version bumps. And the operator no longer risks breaking a bunch of releases just because the ruby version or some required dependency is bumped in their environment.

If we still want BOSH to be able to distinguish between the templating and pre-start phase we could add an additional lifecycle step. If the release author provides a script for that step the ERB rendering is automatically disabled. And for release which don't want to switch a Ruby script which mimics the behaviour of the old template rendering could be provided / injected to speed up the transition.

jpalermo:
Moving this back into draft for now as we work on a solution for job property validation first

peanball:
Please consider IPv6 for the `parse_ip` provided by @ameowlia. `.split(":")` will do us no good here in IPv6 land 😅

mrosecrance:
For now, we (the Broadcom Bosh team) are pausing progress on this. Joseph has left our team and it's less of an immediate priority. If someone else would like to iterate on this in the meantime, feel free, I expect we'll take another look at this in 6ish months.

PR #1049, opened 2025-01-20
Bosh Release Property Validation RFC

jpalermo:
[Markdown View of the RFC](https://github.com/cloudfoundry/community/pull/1049/files?short_path=aef35ea#diff-aef35eadf07abce113df557b9f05d5783ea4c12fb3b0e2b630cae8aa7bf539ad)

### DIFF
--- a/toc/rfc/rfc-draft-add-bosh-release-property-validation.md
+++ b/toc/rfc/rfc-draft-add-bosh-release-property-validation.md
@@ -55,3 +55,72 @@ There are some existing property validations that may go beyond the scope of thi
 has the ability currently for the deployer to pass in validation rules for certificates. Those rules are then applied against
 a different property entirely. This is probably possible to implement on the bosh director side in a custom certificate validator,
 but it doesn't seem like something that would benefit more than this singular job.
+
+## Timeline
+
+We would like to get feedback on this through the first week or two of February and then implement the functionality on 
+the Bosh Director in the second half of February. A new major release of the director would be released at the end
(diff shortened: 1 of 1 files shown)


beyhan:
During the TOC meeting on 28.01.2025 we decided to start the FCP with the goal to accept this RFC.

cweibel:
I like the idea of the validation on the director side.  It can be frustrating to generate all your certificates in credhub only to find out after hours of debugging that the SAN configuration was wrong.

ameowlia (approved):

rkoster (approved):

stephanme (approved):

ChrisMcGowan (approved):

beyhan (approved):

jpalermo:
Hi @teddyking ,
I've not actually used json schema much, but looking around, maybe [$data](https://json-schema.org/blog/posts/hyperborea#referencing-other-parts-of-the-schema) could be used to accomplish the first thing. 

And it looks like [uniqueItems](https://json-schema.org/understanding-json-schema/reference/array#uniqueItems) does the second bit too. Seems like the ability to do uniqueness on non-primitive types must have been added in a recent draft of the schema because some older info says it isn't possible, but I tried it out and it seemed to work with objects or nested arrays just fine.

We are open to adding validations that would be useful to multiple releases. But we probably want to keep the number of custom validations at least somewhat limited. So for more specific things, the recommendation would be to put them in the application logic or a pre-start script of some kind.

PR #1060, opened 2025-01-29
Remove bot user from ARI Terraform Provider

stephanme:
SAP-OSPO-ADMIN github user was invited to cloudfoundry org but the invitation was rejected or expired. As a consequence, the org automation fails.


stephanme:
@pyogesh2, @vipinvkmenon: Please confirm with a thumbs up.

beyhan (approved):

PR #1065, opened 2025-01-31
Remove myself as a tech lead for the CF on K8s WG

davewalter:


beyhan:
Hi @georgethebeatle,

The majority of the TOC must approve the lead change before merging this as described in https://github.com/cloudfoundry/community/blob/f868482da99200516c9a54cd9daac21193e5416b/toc/rfc/rfc-0024-promoting-or-revoking-wg-leads.md?plain=1#L46-L49

georgethebeatle:
Sorry @beyhan, I was not aware of this procedure. In this specific case I think it is no big deal, since the lead is revoking his own role, so refutation is not to be expected, but next time I will make sure I follow these rules more strictly. Thanks for pointing this out!

PR #1077, opened 2025-02-07
rfc: IPv6 Dual Stack Support for Cloud Foundry

peanball:
This PR adds the RFC "IPv6 Dual Stack Support for Cloud Foundry".

For easier viewing, you can see [the full RFC](https://github.com/sap-contributions/cf-community/blob/rfc-ipv6-in-cf/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md) as preview.

### DIFF
--- a/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
+++ b/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
@@ -146,13 +146,34 @@ Finally, some adaptations for the containerized workloads are necessary:
 1. The [environment variables](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#app-system-env) `CF_INSTANCE_IP` and `CF_INSTANCE_INTERNAL_IP` will need IPv6 equivalents, or extension.
 2. Instance identity IDs, i.e. the certificates used by Envoy in the app container, need to support IPv6 addresses. Currently they are issued for the IPv4 address only.
 
-### silk-release
+### Silk
+1. Silk Daemon is responsible for retrieving the network lease for the cell from Silk Controller and providing it to the CNI Plugin.
+   - Enhancement: Add a new configuration property for an IPv6 prefix.
+   - Behavior:
+     - If an IPv6 prefix is provided and the host is IPv6 enabled, Silk Daemon returns both IPv4 and IPv6 networks.
+     - If no IPv6 prefix is provided, the system defaults to IPv4-only behavior.
+2. CNI Plugin is responsible for setting up the network devices on both container and host side. It delegates the IPv4 CIDR from the silk daemon to the IPAM host-local plugin, which allocates addresses from the network.
+   - Enhancements:
+     - Extend the existing logic to support IPv6 address assignment on already created devices.
+     - Configure point-to-point communication, routing tables, and static neighbors (similar to ARP tables in IPv4).
+     - Enable IPv6-specific sysctl security settings.
+     - Store the IPv6 address in the Silk datastore.
+   - Behavior:
+     - The plugin only configures IPv6 if Silk Daemon returns an IPv6 prefix and the host supports IPv6.
+3. CNI Wrapper Plugin currently creates default iptables rules (`netin-*`, `netout-*`, `overlay-*` chains) with default `REJECT` rules and `ALLOW` rules for DNS and "host services." In deployments without vxlan-policy-agent enabled, it also manages iptables egress rules.
+   - Enhancement: Extend it to apply IPv6 firewall rules using ip6tables:
+     - Maintain default REJECT rules for IPv6.
+     - Allow DNS and host service traffic.
+     - Ensure IPv6 egress rules are correctly handled when vxlan-policy-agent is disabled.
+4. VXLAN Policy Agent creates the dynamic iptables rules for ASGs and C2C Policies.
+   - Enhancement: Extend it to support IPv6 ASGs, ensuring that:
+     - Security group rules from Policy Server are properly filtered and applied based on IP version.
+     - The system supports both IPv4-only and dual-stack ASG enforcement.
 
-The CNI plugin needs to be extended to support IPv6 address assignment. Currently, container-facing interfaces are created and assigned an IPv4 address using the host-local IPAM plugin. The required enhancement involves assigning an IPv6 address alongside IPv4 and updating the state files accordingly.
-
-The CNI wrapper plugin must be extended to support IPv6. Currently, it creates default iptables rules (`netin-*`, `netout-*` chains) with default `REJECT` rules and `ALLOW` rules for DNS and "host services." In deployments without vxlan-policy-agent enabled, it also manages iptables egress rules. The enhancement requires adding logic to handle IPv6 rules using ip6tables.
-
-Additionally, the VXLAN Policy Agent must be updated to support IPv6 Address Security Groups (ASGs) received from the Policy Server.
+General Behavior and Compatibility: 
+- The existing IPv4 functionality remains unchanged.
+- If the host supports IPv6 and a prefix is provided to Silk Daemon, additional IPv6 configuration will be applied on top of the current setup.
+- If VXLAN Policy Agent is configured with IPv6 support, dynamic ASGs will also be created as ip6tables rules.
 
 ### Cloud Controller
 
--- a/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
+++ b/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
@@ -69,11 +69,11 @@ networks:
     subnets:
       - az: z1
         cloud_properties:
-          # assign non-overlapping `/80` cidrs from `range`
-          ipv6_prefix_delegation_size: 80
           security_groups:
             - sg-abcd...
           subnet: subnet-12345
+        # assign non-overlapping `/80` cidrs from `range`
+        ipv6_prefix_delegation_size: 80
         dns:
           - 2001:db8:4e04:D139:0000:0000:0000:0253
         gateway: 2001:db8:4e04:D139:0000:0000:0000:0001
--- a/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
+++ b/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
@@ -53,7 +53,8 @@ BOSH is used to provision and configure the VMs that run the components of Cloud
 
 1. Assignment of IPv6 addresses in addition to IPv4, ideally to the same NIC
 2. Assignment of IPv6 CIDR ranges (prefix delegation), e.g. assign a `/80` address, while avoiding overlapping. This is relevant for the use case of address delegation to application workloads hosted on Diego cells.
-
+   In the first iteration, assignment will be of only a single CIDR range (prefix delegation) per VM. This can be extended in the future.
+ 
 #### BOSH Director
 
 Depending on the configuration of the prefix in the cloud config networks section. The Director should:
@@ -122,6 +123,10 @@ instance_groups:
       ipv6_prefix_delegation_size: 80
 
 ```
+
+The preference is to use the deployment-based expression for the prefix size, as it provides better control to the deployment and does not require changes to the cloud config.
+
+Note: Changes to the prefix size must be taken into account during updates carefully, so address ranges remain non-overlapping at all times.
 
 #### BOSH Cloud Provider Interfaces (CPIs)
 
@@ -293,6 +298,11 @@ Cloud Foundry acceptance tests shall be extended to also exercise IPv6 communica
 bosh-bootstrap shall be extended to support IPv6 / dual stack configuration.
 
 bosh-bootstrap is used, among other things, for setting up the environments that are running CATs.
+
+The extensions in bosh-bootstrap include:
+1. Changes to the Terraform configuration to support IPv6 in VPCs and subnets
+2. Use the changes in BOSH to assign IPv6 addresses and CIDR ranges
+3. Setup of a pipeline that exercises CF Deployment with IPv6 enabled within a `bbl` environment.
 
 
 ## Other Topics
--- a/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
+++ b/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
@@ -90,6 +90,39 @@ The “first” CIDR (2001:db8:4e04:D139:0::/80) is then reserved for “interna
 * `2001:db8:4e04:D139:0002:0000:0000:0000/80`
 * …
 
+Another option would be to define the prefix size in the deployment config E.g. in the networks section.
+
+The cloud config would look like this:
+
+```yaml
+networks:
+  - name: diego_cells_ipv6
+    subnets:
+      - az: z1
+        cloud_properties:
+          security_groups:
+            - sg-abcd...
+          subnet: subnet-12345
+        dns:
+          - 2001:db8:4e04:D139:0000:0000:0000:0253
+        gateway: 2001:db8:4e04:D139:0000:0000:0000:0001
+        range: 2001:db8:4e04:D139:0000:0000:0000:0000/64
+        reserved:
+          - 2001:db8:4e04:D139:0000:0000:0000:0000/80
+```
+
+and in the deployment config we could refer to this network and define the desired ipv6 prefix size there:
+
+```yaml
+instance_groups:
+  name: diego_cell
+  networks:
+    - name: diego_cells_ipv6
+      # assign non-overlapping `/80` cidrs from `range`
+      ipv6_prefix_delegation_size: 80
+
+```
+
 #### BOSH Cloud Provider Interfaces (CPIs)
 
 The Cloud Provider Interfaces abstract the underlying infrastructure and apply the logical configuration defined in BOSH to concrete infrastructure elements. Accordingly, the CPIs must be aware and capable of assigning and managing IPv6 addresses for BOSH managed networks.
--- a/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
+++ b/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
@@ -1,7 +1,7 @@
 # Meta
 
-- Name: IPv6 Dual Stack Support for     Cloud Foundry
-- Start Date: 2025-01-22
+- Name: IPv6 Dual Stack Support for Cloud Foundry
+- Start Date: 2025-02-07
 - Author(s): @peanball, @a-hassanin, @fmoehler, @dimitardimitrov13, @plamen-bardarov
 - Status: Draft
 - RFC Pull Request: (fill in with PR link after you submit it)
@@ -10,9 +10,9 @@
 
 IPv6 becomes increasingly prevalent on the internet and slowly also in the networks of large customers.
 
-This RFC proposes to add IPv6 support to various Cloud Foundry components in addition to the existing networking in CF to support IPv4, IPv6, or Dual Stack, i.e. IPv4 *and* IPv6 at the same time. The latter is particularly relevant for basic extension and migration of existing CF foundations.
+This RFC proposes to add IPv6 support to various Cloud Foundry components in addition to the existing networking in CF that is based on IPv6. The goal is to support Dual Stack, i.e. IPv4 *and* IPv6 at the same time. This support is particularly relevant for basic extension and migration of existing CF foundations that are based on IPv4 and allow them to also communicate via IPv6.
 
-Furthermore, the goal is to support IPv6 for both ingress and egress traffic, ensuring that Cloud Foundry can handle inbound and outbound network communication using IPv6, further enhancing compatibility and future-proofing the platform.
+Furthermore, the goal is to add support IPv6 (i.e. allow Dual Stack) for both ingress and egress traffic, ensuring that Cloud Foundry can handle inbound and outbound network communication using IPv4 and IPv6, further enhancing compatibility and future-proofing the platform.
 
 ## Problem
 
@@ -35,7 +35,7 @@ The goal is to use IPv6 best practices and benefits to the full extent with the 
 3. Support of dual stack networking, i.e. IPv4 and IPv6 at the same time. This is primarily for a smoother migration because the IPv6 changes can then be additive.
 4. IPv6 is additive to IPv4 functionality
 
-The ultimate future goal would be to move to pure IPv6 communication but is not a current priority and is not covered in this proposal.
+The ultimate future goal would be to allow a configuration based on pure IPv6 communication, i.e. to allow disabling IPv4 entirely. This is not a current priority and is not covered in this proposal. Options to still connect to IPv4 from an IPv6-only cloud foundry foundation at the edges would be via NAT64 and similar technologies.
 
 Backward compatibility is an important topic. Any changes that would change the default behavior should be placed behind appropriate configuration flags and turned off by default. The absence of IPv6 specific configuration must also remain without side effects to existing CF foundations.
 
@@ -136,7 +136,7 @@ This single machine, i.e. a Diego cell, then has sole control over this network 
 
 Network traffic from and to application containers needs to be controlled, also for IPv6:
 
-1. For inter-container communication, Application Security Groups (ASGs) need to support IPv6.
+1. For inter-container communication, Application Security Groups (ASGs) need to add support for IPv6.
 2. When publicly routable addresses (Global Unicast Addresses (GUAs) in IPv6) are assigned to each application container, traffic from outside the foundation needs to be blocked.
 
 BBS, Executor and Garden components need to be extended to support the ICMPv6 protocol for ASGs.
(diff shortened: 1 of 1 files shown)


toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
```diff
@@ -0,0 +1,200 @@
...
+## Proposal
+
+The goal is to use IPv6 best practices and benefits to the full extent with the following high-level goals:
+
+1. NAT and port forwarding are avoided wherever possible when using IPv6
+2. Application workloads have their own IPv6 address. This allows better correlation between requests and their origin. It is important to note that application workloads shall not be addressable directly from external networks. The ingress and load balancing via Gorouter must remain.
+3. Support of dual stack networking, i.e. IPv4 and IPv6 at the same time. This is primarily for a smoother migration because the IPv6 changes can then be additive.
+4. IPv6 is additive to IPv4 functionality
+
+The ultimate future goal would be to move to pure IPv6 communication but is not a current priority and is not covered in this proposal.
```
rkoster, line 38: IPv4?

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
```diff
@@ -0,0 +1,200 @@
...
+
+#### TCP-Router and Routing API
+
+routing-api and tcp-router need to be verified, and if necessary extended, to support IPv6 addresses in addition to IPv4 addresses.
+
+## Other Topics
+
+### Windows Support
+
+Windows support is currently not in the scope of this RFC.
```
ameowlia, line 200: I think windows should be part of this RFC. If we are going to add IPv6 support we should add it for all apps not just linux.

ameowlia:
Hi @peanball, Thanks for putting this together. 

I know there is a lot here, and a lot in App Runtime Platform's domain. My main thoughts are:

* This is going to touch a lot of things. I sure we will learn about more things as we go.
* I agree with your comment that this should be backwards compatible. All of the defaults for all new properties should default to the "old way".
* I think windows should be in scope, unless there is a good reason not to include it. We maintain two OSes we should support both.

peanball:
@ameowlia, thanks for the feedback. 

Regarding windows: we will need support here in describing what needs to be touched as we (SAP) are not using Windows as target.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: IPv6 networks allow interoperability, e.g. via NAT64 and similar. IPv6-only is still intended as "option" on how to use Cloud Foundry. We will clarify this.

jochenehret:
Hi @peanball ,
to adopt IPv6 in cf-deployment and other projects, we will need an update of **bosh-bootloader**. The tool must be able to configure AAAA records for the load balancer ingress addresses. Not sure what changes are needed for the outgoing traffic. The cloud-configs must be updated with the new CIDRs. There are certainly more changes necessary, is anyone already working on that?

For **cf-deployment**, we will have to update our test bbl environments when bosh-bootloader is ready. The "staging" and "running" security groups must be updated, but that should be moderate effort. There will be certainly more changes necessary...

Best Regards,

  Jochen.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Overview
+
+The largest conceptual change is providing individual addressing for application instances. Diego is a container runtime that is closely integrated with various other components. By extending the addressing schema to per-container addresses, all components that touch application workload information need to be adjusted.
+
+The baseline dual stack support for networks deployed via BOSH is another pillar of the proposed change.
+
+The following sections outline the changes necessary for various core components of  a typical CF foundation.
+
+### BOSH
```
mariash, line 50: Doesn't BOSH support ipv6 already if CPI supports it? - https://bosh.io/docs/guide-ipv6-on-vsphere/

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
```diff
@@ -0,0 +1,200 @@
...
+BOSH stemcells up to Ubuntu Jammy Jellyfish have IPv6 support disabled at kernel level. While it can be enabled, the resulting configuration change requires a VM reboot.
+
+The goal is to keep the already enabled IPv6 support from Ubuntu upstream enabled.
+This will be the case starting with Ubuntu Noble stemcell. Enabling IPv6 in older stemcell releases such as “Ubuntu Jammy” might be seen as backwards incompatible change.
+
+This topic is being discussed and tracked via the [issue os-conf-release#71](https://github.com/cloudfoundry/os-conf-release/issues/71) and must be followed up with appropriate fixes.
+
+### Diego
+
+Currently, the Silk CNI is used to create an overlay network, but also to assign IP addresses for the application containers. This assignment is done from a centrally managed pool of IP addresses.
```
mariash, line 129: Was Silk CNI work specifically not included in this RFC?

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
```diff
@@ -0,0 +1,200 @@
...
+
+The delegation of IPv6 addresses within the cells is managed through the host-local CNI plugin. Each cell is assigned an IPv6 prefix at the configuration level, which is then provided to the host-local IPAM plugin. Using this prefix, the plugin allocates unique IPv6 addresses to individual containers within the cell. To prevent address conflicts and ensure proper tracking, a state file is maintained at the cell level, keeping a record of all assigned addresses.
+
+This single machine, i.e. a Diego cell, then has sole control over this network range and can use it to assign specific IP addresses to its containers. At a network level, traffic is still sent to the Diego cell, and within the Diego cell can be forwarded to the application container based on its IP address directly as part of the networking stack.
+
+Network traffic from and to application containers needs to be controlled, also for IPv6:
+
+1. For inter-container communication, Application Security Groups (ASGs) need to support IPv6.
```
mariash, line 139: I think majority of work would be to support dual-stack. For CNI wrapper plugin which configures ASGs IP address is an arbitrary string.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
```diff
@@ -0,0 +1,200 @@
...
+
+1. IPv6 support for configuring application security groups
+2. …
+
+### Networking
+
+Network Policies that are stored in [the policy-server-api which is part of cf-networking-release](https://github.com/cloudfoundry/cf-networking-release/blob/develop/docs/08-policy-server-api.md).
+
+Network policies may reference IP address ranges as destination. Destinations are stored as strings. The underlying IP ranges are also [strings with a start and end address](https://github.com/cloudfoundry/cf-networking-release/blob/aa25a3f96dc21e8a3e25db10f741e4f678c78464/src/code.cloudfoundry.org/policy-server/api/api.go#L37-L49). The data is ultimately serialized JSON that is stored in a database column. As IPv6 addresses are longer than IPv4 addresses, we might need to bump the table size.
+
```
mariash, line 177: Same here with the dual-stack. If we want to roll this out for existing deployments dual-stack should be supported for backwards compatibility. 

Most of the components only set up IP address right now.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
```diff
@@ -0,0 +1,200 @@
...
+Gorouter is largely based on the Golang implementation for HTTP request processing and routing. This implementation is capable of IPv6 address resolution and communication. Gorouter’s route registry and its endpoint handling are also well contained
+
+For the scope of this RFC, testing and fixes for small unforeseen issues are considered, but no direct work is expected for supporting IPv6 targets.
+
+#### TCP-Router and Routing API
+
+routing-api and tcp-router need to be verified, and if necessary extended, to support IPv6 addresses in addition to IPv4 addresses.
+
+## Other Topics
+
```
mariash, line 197: I think another thing probably to mention is that all components that have API (Cloud Controller, Policy Server, UAA, Credhub, BBS, Rep, Mysql) should be able to accept IPv6 connections. It is probably already managed by the http modules they use, unless some explicitly listen on 127.0.0.1.

mariash: Also, BOSH DNS should resolve IPv6 addresses depending on the client protocol.

mariash: Another thing, CATS should have end to end tests for IPv6 (ASG rules, App listening on IPv6, etc)

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: IPv6 yes, dual stack not yet.

peanball:
Jochen wrote:
>to adopt IPv6 in cf-deployment and other projects, we will need an update of bosh-bootloader. The tool must be able to configure AAAA records for the load balancer ingress addresses. Not sure what changes are needed for the outgoing traffic. The cloud-configs must be updated with the new CIDRs. There are certainly more changes necessary, is anyone already working on that?

@fmoehler, could you support me in extend on that?

ameowlia:
> @ameowlia, thanks for the feedback.
> 
> Regarding windows: we will need support here in describing what needs to be touched as we (the current authors of this RFC) are not using Windows as target.

I've reached out to @ebroberson, our resident Windows expert, to comment more about what changes will be needed for windows.

stephanme:
For cf-deployment I think it makes sense to define an (experimental) ops-file for enabling IPv6 early so that the different components have an integration point for their settings.

Probably it makes sense to add a validation pipeline for IPv6 as soon as bbl and bosh supports it. Similar to the noble validation pipeline.

ebroberson:
Windows enables IPv6 by default as of Windows Server 2008, so I don't think we'll need anything specific there.

beyhan:
> Windows enables IPv6 by default as of Windows Server 2008, so I don't think we'll need anything specific there.

@ebroberson what about the following parts:
https://github.com/sap-contributions/cf-community/blob/639791ad22e3d82ac3f861ad4a9741bde0295ce9/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md?plain=1#L101-L103 and https://github.com/sap-contributions/cf-community/blob/rfc-ipv6-in-cf/toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md?plain=1#L55?

ebroberson:
@beyhan I was referring to anything _else_ that might be Windows-specific on the stemcells / vms.  Windows itself already supports dual-stack with IPv6, so I don't believe we'll need anything further other than what the RFC already specifies.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: Please also note the extended section on Silk that was added by @plamen-bardarov

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: I've changed the wording to "need to add support for IPv6", which implies dual stack. Everything in this RFC is about adding IPv6 to the existing IPv4 and thus reaching dual-stack support.

Do you think we need to be more explicit here?

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: I will add some sentence in the beginning that highlights that this is about adding IPv6 and thus reaching dual stack.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: Based on the comment by @erobertson, it seems like there is no particular specific requirement / deviation for Windows.

We will still require support with the implementation and testing of IPv6 / dual stack related changes on Windows.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: I've added the point about "listeners" generally being IPv6 / dual stack capable, also including BOSH DNS. 

CATs and bbl (thanks @jochenehret) are now mentioned as well.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: @erobertson, I've updated the statement. Could you please cross check that what is written now is correct?

peanball: @ebrobertson, I've updated the statement. Could you please cross check that what is written now is correct?

peanball:
@stephanme, @rkoster, @ameowlia, @ebroberson, @jochenehret, @mariash, @beyhan, I have addressed the feedback provided so far. Thank you for your feedback!

We might need to write a bit more about the bosh-bootstrap / bbl environments. Help here would be appreciated, as I've not worked with bbl environments or bosh-bootstrap yet.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: Modified that statement to a more realistic goal.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
ebroberson: Looks good to me!

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
```diff
@@ -0,0 +1,269 @@
...
+The IP address range configuration and prefix delegation are configured via cloud config, e.g. as follows:
+
+```yaml
+networks:
+  - name: diego_cells_ipv6
+    subnets:
+      - az: z1
+        cloud_properties:
+          # assign non-overlapping `/80` cidrs from `range`
+          ipv6_prefix_delegation_size: 80
```
jpalermo, line 73: cloud_properties are generally reserved for things that are NOT bosh features. A feature intended for only a single CPI to be able to enable something specific in the particular IaaS.

I'm don't have an accurate view of all the pieces we're dealing with here, but if it's a general purpose ipv6 thing, it should be part of the core network properties, not down in cloud_properties.

It may be possible to add new upper level properties without breaking existing CPI APIs, but if not, it's not a huge deal to decide to bump the CPI API version from 2 to 3. We've only done it a single time, so it sounds like a big deal, but really all the CPIs declare their version and the Bosh director asks them which version they support for every call. So implementing new CPI version 3 things isn't a big deal if needed.

peanball: Hi @jpalermo thanks for your feedback!

This is a general feature for IPv6. Prefix delegation is a basic feature of the  general IPv6 networking stacks. While not all hyperscalers currently support prefix delegation on their VMs, most already do.

This is a fundamental feature for enabling IPv6 address delegation, i.e. "IPv6-address per Diego LRP instance container".

CPI version 3 sounds reasonable for adding dual stack support and the various IPv6 bits.

I'm not an expert on BOSH's core network properties.

Would it be possible that you either point me to the right place, or adapt the above example to something that would fit into the core network properties?

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
```diff
@@ -0,0 +1,269 @@
...
+The baseline dual stack support for networks deployed via BOSH is another pillar of the proposed change.
+
+The following sections outline the changes necessary for various core components of  a typical CF foundation.
+
+### BOSH
+
+BOSH is used to provision and configure the VMs that run the components of Cloud Foundry. The following features are needed to support the goals of this proposal:
+
+1. Assignment of IPv6 addresses in addition to IPv4, ideally to the same NIC
+2. Assignment of IPv6 CIDR ranges (prefix delegation), e.g. assign a `/80` address, while avoiding overlapping. This is relevant for the use case of address delegation to application workloads hosted on Diego cells.
```
maxmoehl, line 55: If we can assign prefixes, assigning a `/128` prefix should be equivalent to assigning an IPv6 address (but I'm not in the details, there might be some subtle differences) so maybe we "only" need prefix assignment?

Regarding prefixes for diego: there might be a benefit to using different prefixes for the cell itself and app containers. That way you can define ACLs for all apps / cells via a single prefix. This would require BOSH to be capable of assigning multiple IPv6 prefixes to a single VM.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
beyhan: This has been addressed with https://github.com/cloudfoundry/community/pull/1077/commits/0eec79e146c4c5058a4abd791aaaea3e3cac0535. @fmoehler could you please confirm.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
```diff
@@ -0,0 +1,269 @@
...
+networks:
+  - name: diego_cells_ipv6
+    subnets:
+      - az: z1
+        cloud_properties:
+          security_groups:
+            - sg-abcd...
+          subnet: subnet-12345
+        # assign non-overlapping `/80` cidrs from `range`
+        ipv6_prefix_delegation_size: 80
```
rkoster, line 76: should this be specified at the cloud-config level, or at the deployment level? Is there a use-case for mixing prefixes? As in diego-cells get a /80 but cloud-controller won't need such a prefix.

mariash (approved):

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: Yes. Absolutely.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
a-hassanin: I would separate IP Address assignment and Prefix assignment here 'to not force' a consumer to specify a prefix. At the end, if an IPv6 address is specified without a prefix, the IaaS Provider; like AWS; can always apply the default.

Regarding assigning multiple IPv6 prefixes on a single VM, after the discussion with @peanball, it seems it is not required to be in scope for this RFC, but indeed, it is something we can add later on if required.

peanball: The "not required" statement is not set in stone. We are trying to figure out with this RFC, how to best do it.

I agree that assigning prefixes (which under the hood you always do anyway!) would be good. IF needed, we could have a "lazy" mode where you define just the address and not a CIDR, and it automatically assigns as `/128`. But those are finer implementation details.

The general suggestion, which I find interesting and worth including, is that we could have distinct address ranges for the Diego cells themselves, and for the container workloads on each Diego cell.

This actually works into @rkoster's question about the prefix delegation size, which should be more dynamic as it is at the moment. 

It's really only Diego at the moment that requires prefix delegation, as it does something with those delegated addresses.
Other deployments require a single address and are good. She the "delegation prefix" is then a single address for all deployments that are not Diego. And that for Diego we would have one with a single address, and one with a larger delegated prefix.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: I've now responded to a comment where the topic is related: https://github.com/cloudfoundry/community/pull/1077#discussion_r1982018440

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
maxmoehl: Multi-address allows for quite some interesting setups. For example, when combined with bpm you could have separate prefixes for all metric/logging/... components which other components cannot talk to (e.g. on a gorouter VM, rsyslog would get a "logging" IP, but gorouter itself would not). This would reduce the impact of a compromised service as it can only talk to its peers and not just everyone. It's questionable whether you would actually do something like this in our context as it is quite some effort to get right.

plamen-bardarov: > The general suggestion, which I find interesting and worth including, is that we could have distinct address ranges for the Diego cells themselves, and for the container workloads on each Diego cell.

From the diego/silk perspective, having a dedicated IPv6 prefix per cell exclusively for application containers is beneficial as it prevents conflicts when assigning IPv6 addresses, since the host-local IPAM plugin only checks its local state files and does not verify existing addresses on network interfaces. It also enhances network separation, making it easier to define ACLs if needed for apps/cells

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
fmoehler: i put in a small snippet adding this as another option: https://github.com/cloudfoundry/community/pull/1077/commits/4b8f22cf9d8b0bd37ffef41d5332df7272684642

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
beyhan: My personal suggestion would be to focus on the minimal scope which implements the desired support for IPv6. Multi-address is also possible with IPv4 and IIRC we tried similar things as described in @maxmoehl [comment](https://github.com/cloudfoundry/community/pull/1077#discussion_r1982732684) but had many issues with that and decided against it. It could be that with IPv6 it is easier to get this setup work but It looks to me as topic for the possible future improvement section of this RFC. 

If I got the discussions right another open point is to update the RFC with mentioning support for distinct address ranges for the Diego cells themselves, and for the container workloads on each Diego cell. 

@peanball we didn't start the FCP because of this open discussion. If this resolves until Tuesday next week most probably we can start the FCP.

PlamenDoychev: I do agree with the above statement from @beyhan, if not documented we should document that we do support distinct address ranges for the Diego cells themselves, and for the container workloads.

May-be it is also worth mentioning that multi-address is out of scope for this iteration.

peanball: I'm adding a sentence that a single CIDR will will be assigned. This is also in line with the BOSH config sample that defines a single prefix delegation size and subnet. Anything more elaborate will need a different configuration mechanism.

None of this precludes us to extend in the future however.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
peanball: I'm adding emphasis on this 2nd option, pointing it out as preference. Also I've added a warning that we need to be careful when the prefix size changes.

beyhan:
We discussed this RFC during the TOC meeting on 11th of March and decided to start the FCP with the goal to accept the RFC.

toc/rfc/rfc-draft-ipv6-dual-stack-for-cf.md
```diff
@@ -0,0 +1,312 @@
...
+instance_groups:
+  name: diego_cell
+  networks:
+    - name: diego_cells_ipv6
+      # assign non-overlapping `/80` cidrs from `range`
+      ipv6_prefix_delegation_size: 80
+
+```
+
+The preference is to use the deployment-based expression for the prefix size, as it provides better control to the deployment and does not require changes to the cloud config.
```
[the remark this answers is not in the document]
peanball: It's not the network range but the "step size" for divvying up the range so the smaller segments can be assigned to the resulting VMs in the deployment.

tl;dr: The deployment defines how many instances there will be that need to fit into the overall CIDR provided to the respective BOSH network. This informs the "step size" that can be used to ensure non-overlapping addresses.

The deployment should be agnostic to the environment, but only up to a point. Depending on your environment, you might have a larger or smaller IPv6 CIDR available to give to the overall Diego cell network and to subdivide.

Say you want to lay out your network in the `2001:db8:4e04:D139:0000:0000:0000:0000/96` CIDR.

If you want to maximize the address space for 15 Diego cells, you then might want to assign to each Diego cell a `/100` CIDR from the `/96` Diego network CIDR:
* `2001:db8:4e04:D139:0000:0000:0000:0000/100` (according to our examples: reserved for the deployment, DNS, etc.)
* `2001:db8:4e04:D139:0000:0000:1000:0000/100` - cell 1
* `2001:db8:4e04:D139:0000:0000:2000:0000/100` - cell 2
* `2001:db8:4e04:D139:0000:0000:3000:0000/100` - cell 3
* `2001:db8:4e04:D139:0000:0000:4000:0000/100` - cell 4
* `2001:db8:4e04:D139:0000:0000:5000:0000/100` - cell 5
* `2001:db8:4e04:D139:0000:0000:6000:0000/100` - cell 6
* `2001:db8:4e04:D139:0000:0000:7000:0000/100` - cell 7
* `2001:db8:4e04:D139:0000:0000:8000:0000/100` - cell 8
* `2001:db8:4e04:D139:0000:0000:9000:0000/100` - cell 9
* `2001:db8:4e04:D139:0000:0000:a000:0000/100` - cell 10
* `2001:db8:4e04:D139:0000:0000:b000:0000/100` - cell 11
* `2001:db8:4e04:D139:0000:0000:c000:0000/100` - cell 12
* `2001:db8:4e04:D139:0000:0000:d000:0000/100` - cell 13
* `2001:db8:4e04:D139:0000:0000:e000:0000/100` - cell 14
* `2001:db8:4e04:D139:0000:0000:f000:0000/100` - cell 15

Each cell will get ≈268 million IPv6 addresses from their `/100` CIDR range.

But if you decide that you want to have 350 Diego cells, you would pick a smaller "step" of e.g. `/108` that would give each cell ≈1 million and could fit 4096 cells.


But if your initial CIDR for the Diego network is a `/64`, you have other considerations. Because I don't want to prescribe a size for Diego, or for the minimum CIDR to give to Diego, it might be best for the deployment to decide how bit a slice to expect from the overall CIDR for each of the VMs.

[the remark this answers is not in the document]
rkoster: Re-reading the full paragraph, and it all makes sense. Thanks for the clarification.

stephanme (approved):

rkoster (approved):

ChrisMcGowan (approved):

PR #1078, opened 2025-02-10
Move OSBAPI to ARI WG - Proposal

pivotal-marcela-campo:
As discussed in latest Service Management WG update, the service group has little activity and it would be worth discussing options to relocate the projects to other working groups.

This PR is to move OSB API maintainance to ARI

Why?

- Changes to OSB API don't make sense if brokers and platforms don't implement them.
- CAPI has not kept up to date with OSB API, and hence doesn't make sense to do more changes to it apart from minor bugfixes.
- There's currently not much activity in OSB API, so not much overhead
- CAPI has more context on what features they would be willing to implement/accept PRs for than the service broker authors. Hence ideally any proposed change should be anyway be discussed with CAPI for feedback before adding it to the API
- In the past, CAPI members maintaining the services apis were involved in evolving OSBAPI
- Changes to the API have historically required buying in from platforms and a proof of concept implementation in the platform, as a means of viability. ARI is in a better place to provide that than service broker authors, being the maintainers of CAPI.

Approvers:

- There are currently no active approvers in the OSBAPI.
- There is still at least one member of the old OSB API project within ARI (Sam Gunaratne) that could at least help the WG ramp up on it.
Note: I have added one approver as otherwise the automation would fail, but there are no active approvers at the moment. I didn't want to make any assumptions on who would be in that position within ARI.


beyhan:
We will wait until the ARI WG discuss this internally. 

@Gerg could you please put thumps up if you are OK with this.

rkoster (approved):

Gerg:
We will discuss in next ARI WG meeting

stephanme (approved):

ChrisMcGowan (approved):

beyhan (approved):

ameowlia (approved):

PR #1085, opened 2025-02-12
RFC: Password Deprecate Request

strehle:
Short:  Deprecate Password Grant in UAA. 
Consequence is: Build an alternative in cf login / cf auth because until now the password grant is used always in case of CF user authentications. UAA supports JWT bearer, but this should be improved in order to consume it and to not block customers, because of missing functionality.

Most of technical details in https://github.com/cloudfoundry/uaa/issues/3285

@cloudfoundry/toc 

[For easier viewing](https://github.com/strehle/community/blob/rfc-deprecate-passwords/toc/rfc/rfc-draft-deprecate-passwords.md)


beyhan:
@a-b could you please provide feedback if you have any on this RFC because it also proposes CF CLI changes.

a-b:
Thank you for bringing this to our attention. We support improvements in security as long as CF CLI remains backward compatible with older deployments.

beyhan:
@a-b do you have any commends on the proposed phased approach in https://github.com/strehle/community/blob/1494e6b766f3b7dcacf76bd8eacd14e75097307d/toc/rfc/rfc-draft-deprecate-passwords.md?plain=1#L64-L79?

a-b:
I am curious about the expiration and rotation of tokens. Should we implement a policy that requires regular rotation by default?

ameowlia (approved):

ChrisMcGowan (approved):

beyhan (approved):

PR #1100, opened 2025-03-13
Add Ahmed to Stemcell Release Engineering (BOSH)

a-hassanin:
Somehow I was missing from there


jpalermo (approved):

beyhan (approved):

beyhan:
For the "Stemcell Release Engineering (BOSH)" I think we don't need to execute the [contributions-for-user.sh](https://github.com/cloudfoundry/community/blob/main/toc/working-groups/contributions-for-user.sh) because the area is only for CI access.

rkoster (approved):

PR #1139, opened 2025-05-05
Support multiple orgs in inactive user deletion

beyhan:
With the [change](https://github.com/cloudfoundry/community/commit/208be3b2ebb453d2e827fa3bc25b9aef54f3849e) we introduced multiple org support to [contributors.yml](https://github.com/cloudfoundry/community/blob/main/orgs/contributors.yml). This change adapts the inactive user deletion to work with the new structure of [contributors.yml](https://github.com/cloudfoundry/community/blob/main/orgs/contributors.yml).


rkoster (approved):

beyhan:
Merging the change with only one approval because I would like to fix the failing automation.

PR #1151, opened 2025-05-09
Add ssunka to app runtime platform logging and metrics

ssunka:
Will be involved in active contribution for Logging and Metrics area moving forward.
Contribution so far for metric-store -> https://gist.github.com/ssunka/3c44a44076fc1105f67921dbc67fd6a4

### DIFF
--- a/orgs/contributors.yml
+++ b/orgs/contributors.yml
@@ -193,6 +193,7 @@ orgs:
     - vkalapov
     - VRBogdanov
     - wayneeseguin
+    - weili-broadcom
     - WeiQuan0605
     - weresch
     - winkingturtle-vmw


beyhan:
@cloudfoundry/wg-app-runtime-platform-logging-and-metrics-approvers we will need two approvals from the approvers in the "Logging and Metrics" area because it has more than 4 active approvers.

ameowlia:
@ssunka, in order to become an approver you must have 20 "substantial" contributions ([see policy here](https://github.com/cloudfoundry/community/blob/main/toc/ROLES.md#approver)).

❓ Can you please prune your gist to only include substantial contributions?

PR #1154, opened 2025-05-09
Add Stuart as approver

anita-flegg:
backup for Anita


beyhan:
@pspinrad, @snneji Would one of you please review and approve this? Then @anita-flegg can merge. The area is small that is why we need only one approval.

anita-flegg:
@beyhan , Yes, @snneji and @psinrad have both left, and should no longer have access.
@stuclem is my new backup.

Thanks!

PR #1157, opened 2025-05-14
Add repo for `bosh-common`

aramprice:
This is part of the work to extract this gem from cloudfoundry/bosh

See: https://github.com/cloudfoundry/bosh/issues/2612

### DIFF
--- a/orgs/orgs.yml
+++ b/orgs/orgs.yml
@@ -218,6 +218,9 @@ orgs:
         description: BOSH CLI v2+
         has_projects: false
         has_wiki: false
+      bosh-common:
+        default_branch: main
+        has_projects: true
       bosh-compile-action:
         default_branch: main
         has_projects: true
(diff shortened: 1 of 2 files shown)


aramprice:
There wasn't any obvious prior art so I'm not sure if this is all that's required for this change.

beyhan (changes requested):
Hi @aramprice,

the repository should be also added to the https://github.com/cloudfoundry/community/blob/main/orgs/orgs.yml like for example done in https://github.com/cloudfoundry/community/pull/1140.

aramprice:
@beyhan - thanks, updated (hopefully correctly)

orgs/orgs.yml
```diff
@@ -218,6 +218,9 @@ orgs:
         description: BOSH CLI v2+
         has_projects: false
         has_wiki: false
+      bosh-common:
+        default_branch: main
+        has_projects: true
```
rkoster, line 223: why do you need projects?

stephanme (approved):

ChrisMcGowan (approved):

beyhan (approved):

rkoster (approved):

orgs/orgs.yml
aramprice: Because I copy-pasted from other sections. Happy to remove (or have someone remove it)

beyhan: All good. I will check on it and correct if need.

PR #1162, opened 2025-05-20
Inactive users to be removed from ARI WG

stephanme:
According to the rules for inactivity defined in [RFC-0025](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0025-define-criteria-and-removal-process-for-inactive-members.md), the following users should be removed from ARI approvers/reviewers/bots:
- @dsboulder
- @rsampaio
- @calebwashburn
- @ccjaimes
- @danielfor
- @dalvarado
- @reedr3

According to the [revocation policy in the RFC](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0025-define-criteria-and-removal-process-for-inactive-members.md#remove-the-membership-to-the-cloud-foundry-github-organization), users have two weeks to refute this revocation, if they wish by commenting on this pull-request and open a new pull-request to be re-added as approver after this one is merged.

Inactivity was detected by a github action, see #1141

PR #1172, opened 2025-05-28
ARI: merge OSB API area into CAPI

stephanme:
- OSB API area got orphaned when the last approver was removed by #1162 (inactive user cleanup)
- decision in ARI WG meeting on 25-05-27 to add the repos to the CAPI area

### DIFF
--- a/toc/working-groups/app-runtime-interfaces.md
+++ b/toc/working-groups/app-runtime-interfaces.md
@@ -299,12 +299,6 @@ areas:
   - cloudfoundry/staticfile-buildpack
   - cloudfoundry/staticfile-buildpack-release
 
-- name: OSB API
-  approvers: []
-  repositories:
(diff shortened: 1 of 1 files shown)


stephanme:
In the end, this is just documentation as the repos are not located in `cloudfoundry` github org. Once merged, the access rights in github org [openservicebrokerapi](https://github.com/openservicebrokerapi) should be consolidated.

We should consider moving the repos into the `cloudfoundry` org to benefit from the org automation. [rfc-0036-multiple-github-orgs.md](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0036-multiple-github-orgs.md) requires that a WG contains only repos from one CFF github org.
On the other hand, https://www.openservicebrokerapi.org/ aims to define a platform independent service integration which may justify an own github org.

Gerg (approved):

PR #1214, opened 2025-06-20
feat: Remove the ruby call and adjust the gist file url

ZPascal:
### PR Summary
This pull request introduces the following improvements:

#### Removal of Ruby Dependency
- Replaces the Ruby-based YAML-to-JSON conversion in with the modern [yq](https://github.com/mikefarah/yq) CLI tool
- The Ruby invocation previously used for parsing YAML is now fully replaced by `yq -o=json .`, reducing external dependencies and increasing maintainability
- - The parsing logic is centralized in a new shell function, and results are collected into a single JSON array for improved downstream usage

#### URL Encoding Adjustment
- In , the URI-safe file naming step has been enhanced. `contributions-for-user.sh`
- After URL-encoding filenames with `jq -sRr '@uri'`, a series of `sed` substitutions is applied to:
    - Replace spaces (`%20` →` `)
    - Remove newlines (`%0A`)
    - Replace slashes ( `%2F`→ `and`)
- This ensures generated files are better suited for the specific needs and readable

#### Additional Notes
- Both scripts now explicitly check for the availability of and before proceeding, ensuring a smoother user experience and clearer error messages. `yq` & `jq`

**In summary:**
This PR modernizes YAML parsing by eliminating the Ruby runtime dependency and improves custom handling of encoded filenames, making the scripts more portable and robust.

### DIFF
--- a/toc/working-groups/contributions-for-user.sh
+++ b/toc/working-groups/contributions-for-user.sh
@@ -29,6 +29,16 @@ usage_checks() {
     exit 1
   fi
 
+  if [[ ! -e $(which yq) ]]; then
+    echo 'The `yq` cli (https://github.com/mikefarah/yq) is required for this script' >&2
+    exit 1
+  fi
+
+  if [[ ! -e $(which jq) ]]; then
+      echo 'The `jq` cli (https://github.com/jqlang/jq) is required for this script' >&2
+      exit 1
+  fi
+
   if gist_url=$(echo "test" | gh gist create -d test -f test - 2>/dev/null); then
     gh gist delete "${gist_url}" --yes >/dev/null 2>&1
   else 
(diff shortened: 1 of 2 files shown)


ZPascal:
@beyhan Can you please check the PR? Thank you!

beyhan:
@geofffranks as initial contributor of this script could you please provide your feedback? I couldn't assign you as reviewer.

rkoster:
Ruby which is already available on MacOS is being replace by a more bespoke dependency. Also there are now 2 really simliar dependencies, as in there is nothing jq can do yq can not. Introducing yq as a dependency only makes sense to me if it would replace ruby and jq.

ZPascal:
> Ruby which is already available on MacOS is being replace by a more bespoke dependency. Also there are now 2 really simliar dependencies, as in there is nothing jq can do yq can not. 

@rkoster I've checked if Ruby is part of the default shipment of Mac OS X, and it is not the case. `Ruby` is available via Homebrew, and `yq` is also available via Homebrew. The difference is that you're installing `Ruby` as a full programming language, and with `yq`, only a binary to handle YAML/JSON. 
**Background:** I'm also using Linux/ Debian by default as a distribution for my device, and I have not installed any Ruby version. I've also faced issues running the script locally because Ruby was not installed, and it was also not documented that Ruby is necessary to run it.

>Introducing yq as a dependency only makes sense to me if it would replace ruby and jq.

I'll check the possible options to adjust the script and remove Ruby and jq completely.

> I agree, lets get rid of jq as well, and just use yq on the json stuff too.

@geofffranks I'll adapt the code base.

>parsable-working-groups.sh is used in [github workflow remove-individual-access](https://github.com/cloudfoundry/community/blob/main/.github/workflows/remove-individual-access.yml).
> Need to check if yq is part of the used ubuntu image and install it if not.

@stephanme Thank you for the information. I'll adapt the GH action.

stephanme (approved):

rkoster (approved):

PR #1216, opened 2025-06-21
Archive bosh-windows-acceptance-tests

aramprice:
These have been merged into bosh-windows-stemcell-builder


beyhan (changes requested):
We should also delete the repo from the FI WG repo list: https://github.com/cloudfoundry/community/blob/main/toc/working-groups/foundational-infrastructure.md?plain=1#L406

stephanme (approved):

PR #1219, opened 2025-06-25
Add Bhuvisha as contributor

vipinvkmenon:
Bhuvisha joins the team that maintains the terraform-provider-cloudfoundry

### DIFF
--- a/orgs/contributors.yml
+++ b/orgs/contributors.yml
@@ -22,6 +22,7 @@ orgs:
     - Benjamintf1
     - beyhan
     - bgandon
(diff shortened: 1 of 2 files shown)


stephanme (changes requested):
@Bhuvisha03 

In order to become an approver, you should be a reviewer for at least 3 months and have 20 contributions (issues, slack discussions, reviews etc also count as contributions).
Please see [ROLES.md](https://github.com/cloudfoundry/community/blob/main/toc/ROLES.md#approver).

I suggest that you apply for the reviewer role first.

PR #1220, opened 2025-06-25
RFC-0040 Enhance Stack Handling

FloThinksPi:


rkoster:
Hi @FloThinksPi,

Thank you for drafting RFC-0040! It provides a comprehensive overview of enhancements to stack handling in Cloud Foundry, and I appreciate the effort you’ve put into addressing these critical issues.

However, I feel the RFC could benefit from being split into individual, logically scoped RFCs to improve focus and facilitate discussion. For example:

1. Improved Logical Stack Management: Covering changes to CF API for stack states like deprecated, locked, disabled, and their timestamps.

2. Bring Your Own Stack: Discussing the introduction of user-managed stacks via container image references and the related CF API changes.

3. Stack Release Cycle Alignment: Addressing the proposal to build stacks for every Ubuntu LTS release.

4. Org-Scoped Stack Management: Introducing organization-specific stack visibility and control.

Each topic is significant and detailed enough to warrant its own RFC. Splitting them would allow contributors to dive deeper into specific areas, streamline discussions, and prioritize implementation efforts more effectively.

On a higher level I'm also worried about handing responsibility for stacks in the hands of app developers. One of the great features of CF is that platform operators can fix CVEs in for example OpenSSL with a single bosh deploy.

I understand the desire to fix this, but maybe a compromise could be found in making stack management an admin only feature (using the cf cli). Ideally using a blobstore first approach so it works in airgapped environments.

Last but not least I'm wondering about the roll out mechanisms for handeling custom stack updates, given these changes are not orchestrated through bosh, how do we ensure we don't over we Diego? There needs to be some sort of global max_in_flight setting to control the total number of apps that are being restarted at the same time.

stephanme:
I'm not sure that additional discussion threads help. The [RFC README](https://github.com/cloudfoundry/community/blob/main/toc/rfc/README.md#public-discussion-period) clearly states that the discussion shall happen in the PR.

I share @rkoster's opinion that this RFC is rather big and deserves a split. "Improve logical stack management in CF API"  and "Bring your own stack" are proposals that bring value w/o each other and which can be implemented independently. 

"Provide a stack with every ubuntu LTS" is in my eyes not a good RFC candidate. The idea is clear and has been brought up in the past but in the end it is a question of resources and commitment. Concrete RFCs for the next stack like [rfc-0039-noble-based-cflinuxfs5](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0039-noble-based-cflinuxfs5.md) are more helpful as they indicate commitment by the author and make clear that work will really start.

The RFC README doesn't say anything about the scope/size of an RFC and we have RFC of all sizes (from smaller process related RFC like introducing the reviewer role up to long running CF  API v2 removal and manifest v2). I would strive for not-too-big RFCs so that the corresponding implementation issues can get closed one day.

FloThinksPi:
@rkoster to answer to your comment

>I understand the desire to fix this, but maybe a compromise could be found in making stack management an admin only feature (using the cf cli). Ideally using a blobstore first approach so it works in airgapped environments.

We could extend the feature flag to have 3 modes - `off`, `admin-only` and `on`. 
Also we could allow an admin to set a `docker://URL` stack as system stack in the stacks table. This one can only be done by admins.
So then this feature to consume a stack from the outside could actually be an admin only thing. However not sure how much value this adds for an operator/admin. Note that by default the feature-flag anyway will be disabled so an operator firstly has to decide to provide this function in his foundation deliberately. In an airgapped environment an operator would not enable this feature i guess since its pointless. When they would run their own container registry in that environment they could also just use the usual bosh mechanism to deploy the stack onto the diego cells anyway. 

>On a higher level I'm also worried about handing responsibility for stacks in the hands of app developers. One of the great features of CF is that platform operators can fix CVEs in for example OpenSSL with a single bosh deploy.

I also thought intensively over that, in the end we actually already allowed this when we introduced "bring your own buildpack" to some degree. If you run your own buildpack - or even a system one - you anyway have to regularly restage to consume patches to your buildpack or languages libs. In case your application does not do dependency version pinning at all or does just pin direct dependencies but not transitive ones. E.g. patched log4j libs as an example everyone might know of. Even though CF gives you some things automatically in my experience users didn't know that sometimes they have to restage to consume new libs as they (naively)thought the system takes care if it. For example take the python buildpack - you only get a CVE patched in the python interpreter if you restage! Same goes for the Go Compiler, Ruby or Java JRE.

What may thus be desirable when allowing freedom like custom buildpacks or custom stacks is to explain the boundary conditions very well and make them transparent to the user. Can be maybe added to the RFC/a new one that we make this more transparent than today either in the API with a special flag(e.g. `security-lifecycle-status:` or something like that that shows you a restage is required) visible in `cf apps` or as Documentation.
- System Buildpack ➕  System Stack ➡️ You regularly have to restage to deploy new updates of the buildpack or you apps dependencies(if not pinned)
- Custom Buildpack ➕  System Stack ➡️ You regularly have to bump the buildpack version AND have to restage to deploy new updates to the buildpack or you apps dependencies(if not pinned)
- Custom Buildpack ➕  Custom Stack ➡️ You regularly have to bump the buildpack version AND the Stack version AND have to restage to deploy new updates of the buildpack or you apps dependencies(if not pinned)

Deliberately in the RFC only Custom Stack + Custom buildpack shall be supported so bring your own everything. And firstly the operator has to allow its users with the feature flag to use this feature. Secondly a user has to willingly move away from the system buildpack and system stack - this cannot be an accidental thing.  So if better documented i think if a user wants to deliberately care for his buildpack and stack an operator may be enabled to allow him that. As an operator of a foundation compliance can still be validated with a scan over the CF API programmatically in case the operator is responsible for the apps on the foundation. Or decide to not enable that feature at all in case he as concerns.

>Last but not least I'm wondering about the roll out mechanisms for handeling custom stack updates, given these changes are not orchestrated through bosh, how do we ensure we don't over we Diego? There needs to be some sort of global max_in_flight setting to control the total number of apps that are being restarted at the same time.

Since we run at large scale with the docker feature flag enabled on our Foundations we got quite some experience already. In CF the custom stack proposal from a diego point of view is nothing else than a docker app. Diego is - as outlined in the RFC - unware of lifecycles thats a CF API concept. It just knows LRPs and Tasks with either a baseLayer beeing on the disk or from a container registry see https://github.com/cloudfoundry/bbs/blob/main/docs/031-defining-lrps.md maybe i missed something here but that means we can derive the behaviour of custom stacks for diego from the already known behaviour of docker apps withing CF since from a diego point of view they are identical.
Also the update behaviour of the stack is the same as with lifecycle docker. E.g. if you specify a container with tag `latest` on a app restart you run with the newest stack similar as when we bump the system stack currently.
We simply just allow to combine lifecycle docker and buildpack in the sense that the CF API schedules a staging task with a baseLayer from a container registry. 

Thinking about that now - maybe it makes sense for the feature flag for custom stacks to be only activate-able when `diego_docker` https://docs.cloudfoundry.org/adminguide/docker.html feature flag is also set. Opinions on that ?

toc/rfc/rfc-0040-enhance-stack-handling.md
```diff
@@ -0,0 +1,258 @@
+# Meta
+
+- Name: Provide possibillity to use custom stacks in Cloud Foundry
```
beyhan, line 3: The title of this RFC fits better to the [next RFC](https://github.com/cloudfoundry/community/pull/1251) in this context. I think the titles have been swapped by mistake.

beyhan (changes requested):
Please follow the RFC [draft creation process](https://github.com/cloudfoundry/community/blob/b399bb57e17a73ffc9276af2bf1627b092c08e40/toc/rfc/README.md?plain=1#L9-L12) and change the name of the file to `toc/rfc/rfc-draft-enhance-stack-handling.md` because our automation generates and assigns the RFC number when it is accepted and merged.

toc/rfc/rfc-0040-enhance-stack-handling.md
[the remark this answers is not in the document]
FloThinksPi: corrected

toc/rfc/rfc-0040-enhance-stack-handling.md
```diff
@@ -0,0 +1,258 @@
...
+- It allows cf admins to phase out a stack via multiple small steps and
+  gives both customer and admins a paved road forward to move from one
+  stack to another.
+
+- Better awareness of the deprecation and removal of a stack by
+  printing a deprecation notices and timelines in the app logs
+
+- It deliberately shifts the unavailability of a stack towards
+  lifecycle operations and not actual app downtime - making it a more
+  pleasant experience for CF users and operators alike.
```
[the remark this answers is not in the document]
FloThinksPi: Added an example to make clear what the more pleasant refers to in comparison - a downed app.

FloThinksPi:
For reference is peed a few minutes today to make the adoption in the cloud_controller_ng https://github.com/cloudfoundry/cloud_controller_ng/pull/4475/files its not ready yet, likely has failing tests etc. but gives a quick overview whats about to change with this RFC code wise roughly.

toc/rfc/rfc-0040-enhance-stack-handling.md
```diff
@@ -0,0 +1,258 @@
...
+cadence and Support](https://ubuntu.com/about/release-cycle). With
+Canonical\'s stop of standard security maintenance for ubuntu
+18.04(Bionic Beaver) as a consequence also the CFLinuxFS3 stack could
+not be maintained anymore and was deprecated. A successor CFLinuxFS4 was
+offered based on ubuntu 22.04(Jammy Jellyfish).
+
+Stacks are used for the staging of an application meaning buildpacks run
+on the stack to produce the droplet. This in return means that:
+
+- The buildpack is stack specific meaning it can/will break on an
```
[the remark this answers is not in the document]
stephanme: Agreed, you can't hope that todays buildpacks work on a future stack.

I misunderstood the "buildpack is stack specific". Thought it references the optional `stack` parameter of [POST /v3/buildpacks](https://v3-apidocs.cloudfoundry.org/version/3.198.0/index.html#create-a-buildpack).

toc/rfc/rfc-0040-enhance-stack-handling.md
```diff
@@ -0,0 +1,258 @@
...
+Consequently, this information needs to be used inside the CF API at app
+creation time and staging time to check if a stack CAN be used based on
+the current server time. To be able to be used to create an app object
+the stack MUST be not locked or disabled. To be able to stage an app a
+stack MUST not be disabled. The respective endpoints/controllers in the
+CF API need to be extended with a corresponding check and present
+meaningful error messages in case the conditions to create an app or
+stage an app are not meet. Error messages SHOULD refer to the date times
+and include since when a stack is deprecated, when it got locked or
+disabled to enrich the response to the CF user.
```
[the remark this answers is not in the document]
stephanme: Thanks for the clarification. Let's see what others think about the automatic, time-based stack deprecation.

toc/rfc/rfc-draft-enhance-stack-handling.md
```diff
@@ -0,0 +1,249 @@
...
+
+This model guarantees that stack usage is consistently enforced based on time, and all state transitions are predictable and transparent.
+It also provides a interface for an operator in which no externally timed api call for a stack state change is required.
+
+The CF API SHOULD use the provided timestamps to influence the logging and error messages in the following way:
+
+- It MAY add a log line into the Staging/Restaging logs of apps with a locked stack. It SHOULD produce the deprecation warning, optionally with color support, to underline the importance of a deprecated stack.
+- It MAY add the time since when a stack is deprecated/locked/disabled to the Staging/Restaging logs.
+- It MAY add the time when future state transitions will happen to the Staging/Restaging logs, so when is going to be locked/disabled.
+- The `description` field MAY be used in addition in Staging/Restaging logs to enable CF Admins to include a custom message in the apps staging logs
```
stephanme, line 169: CF CLI should also print out the current stack state and the deprecate/lock/disable_at timestamps.

toc/rfc/rfc-draft-enhance-stack-handling.md
```diff
@@ -0,0 +1,249 @@
...
+put onto the local filesystem of the Diego cell and every app instance
+(Long Running Process(LRP)/Task) start will fail because the runtime did not find the
+preloaded stack in its local filesystem. One thus is not able to stop
+shipping/delivering the outdated, insecure stack anymore without causing
+downtimes to all apps still using it.
+
+### Adoption Timelines
+
+Currently a stack is only shipped ever 4 years skipping one LTS version
+of ubuntu entirely. This creates a situation where the old stack e.g.
```
beyhan, line 87: There is no general agreement in the CF community about the versions of Ubuntu LTS which should be adopted but in general all LTS versions starting from Trusty has ben adopted with exception of Focal which was agreed in [this RFC](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0001-jammy-os.md). That is why, this statement is not true that we adopt Ubuntu LTS versions every four year.

beyhan (changes requested):

toc/rfc/rfc-0040-enhance-stack-handling/current_stack_usage.png
beyhan, line 1: Please also change the folder of this file to rfc-draft-enhance-stack-handling because it is not yet clear which number will be assigned to this RFC and it will be renamed when the number is generated after approval and merge.

beyhan (changes requested):

toc/rfc/rfc-draft-enhance-stack-handling.md
```diff
@@ -0,0 +1,249 @@
...
+- deprecate_at -> Timestamp after which the deprecation notice will be
+  printed
+
+- lock_at -> Timestamp after which staging new apps with a that stack
+  fail
+
+- disable_at -> Timestamp after which restaging existing apps with a that
+  stack fail
+
+Stack lifecycle management is controlled exclusively by a set of timestamps, not by an explicit "state" field. The system determines the current state of a stack (active, deprecated, locked, or disabled) by comparing the current server time to these timestamps. This approach ensures that stack usability is automatically and unambiguously defined by time-based transitions. Additional information can already be added via the `description` field.
```
Gerg, line 153: I'm not sure this level of automation/complexity is needed for something relatively rare (stack deprecations). A simpler implementation would be operator-configurable `state` and `deprecation_message` fields.

Gerg: Operators could implement their own automated deprecation flows on top of the CF APIs, of course, CF just wouldn't be opinionated about how to do it.

toc/rfc/rfc-draft-enhance-stack-handling.md
FloThinksPi: Removed that section 👍🏼

toc/rfc/rfc-draft-enhance-stack-handling.md
```diff
@@ -0,0 +1,249 @@
...
+combination is supported. It's not used to check what's actually
+available on diego though.
+
+In this proposal we extend the global stack management and table by
+following workflows inside the CF API.
+
+- Mark a stack as deprecated -> staging and start process logs a
+  prominent warning. This brings better awareness than release notes.
+
+- Mark a stack as locked -> prevent using the stack for new apps
```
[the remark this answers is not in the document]
FloThinksPi: I cannot share concrete numbers but from a representable sample count, an extremely impressive total app count after a few months of switching the default stack a small single percent number was using the new default stack. Extrapolating that would lead to being able to remove an old stack in the range of multiple decades i would say if i remember correctly.
We thus actively moved applications from one stack to another via changing stack - restaging zero downtime - checking if the app comes up again if not reverting the stack back. This moved the majority of apps to a new stack successfully but the leftover - still after multiple years a quite substantial number of apps remain on FS3. 

I see this from the angle of a PaaS provider in which we have limited control over the workloads but need tools to gain influence on the workloads itself and create incentives to use certain usage patterns rather than others. These tools would also be of value for a small company hosting a instance just for themselfs or via a on premises offering by a vendor. Thats why this ADR goes hand in hand with https://github.com/cloudfoundry/community/pull/1251 as only combined they offer a better user and operator experience.

toc/rfc/rfc-draft-enhance-stack-handling.md
```diff
@@ -0,0 +1,249 @@
...
+  (existing apps CAN still use the stack and CAN deploy updates). This already prevents
+  blue-green deploy scenarios where a rolling update is
+  programmed client side by creating new CF Applications.
+  However all existing apps using a locked stack SHOULD continue to run.
+
+- Mark a stack as disabled -> prevent using the stack for any app
```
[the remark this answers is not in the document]
FloThinksPi: We brought in this intermediary state to have a (optional) step in between. An operator can also decide to just move from depricated to disabled straight without setting a timestamp for locked. That being said the locked stack is important for very very large foundations where the disabled stack would create to much disturbance at once and support load. With the gradual exposure with a locked state one can do this at least in 2 steps. Optimally i would liked to propose this as an per organisation setting however that is a much larger change in the CC and in the user expeience i thought so i`d like to firstly bring this in on the same logical level that exists today - globally as the stacks are only global assets. After that and after gathering more experience one may come forward with additional optimizations to the workflow/UX/process like:
- Individual Stack states per Org
- A Stacks visibillity mapping like with services which steer which org/space can use which stack 
- etc. 

But i think its more valuable to firstly improve something in the global stack level before goint into org scoped world. So the locked state in short is just to have not one big step in the deprication process but rather 2 smaller ones to reduce/distribute effects on large consumer bases.

toc/rfc/rfc-draft-enhance-stack-handling.md
FloThinksPi: Open to also have just different states - i just thought its a lot easier to utilize this for an operator without the need to write a lot of glue code logic for the workflow outside that automates the calls to the CF API at the right time. Especially when having a lot of landscapes. 

The implementation is with timestamps actually quite simple - did a PoC for that quickly how this RFC could be implemented https://github.com/cloudfoundry/cloud_controller_ng/pull/4475/files#diff-58985a23dbdd10b651e8a688c84727b4be283d134d15d5ad1cbbb7a7d793317fR46-R58 
Most of it are tests. 

I also thought even when implementing own workflows one can just set the timestamp to the current time-minus a minute and gain the same behaviour as flags if desired. So external workflows are fully possible still although i cannot think about many differences from the build in workflow since an operator can also not set some timestamps if he doesnt want these states e.g. dont set locked go straight to disabled at a certain date. 

Just pulling in Beyhans concern about timestamps too:
>Having state implemented implicitly by timestamps makes testing and debugging harder in my opinion. Also, there could be CF foundations deployed all over the world. Yes, we will use UTC but people using different CF API clients have to translate the time and compute a state out of it in case they want to present a friendly msg to CF users. I will prefer to have an explicit state but if we can't agree on this we could have a voting about this.

So we could either leave the dates away in all error messages and just say its locked. But i guess this is also not user friendly. If we would use a freetext field like the stack description for a deprication message that anounces the timeframes e.g. `This stack is depricated  since 05/03/2025 21:30 UTC, it will be locked after 05/04/2025 21:30 UTC and will be disabled after 05/05/2025 21:30 UTC` Also in a freetext field you cannot pass proper date information if you want to let the consumer know times when what happens in a user friendly way. I think even if we print UTC times we are a lot better than before beeing it in a freetext message or programatically with the timestamps in the stacks table for each state. 

So im in favor of timestamps and programatic messages(enhanced with customizeable freetext):  `having_timestamp > timestamp_in_freetext > state > state_infreetext`. 
`timestamp > state` because its easier for an operator to handle and in any case has more information for the consumer that can be displayed. 
`programatic_info > freetext` because it reflects the true system state and a missmatch between system behaviour and stack description freetext cannot happend. 

But also fine if we all decide to use a state based aproach to change that. Maybe we vote for that specific thing with emojis on comments ?

FloThinksPi: Option 2: Timestamps with deprication/error messages from freetext field.

FloThinksPi: Option 3: States with automatic messages/errors that are extended by a freetext field.

FloThinksPi: Rewritten the RFC for option 3.

stephanme:
We start the final comment period with the goal to accept this RFC on 9th Sept.

toc/rfc/rfc-draft-enhance-stack-handling.md
```diff
@@ -0,0 +1,226 @@
...
+
+With the deprecation of the CFLinuxfs3 stack it became obvious that
+removing unsupported stacks from an existing and heavily used CF
+Foundation comes with massive problems. Ultimately a stack cannot be
+removed without all depending apps being migrated to a new stack,
+otherwise downtimes of applications that rely on the removed stack will
+occur. This RFC proposes improvements in CF to shift this unavailability
+towards lifecycle operations early and not actual app downtime - making it a
+more pleasant experience for CF users and operators alike.
+To mitigate the downsides of this approach, RFC-0041 proposes to provide custom stacks functionality.
```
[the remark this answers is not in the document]
beyhan: https://github.com/cloudfoundry/community/pull/1220/commits/b8cee24982d49a099bb9434f3bde27a258b902a3 addresses this comment.

PR #1229, opened 2025-06-30
feat: Add the org and space based certificates for syslog drains

ZPascal:
# Description

This RFC proposes to introduce support for **organization- and space-scoped client certificates** for Cloud Foundry Loggregator syslog drains using mutual TLS (mTLS), covering both **HTTPS** and **syslog+TLS** protocols. By issuing certificates at the org or space level instead of per application, this initiative will simplify certificate lifecycle management, enable centralized rotation, and facilitate integration with central certificate authorities. The change targets the reduction of operational overhead and the enhancement of tenant-level security.

## Involved Working Groups:
@cloudfoundry/toc
@cloudfoundry/wg-app-runtime-platform-logging-and-metrics-approvers
@cloudfoundry/wg-app-runtime-platform-logging-and-metrics-reviewers

### DIFF
--- a/toc/rfc/rfc-draft-org-space-based-certificates-for-syslog-drains.md
+++ b/toc/rfc/rfc-draft-org-space-based-certificates-for-syslog-drains.md
@@ -1,0 +1,49 @@
+# Meta
+[meta]: #meta
+- Name: Support Org- and Space-Based Certificates for Loggregator Syslog TLS/mTLS Drains
+- Start Date: 2025-06-30
+- Author(s): ZPascal
+- Status: Draft
+- RFC Pull Request: [community#1229](https://github.com/cloudfoundry/community/pull/1229)
+
+
+## Summary
+
+This RFC proposes to introduce support for **organization- and space-scoped client certificates** for Cloud Foundry Loggregator syslog drains using mutual TLS (mTLS), covering both **HTTPS** and **syslog+TLS** protocols. By issuing certificates at the org or space level instead of per application, this initiative will simplify certificate lifecycle management, enable centralized rotation, and facilitate integration with central certificate authorities. The change targets the reduction of operational overhead and the enhancement of tenant-level security.
+
+## Problem
+
+Currently, Cloud Foundry syslog drains using mTLS rely on **application-specific client certificates**. This causes high operational overhead, especially in large environments, because:
+
+- Certificates are tied to individual app instances and require frequent re-issuance.
+- Rotating certificates across many apps is error-prone and challenging to coordinate.
+- Managing a large number of app-level certificates leads to poor scalability and maintainability.
+
+Without org/space-scoped certificates, operators struggle with scaling, secure rotation, and integration with enterprise PKI.
+
+## Proposal
+
+The following capabilities are proposed:
+
+1. **Org/space-scoped certificate issuance**: When a syslog drain is created, the drain metadata will include the associated org and space. The Loggregator system must retrieve or generate a client certificate scoped to that org or space.
+2. **Drain-level mTLS configuration**: The syslog adapter or HTTPS forwarder must use the org/space certificate when establishing mTLS connections to the remote drain endpoint.
+3. **Centralized and automatic rotation support**: Certificates should be automatically renewed and reloaded with minimal or zero impact to traffic flow.
+4. **Audit and visibility**: Logs and drain metadata must include the org/space identity used for certificate-based authentication.
+
+### Example configuration
+
+```bash
+cf create-user-provided-service ORG-NAME -p '{"ca":"-----BEGIN CERTIFICATE-----\nMIIH...-----END CERTIFICATE-----"}'
+```
+
+or
+
+```bash
+cf create-user-provided-service SPACE-NAME -p '{"ca":"-----BEGIN CERTIFICATE-----\nMIIH...-----END CERTIFICATE-----", "cert":"-----BEGIN CERTIFICATE-----\nMIIH...-----END CERTIFICATE-----","key":"-----BEGIN PRIVATE KEY-----\nMIIE...-----END PRIVATE KEY-----"}'
+```
+
+### Configuration Flow
+
+- When an app binds a syslog drain, the Cloud Controller should include org and space GUIDs in the drain metadata.
+- The system retrieves a Certifacte Authority for that org or space from a binding.
+- The drain connection must use this certificate for TLS/mTLS authentication.


toc/rfc/rfc-draft-org-space-based-certificates-for-syslog-drains.md
```diff
@@ -0,0 +1,49 @@
...
+
+```bash
+cf create-user-provided-service SPACE-NAME -p '{"ca":"-----BEGIN CERTIFICATE-----\nMIIH...-----END CERTIFICATE-----", "cert":"-----BEGIN CERTIFICATE-----\nMIIH...-----END CERTIFICATE-----","key":"-----BEGIN PRIVATE KEY-----\nMIIE...-----END PRIVATE KEY-----"}'
+```
+
+### Configuration Flow
+
+- When an app binds a syslog drain, the Cloud Controller should include org and space GUIDs in the drain metadata.
+- The system retrieves a Certifacte Authority for that org or space from a binding.
+- The drain connection must use this certificate for TLS/mTLS authentication.
```
beyhan, line 49: Could you please add some words about the backwards compatibility of this flow with the current flow to create and consume a syslog drain.

toc/rfc/rfc-draft-org-space-based-certificates-for-syslog-drains.md
```diff
@@ -0,0 +1,49 @@
...
+- Name: Support Org- and Space-Based Certificates for Loggregator Syslog TLS/mTLS Drains
+- Start Date: 2025-06-30
+- Author(s): ZPascal
+- Status: Draft
+- RFC Pull Request: [community#1229](https://github.com/cloudfoundry/community/pull/1229)
+
+
+## Summary
+
+This RFC proposes to introduce support for **organization- and space-scoped client certificates** for Cloud Foundry Loggregator syslog drains using mutual TLS (mTLS), covering both **HTTPS** and **syslog+TLS** protocols. By issuing certificates at the org or space level instead of per application, this initiative will simplify certificate lifecycle management, enable centralized rotation, and facilitate integration with central certificate authorities. The change targets the reduction of operational overhead and the enhancement of tenant-level security.
```
beyhan, line 12: Regarding "By issuing certificates at the org or space level instead of per application". Do you mean for the syslog drain use case? The https://docs.cloudfoundry.org/devguide/deploy-apps/instance-identity.html feature is used in different use cases.

beyhan (changes requested):

stephanme (changes requested):
Can you add sections with proposed changes per CF component (e.g. Cloud Controller, loggregator, CF CLI). This helps to better understand the consequences of this RFC and required efforts. It also makes it easier to invite the affected WG areas for a review.

toc/rfc/rfc-draft-org-space-based-certificates-for-syslog-drains.md
```diff
@@ -0,0 +1,49 @@
...
+```bash
+cf create-user-provided-service ORG-NAME -p '{"ca":"-----BEGIN CERTIFICATE-----\nMIIH...-----END CERTIFICATE-----"}'
+```
+
+or
+
+```bash
+cf create-user-provided-service SPACE-NAME -p '{"ca":"-----BEGIN CERTIFICATE-----\nMIIH...-----END CERTIFICATE-----", "cert":"-----BEGIN CERTIFICATE-----\nMIIH...-----END CERTIFICATE-----","key":"-----BEGIN PRIVATE KEY-----\nMIIE...-----END PRIVATE KEY-----"}'
+```
+
```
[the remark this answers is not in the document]
chombium: @stephanme The call above will create credentials which have to be bound to the app. The problem with this approach is that these credentials are available to the app via the `VCAP_SERVICES` and will not be available to the Syslog Agent when creating the Syslog Drain. If this is accepted an app will have to somehow share the credentials with the Syslog Agent which will be hard to do as the Syslog Agent is a special kind of user provided service  implemented as part of CF and not via an external service broker. An external broker could have collected all needed credentials and open a connection, but this is not the case.

toc/rfc/rfc-draft-org-space-based-certificates-for-syslog-drains.md
chombium: IMO, this is only for the certificates used for Syslog drains and has nothing to do with the instance identity certificates.

chombium (changes requested):
IMHO, we don't need this change. What we need is a better documentation how to achieve this with the current way how syslog drains are created and work and the other CF tools we have at hand.

chombium:
I understand the problem and complexity of managing certificates which this RFC tres to address.

At the moment, with the current implementation, one can create a single user provided service defining a Syslog Drain and bind it to multiple apps.

```mermaid
flowchart LR

app1["app 1"]
ups["User Provided Service"] --> b1["service binding 1"] --> sl1["Syslog Drain 1"]
app1 --> b1
app2["app 2"]
ups["User Provided Service"] --> b2["service binding 2"] --> sl2["Syslog Drain 2"]
app2 --> b2
app3["app 3"]
ups["User Provided Service"] --> b3["service binding 3"] --> sl3["Syslog Drain 3"]
app3 --> b3
appn["app n"]
ups["User Provided Service"] --> bn["service binding n"] --> sln["Syslog Drain n"]
appn --> bn
```

This way one can share the same Syslog Drain certificates between apps.

We've also used to have a [cf drain CLI plugin](https://github.com/cloudfoundry/cf-drain-cli) for managing `single` and `space drains` which we've deprecated few years ago because of low usage.


The Syslog Drains are created as user provided services and hence are scoped to a space. One can share a service between spaces in an org with `cf share-service` and with that if the service is shared among all of the spaces in an organization it will have an organization wide syslog drain.

The only possible downsides of this approach are:
-  one user provided service Syslog Drain per Syslog Drain URL. From my personal experience, I have rarely seen that a single organization uses different urls for their syslog drains
-  possible duplication if the same certificates (ca, cert and key). IMO, we should not stress much about few more bytes in the Cloud Controller's database


This RFC:
- opens a possibility to reuse credentials with different Syslog drain URLs which practically means that different client connections can be identified with the same client certificates which is a bad security practice
- breaks the `cf cups` call as the command for creating Syslog Drain a specific `-l` parameter followed by the Syslog drain url should be used. `cf cups mydrain -l https://my-drain.example.com -p {...}`


IMHO, we don't need this change. What we need is a better documentation how to achieve this.

Here is an example of the whole flow:
Let's say we have an organization org1 with two spaces, space1 and space2 and an org2 with space3. There is an app1 in the space1, app2 in space2 and app3 in space3.

```shell
# target org1 and space1
cf target -o org1 -s space1

# create the syslog drain
cf cups mydrain -l https://my-drain.example.com -p {...}

# bind app1 to mydrain
cf bind-service app1 mydrain

# share the service with space2
cf share-service mydrain -s space2

# share the service with space3 in org2
cf share-service mydrain -s space3 -o org2

# target space2
cf target -s space2

# bind app2 to mydrain
cf bind-service app2 mydrain

# target org2 space3
cf target -o org2 -s space3

# bind app3 to mydrain
cf bind-service app3 mydrain
```

stephanme:
Based on the discussion above I suggest to close this RFC. There is no need for an RFC to improve the documentation :-)

beyhan:
@ZPascal will be great if the outcome of this is to improve the existing docs.

chombium:
Hi,

I did few more tests of the things that I've written previously and found out that I was wrong about service instance sharing and have mixed "normal" platform/CF Market provided services with user-provided-services :-/ The diagram about sharing one/reusing an existing service instance is still valid in a single space.

I've tried two more things:
1. Share a user-provided-service instance with [cf share-instance call](https://docs.cloudfoundry.org/devguide/services/sharing-instances.html) and found out that the `user provided services cannot be shared`. The error comes from the following Cloud Controller's [validation](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/actions/service_instance_share.rb#L75-L80).
2. Add a [service key](https://docs.cloudfoundry.org/devguide/services/service-keys.html) to a user-provided-service and saw that that's also not supported. The error was `Binding parameters are not supported for user-provided service instances` and comes from the Cloud Controller's [service_credential_bindings_controller](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/controllers/v3/service_credential_bindings_controller.rb#L278)

This means that the RFC is still valid.

IMO, as the Syslog Drains are only one type of user-provided-services we should definitely not change the behavior for all user-provided-services or even worse, add special handling for Application Syslog Drains.

The only way forward with this RFC which I see, is to adjust the whole Syslog Drain creation process via the `cf create-user-privided-service`, so that we add a parameter which will define if the drain is valid for a space or an organization.

- A cf user would call something like `cf create-user-provided-service -l "https://mydrain.com:5678 -p {"ca"":"ca_cert", "cert":"cert_data", "key":"key_data", "drain-validity":"org"}`

- `drain_validity` is the "code name"(we need to find something better) will be either "space" or "org". The default value will be "" to keep the things backward compatible.

- The Cloud Controller should parse the value of `drain_validity` and:
  - "" empty - don't create service bindngs
  - "space" - create bindings for all apps in the space
  - "org" - create bindings for all apps in the org

The Cloud Controller's [syslog drain url controller](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/unit/controllers/internal/syslog_drain_urls_controller_spec.rb) will be unchanged.

The Syslog Agent won't be adjusted as all the heavy lifting will be done by the Cloud Controller. The only possible thing to adjust would be error handling.

There are three major challenges with this approach:
1. The Cloud Controller will have to follow what's happening with the apps (push, delete and similar) in a space or org and what's happening with the service instance(create, update, delete) and create, delete or update many service bindings at once depending on the number of the apps.
2. As the Syslog Drain validation is done later in the Loggregator's Syslog Agent, it may happen that the Cloud Controller will have to create many bindings which won't work.
3. The Syslog Agent will create a Syslog Drain for each application in a space or an org. Making exceptions for a specific app might be useful. An app dev should be able to create a separate Syslog Drain for a particular app and don't use the space or org Syslog Drain. This could be done by deleting the binding for the app, but it will complicate the things in the Cloud Controller with some sort of an exclude list.

Whatever we do, there will be some non-trivial changes in the Cloud Controller needed. We should keep the implementation effort and the complexity in mind and the added value for the app devs. This is a nice to have feature, but I cannot estimate it's possible adoption and usage. Tbh, I understand that having org/space Syslog drains will reduce the management effort for the app devs, but based on the implementation effort, I'm not totally convinced that we need this. What do you think about this? It would be great to hear from someone from the CAPI team.

We have one [issue](https://github.com/cloudfoundry/loggregator-agent-release/issues/52) about org and space Syslog drains from @Benjamintf1, the former lead of the CF Logging and Metrics Team. I don't know if he had some other ideas about the implementation which he would like to share...

stephanme:
I don't think that there is a blocker making UPS instances sharable or support service keys. It may become more complex to find all bindings that need to be updated when the UPS instance is updated. Bindings to a UPS are updated directly when the credentials of the UPS are updated - in contrast to service bindings to regular service instances where a new binding has to be created to update credentials.

I think this would be a good step towards reducing the differences of user-provided services and manage services. We had a similar discussion in the [[RFC] Service Credential Binding Rotation for Apps](https://github.com/cloudfoundry/community/pull/1198#discussion_r2206646835)

PR #1231, opened 2025-07-03
Fix failing github sync settings action

beyhan:
Our Sync Github Organization Settings is failing with:

```
{"component":"peribolos","level":"fatal","msg":"Configuration failed: failed to configure cloudfoundry members: users in both roles: gerg","severity":"fatal","time":"2025-07-03T00:16:58Z"}
```

I suspect that https://github.com/cloudfoundry/community/pull/1230 configured Greg with upper case and this is causing an issue with the Peribolos tooling. This is a try to fix that.

### DIFF
--- a/toc/TOC.md
+++ b/toc/TOC.md
@@ -124,7 +124,7 @@ execution_leads:
 - name: Chris Weibel
   github: cweibel
 - name: Greg Cobb
-  github: Gerg
+  github: gerg
 - name: Ruben Koster
   github: rkoster
 - name: Stephan Merker


stephanme (approved):
This should fix it for now. The underlying question is: are github user ids case-sensitive or not. The org automation treats them case-sensitive (gerg != Gerg).
Looks like Github treats ids case-insensitive (e.g. both https://github.com/Gerg and https://github.com/gerg lead to the same page).
Mid-term I can adapt the org automation if it is well specified by github.

PR #1239, opened 2025-07-08
Add Plamen Bardarov as reviewier in App Runtime Interfaces WG

plamen-bardarov:


beyhan:
The WG doesn't have more than three active approvers that is why one approval in this case should be fine. Ping @cloudfoundry/wg-app-runtime-interfaces-buildpacks-stacks-approvers

stephanme:
@arjun024, @ForestEckhardt, @robdimsdale, @simonjjones, @Gerg: Can you please vote?
If I hear nothing, I plan to merge this PR end of the week.

PR #1243, opened 2025-07-08
add neowulf to org/contributors.yml

neowulf:


beyhan:
Hi @neowulf,

thank you for this pr. Could you please sing the CLA. If you have any issues with that @rkoster and @ameowlia could help. Please also add a short description about why you would like to be added.

PR #1244, opened 2025-07-09
Add Stefan Yonkov as reviewer for cloudfoundry/cf-java-client

s-yonkov-yonkov:


s-yonkov-yonkov:
@stephanme As discussed in our last meeting I will be one of the deploy-service representatives as reviewer for this project.

beyhan:
We have four approvers in this area but I don't think all of them are active. That is why, it looks to me that one approval from @cloudfoundry/wg-app-runtime-interfaces-java-tools-approvers will be fine in this case. If we get more even better. @cloudfoundry/wg-app-runtime-interfaces-java-tools-approvers will be great if you could put thumps up or +1 on this pr if you agree with the proposal.

Kehrlann (approved):

Kehrlann:
Works for me (Java Tools approver). Ok for single approval too. 👍

stephanme (approved):

stephanme:
Welcome @s-yonkov-yonkov as reviewer in the Java Tools area.

PR #1245, opened 2025-07-09
Add Kristian Atanasov as reviewer for cloudfoundry/cf-java-client

theghost5800:


beyhan:
We have four approvers in this area but I don't think all of them are active. That is why, it looks to me that one approval from @cloudfoundry/wg-app-runtime-interfaces-java-tools-approvers will be fine in this case. If we get more even better. @cloudfoundry/wg-app-runtime-interfaces-java-tools-approvers will be great if you could put thumps up or +1 on this pr if you agree with the proposal.

theghost5800:
I have updated the PR.

Kehrlann (approved):

beyhan (approved):

stephanme (approved):

stephanme:
Welcome @theghost5800 as reviewer in the Java Tools area.

PR #1256, opened 2025-07-23
[RFC] Document CF API V3 with OpenAPI

FloThinksPi:
Just as a side note this was already briefly discussed in the App Runtime Interfaces WG and had positive feedback that something like that is thinkable for the WG

[Link for easy viewing](https://github.com/FloThinksPi-Forks/community/blob/rfc-openapi/toc/rfc/rfc-draft-document-cf-api-with-openapi.md)

### DIFF
--- a/toc/rfc/rfc-draft-document-cf-api-with-openapi.md
+++ b/toc/rfc/rfc-draft-document-cf-api-with-openapi.md
@@ -1,0 +1,92 @@
+# Meta
+
+- Name: Documenting the CF API with OpenAPI
+- Start Date: 2025-07-23
+- Author(s): @flothinkspi
+- Status: Draft
+- RFC Pull Request: [community#1256](https://github.com/cloudfoundry/community/pull/1256)
+
+## Summary
+
+With the acceptance of
+[RFC-32](https://github.com/cloudfoundry/community/blob/d7b48620d0da3bcadeea18aaf64f6fa36c329e7f/toc/rfc/rfc-0032-cfapiv2-eol.md),
+a problem arises for users of CF who rely on unsupported client
+libraries that are not maintained and just work with the deprecated CF V2 API. To ensure the success of
+[RFC-32](https://github.com/cloudfoundry/community/blob/d7b48620d0da3bcadeea18aaf64f6fa36c329e7f/toc/rfc/rfc-0032-cfapiv2-eol.md)
+and to enable users to better transition to the V3 API, and also future-proof the CF API documentation, this RFC proposes
+to document the CF API V3 using the OpenAPI specification.
+
+## Problem
+
+## Motivation
+
+1. Improve the overall CF user experience, increasing the adoption and
+popularity of CloudFoundry.
+
+2. Support the adoption of
+[RFC-32](https://github.com/cloudfoundry/community/blob/d7b48620d0da3bcadeea18aaf64f6fa36c329e7f/toc/rfc/rfc-0032-cfapiv2-eol.md)
+by offering an more user-friendly path for users away from the CF V2 API.
+
+3. Provide a widely adopted documentation format (OpenAPI) that enables many new usecases from programatic access to the CF API docs and schema to generating Documentation pages, automatic tests of documentation vs implementation or also better AI Agent integration.
+
+## Proposal
+
+### Documenting the CF V3 API with OpenAPI
+
+With [OpenAPI](https://www.openapis.org/), APIs can be defined in a
+machine-readable format (YAML), enabling the generation of client
+libraries for various programming languages and the rendering of
+interactive HTML API documentation. The OpenAPI specification is widely
+adopted and supported by many tools. A [Proof of Concept
+(PoC)](https://github.com/FloThinksPi/cf-api-openapi-poc) and [another Proof of Concept](https://github.com/cloudfoundry-community/capi-openapi-spec) is already
+exploring the documentation of the CF V3 API as an OpenAPI spec,
+providing a glimpse (currently with incorrect content) of the potential
+look and feel of this documentation format.
+
+The CloudFoundry organization MAY document the CF API V3 using the
+OpenAPI specification. Which then can be used by CF Maintainers or CF Users to
+leverage a huge toolchain around OpenAPI to generate client libraries, documentation pages, and more. This MAY render a transition like for example from a unsupported client library that only supports the CF V2 api towards using the CF V3 api a lot more easier since the available toolchain around OpenAPI is big, valuable, a time saver and then available to CF Users.
+Such usecases include for example:
+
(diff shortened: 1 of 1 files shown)


toc/rfc/rfc-draft-document-cf-api-with-openapi.md
```diff
@@ -0,0 +1,92 @@
...
+- [Swagger Codegen](https://swagger.io/tools/swagger-codegen/) for generating client libraries
+- [OpenAPI Generator](https://openapi-generator.tech/) for generating client libraries and server stubs in various languages.
+- [Redoc](https://redocly.com/redoc/) for generating static API documentation.
+- [Postman](https://www.postman.com/) for API testing and documentation.
+- [Dredd](https://dredd.org/en/latest/) for testing API documentation against the implementation.
+- [AI Agents](https://github.com/janwilmake/openapi-mcp-server) for improved integration with AI-powered tools.
+
+### Phases
+
+**Implementation** - Convert <https://github.com/cloudfoundry-community/capi-openapi-spec> to a repository in the official cloudfoundry organization.Document the CF API V3 using the OpenAPI specification. Also generate a html documentation to replace the current page. Optionally build up, if possible, a validation(integration test) that validates if the OpenAPI spec is in sync with the actual implementation of the CF API V3.
```
Samze, line 61: **thought** I see this is using perl code to generate the API. There isn't much (any?) perl in the CF OSS so I'd be concerned about the maintainability of continuing this approach.

toc/rfc/rfc-draft-document-cf-api-with-openapi.md
```diff
@@ -0,0 +1,92 @@
...
+- [Swagger UI](https://swagger.io/tools/swagger-ui/) for interactive API documentation.
+- [Swagger Codegen](https://swagger.io/tools/swagger-codegen/) for generating client libraries
+- [OpenAPI Generator](https://openapi-generator.tech/) for generating client libraries and server stubs in various languages.
+- [Redoc](https://redocly.com/redoc/) for generating static API documentation.
+- [Postman](https://www.postman.com/) for API testing and documentation.
+- [Dredd](https://dredd.org/en/latest/) for testing API documentation against the implementation.
+- [AI Agents](https://github.com/janwilmake/openapi-mcp-server) for improved integration with AI-powered tools.
+
+### Phases
+
```
Samze, line 60: **thought** Should the spec/generation live just live in the https://github.com/cloudfoundry/cloud_controller_ng repository?

Samze, line 60: **thought**  https://github.com/rswag/rswag would be an interesting approach, create request specs that can be used to generate open api spec. That way we get a open api doc and some verification in a single definition. However that would mean creating whole bunch of new requests specs.

FloThinksPi: I tried that before the issue with that is that we dont use native rails - e.g. we dont use active records but sequel library for the models. Rswag just worked for me when its a absolute default ruby project. But i also like to give it a shot again since i last tried it 1-2 years ago. Optimally we could use something like that but in any case writing the OpenAPI spec by hand is already an improvement from writing the HTML/MD page oneself i guess. Especially for the consumer side the format is the only thing that matters. How its produced is secondary for a CF consumer, relevant for the Working Group though.

FloThinksPi: Yep definitely, the idea is to work on it in a separate repo until we want to switch the leading documentation format - then we copy it over into the CC_NG repo, including the processes(release webpage etc.) which likely will be based on GitHub actions - and archive the one.

toc/rfc/rfc-draft-document-cf-api-with-openapi.md
stephanme: I would start with a POC phase that demonstrates different options how to convert the existing implementation into an OpenAPI spec including how CAPI devs would update it. One non-trivial endpoint should be enough but this endpoint should be implemented completely.
Whether we use the mentioned approach from the community project or not is a detail.

toc/rfc/rfc-draft-document-cf-api-with-openapi.md
```diff
@@ -0,0 +1,92 @@
...
+**Implementation** - Convert <https://github.com/cloudfoundry-community/capi-openapi-spec> to a repository in the official cloudfoundry organization.Document the CF API V3 using the OpenAPI specification. Also generate a html documentation to replace the current page. Optionally build up, if possible, a validation(integration test) that validates if the OpenAPI spec is in sync with the actual implementation of the CF API V3.
+
+**Checkpoint** - TOC review and approval of the project, ensuring the
+OpenAPI documentation is complete in comparison to the current one and of appropriate quality. Replacement of the current CF API V3 website with representing the OpenAPI spec. Also release the OpenAPI spec for programmatic access to the CF API V3 documentation.
```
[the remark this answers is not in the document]
stephanme: First checkpoint is probably the POC (one endpoint):
- How to create the initial spec from existing API documentation? (the mentioned POC shows one way)
- Does the dev process work?
- How is it integrated into CI and CAPI release process?
- Can we test the spec against the implementation?
- Maybe there is a way to generate the spec from source code. How would this look like? Where can we maintain metadata like examples, manually written docs etc in the sources?

Next phases could then be:
- complete implementation
- checkpoint as mentioned
  - Is the spec complete and usable?
- remove old API documentation

stephanme (changes requested):
Another point to think about: how to continue with the existing CF client libs: cf-java-client and go-cfclient?

Do we maintain multiple client libs? The existing ones for compatibility and the potentially new ones that are (partly) generated from the OpenAPI spec.

The existing ones provide additional value on top of the pure language bindings:
- authentication support and token refresh
- pagination handling
- slicing long guid lists into chunks that fit into urls
- higher level operations (cf-java-client has an operations api but only for v2)
- integration of additional api endpoints like loggregator

This goes probably beyond this RFC. The spec clearly helps to get a language binding for the CF API from scratch which is good if no library exists yet or is hopelessly outdated. But if usable client libs exist (java, golang, also python), it becomes a longer way.

toc/rfc/rfc-draft-document-cf-api-with-openapi.md
Gerg, line 61: Since it looks like @wayneeseguin is the sole maintainer of that repo, we should involve them in this discussion.

FloThinksPi:
I made some adoptions and worked in @beyhan and @stephanme comments.

Gerg (approved):

stephanme (approved):

rkoster (approved):

PR #1258, opened 2025-07-23
Create repo for consolidation of blobstore CLI's

aramprice:
This repo is expected to contain the blobstore CLI's used in the BOSH ecosytem, at present these are:
- https://github.com/cloudfoundry/bosh-azure-storage-cli
- https://github.com/cloudfoundry/bosh-gcscli
- https://github.com/cloudfoundry/bosh-s3cli
- https://github.com/cloudfoundry/bosh-ali-storage-cli

See also
- https://github.com/cloudfoundry/cloud_controller_ng/pull/4443

### DIFF
--- a/orgs/orgs.yml
+++ b/orgs/orgs.yml
@@ -196,6 +196,10 @@ orgs:
         has_projects: false
         has_wiki: false
         private: true
+      bosh-blobstore-clis:
+        description: CLIs for interacting with blobstores
+        has_projects: false
+        has_wiki: false
       bosh-bootloader:
         allow_merge_commit: false
         allow_squash_merge: false
--- a/toc/working-groups/foundational-infrastructure.md
+++ b/toc/working-groups/foundational-infrastructure.md
@@ -368,6 +368,7 @@ areas:
   - cloudfoundry/bosh-azure-storage-cli
   - cloudfoundry/bosh-apt-resources
   - cloudfoundry/bosh-bbl-ci-envs
+  - cloudfoundry/bosh-blobstore-clis
   - cloudfoundry/bosh-bootloader
   - cloudfoundry/bosh-cli
   - cloudfoundry/bosh-common


toc/working-groups/foundational-infrastructure.md
```diff
@@ -368,6 +368,7 @@ areas:
   - cloudfoundry/bosh-azure-storage-cli
   - cloudfoundry/bosh-apt-resources
   - cloudfoundry/bosh-bbl-ci-envs
+  - cloudfoundry/bosh-blobstore-clis
```
beyhan, line 371: In the mean time there is an RFC about this topic. The RFC suggests to add a new are called "Storage CLI" in https://github.com/cloudfoundry/community/blob/8fc25e9393de1624c84c77f71b2786134b35c852/toc/rfc/rfc-draft-cc-blobstore-storage-cli.md?plain=1#L39-L48. We should follow the RFC and create a new area for it combining the suggested approvers from CAPI.

orgs/orgs.yml
```diff
@@ -196,6 +196,10 @@ orgs:
         has_projects: false
         has_wiki: false
         private: true
+      bosh-blobstore-clis:
+        description: CLIs for interacting with blobstores
+        has_projects: false
+        has_wiki: false
```
beyhan, line 202: The [RFC](https://github.com/cloudfoundry/community/blob/8fc25e9393de1624c84c77f71b2786134b35c852/toc/rfc/rfc-draft-cc-blobstore-storage-cli.md?plain=1#L39-L48) is talking about `storage-cli` repository. I think it make sense to keep it more generic without bosh in the name and storage instead of blobstore. If we want to have it plural like `storage-clis` we should provide that feedback in the RFC and try to agree on that in the RFC. I don't have strong opinion on this.

beyhan (changes requested):

PR #1263, opened 2025-07-29
Disable enforcing branch-protection for admins

beyhan:
This disables branch-protection rules for admins in FI WG repositories included in this pr. We use Github deploy keys in our automation which doesn't work with the default branch protection rules. Concrete, following Github setting is deactivated for admin users:

- Do not allow bypassing the above settings

### DIFF
--- a/orgs/branchprotection.yml
+++ b/orgs/branchprotection.yml
@@ -360,6 +360,397 @@ branch-protection:
               teams: ["wg-foundational-infrastructure-bots"]
           include: [ "^main$" ]
 
+        bosh:
+          allow_deletions: false
+          allow_disabled_policies: true
+          allow_force_pushes: false
+          enforce_admins: false
+          include:
+          - ^main$
+          - ^v[0-9]*$
+          protect: true
+          required_pull_request_reviews:
+            bypass_pull_request_allowances:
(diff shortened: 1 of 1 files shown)


stephanme (approved):

beyhan:
> What is the justification for still using deployment keys? Is there a technical reason?

@rkoster there is discussion about this in the bosh-core-dev channel [here](https://cloudfoundry.slack.com/archives/C094QRLEQ/p1753254638687879). I think we should unblock the situation and discuss further in the FI WG meetings.

rkoster (approved):
Let's merge to unblock the releases, but we should at the ToC level decide on our stance on deploys keys and if we want to support those we should look into adopting branch protection rulesets for finer grained control.

rkoster:
> > What is the justification for still using deployment keys? Is there a technical reason?
> 
> 
> 
> It seems like a better security posture to have per-repo keys vs a bot account key that has privileged access to all the repos

Yeah good point, the after refreshing my context on the matter my issue with deployment keys in the past was the fact that they did not work with branch protection rules. These days however there seem to be multiple options: the allow admin override for classic branche protection rules and native exclusion support in branch protection rulesets.

PR #1269, opened 2025-07-31
Add branch protection rules for uaa and uaa-release

duanemay:
We use Github deploy keys in our automation for uaa and uaa-release, which doesn't work with the default branch protection rules.

### DIFF
--- a/orgs/branchprotection.yml
+++ b/orgs/branchprotection.yml
@@ -941,3 +941,37 @@ branch-protection:
             bypass_pull_request_allowances:
               teams: ["wg-service-management-cloud-service-broker-bots"]
           include: [ "^main$" ]  
+
+        uaa:
+         protect: true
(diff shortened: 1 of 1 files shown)


cweibel (approved):
Known issue that deploy keys do not work with branch protection

duanemay:
@cweibel Thank you for pointing that out, this should get our bot account back to working, we can go from there.

beyhan:
> At this point doesn't it make sense to just incorporate the enforce admins false change into the core branch protection automation? From a permissions point of view it would only affect working group leads and toc members who can already change branch protection rules via the UI to temporary disable admin enforcement, so we are already relying on their good intentions anyway.

@rkoster I added this as discussion point to the TOC agenda next Tuesday. It sounds reasonable to avoid to much static configuration in our branchprotection.yml.

PR #1282, opened 2025-08-11
Disable enforcing branch-protection for jumpbox deployment

benjaminguttmann-avtq:
Required for deploy keys and was missed in https://github.com/cloudfoundry/community/pull/1263


stephanme:
Duplicates #1278

PR #1290, opened 2025-08-15
Remove user ifindlay-cci

beyhan:
The [Sync Github Organization Settings](https://github.com/cloudfoundry/community/actions/runs/16989865344/job/48166497010) is failing because there is a issue with the github user @ifindlay-cci. This pr suggest to remove the user until we clarify the issue and the change can be reverted when we have clarity on it and it doesn't fail the automation.

### DIFF
--- a/orgs/contributors.yml
+++ b/orgs/contributors.yml
@@ -77,7 +77,6 @@ orgs:
     - hibell
     - hsinn0
     - iaftab-alam
-    - ifindlay-cci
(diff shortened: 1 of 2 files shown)


ifindlay-cci:
I am getting spammed by requests to join. @beyhan 

I decline the request and a week later I am invited again.

Make it stop.

stephanme (approved):

PR #1335, opened 2025-10-14
Update bosh-agent-index branch protection

aramprice:
This changes the branch protection for bosh-agent-index to match that used by bosh-cli.

Not sure if this is the correct incantation for this change.


beyhan (approved):

beyhan:
This is a use case which we agreed in the TOC to allow and to not further block the pipeline I'm going to merge this.

PR #1338, opened 2025-10-15
Adjust branch protection for socks5-proxy

aramprice:
To address CI failures, see:
https://bosh.ci.cloudfoundry.org/teams/main/pipelines/socks5-proxy/jobs/bump-deps/builds/30

### DIFF
--- a/orgs/branchprotection.yml
+++ b/orgs/branchprotection.yml
@@ -770,6 +770,23 @@ branch-protection:
             require_code_owner_reviews: true
             required_approving_review_count: 1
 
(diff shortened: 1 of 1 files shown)


rkoster (approved):

beyhan:
@aramprice unfortunately the automation is taking its time. The GitHub action is divided into two steps. First apply the approver and reviewer configs and then the branch protection. The first step takes ~4 hour and if there is a new pr merged in between the second step is canceled. In our case there were some merges in between. It looks like in ~1 hour the protection step will start.

PR #1345, opened 2025-10-23
Added 'app-autoscaler' repo with admin rights for app-autoscaler team

bonzofenix:
As part of our migration from app-autoscaler-release repo to app-autoscaler we need admin access to the unarchived app autoscaler https://github.com/cloudfoundry/community/pull/1326

### DIFF
--- a/orgs/orgs.yml
+++ b/orgs/orgs.yml
@@ -2662,6 +2662,7 @@ orgs:
         members: []
         privacy: closed
         repos:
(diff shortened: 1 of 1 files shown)


Gerg (approved):

beyhan:
Merging because we have approvals from the majority of the TOC member and to not block people. FYI @cweibel and @rkoster

PR #1347, opened 2025-10-24
add itsouvalas to app-runtime

itsouvalas:


itsouvalas:
@stephanme as suggested

PR #1389, opened 2025-12-12
RFC for introducing "CF on KinD"

c0d1ngm0nk3y:
> Cloud Foundry local setup is currently cumbersome and has a steep learning curve. This makes it difficult for new users to get started and slows down component development. We propose a simple and fast way to run Cloud Foundry locally, providing a better first-time user experience. Additionally, this enables developers to rapidly prototype, develop, and test new ideas in an inexpensive setup.

[preview](https://github.com/sap-contributions/cf-community/blob/cf-on-kind-rfc/toc/rfc/rfc-draft-cf-on-kind.md)

### DIFF
--- a/toc/rfc/rfc-draft-cf-on-kind.md
+++ b/toc/rfc/rfc-draft-cf-on-kind.md
@@ -12,7 +12,7 @@ Cloud Foundry local setup with Bosh and cf-deployment is currently cumbersome an
 
 ## Problem
 
-Setting up Cloud Foundry locally is currently cumbersome which creates a significant barrier to entry.
+Setting up Cloud Foundry locally using Bosh and cf-deployment is currently cumbersome which creates a significant barrier to entry.
 
 First-time users encounter significant difficulty when they simply want to try out the `cf push` workflow. Currently, there is no local setup using existing community tools and technologies that can be ready within minutes.
 


toc/rfc/rfc-draft-cf-on-kind.md
```diff
@@ -0,0 +1,36 @@
+# Meta
+[meta]: #meta
+- Name: CF on KinD
+- Start Date: 2025-12-12
+- Author(s): @beyhan, @c0d1ngm0nk3y, @loewenstein-sap, @modulo11, @mvach, @nicolasbender, @pbusko
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (fill in with PR link after you submit it)
```
c0d1ngm0nk3y:
```suggestion
- RFC Pull Request: https://github.com/cloudfoundry/community/pull/1389

PlamenDoychev:
It's not clear to me who we're targeting with this proposal. 
Shouldn't we;
1) position it for those who want to do an initial CF try-out and experience "cf push" locally. 
2) In addition to 1) target component developers as well but with some "limitations", because it will not be possible local development for all CF components.

loewenstein-sap (commented):
Hi @tlakner,

thank you for your interest in this RFC. You say

> The RfC refers to CF Classic (and not CF Korifi). Suggestion is to also state this in the description more clearly. 

I would argue that "Cloud Foundry" without any additional attribute means the classical Cloud Foundry deployed by Bosh as we know it. Anyway, let me suggest we mention Bosh and cf-deployment explicitly to make this entirely clear.

> I also agree to Plamen to state more clearly the main use case for CF on kinD - local development environment for CF component developers who get a lower entry barrier with this approach than before.

I would argue that we explicitly address first-time users and Cloud Foundry component developers in the Problem section. What else would you suggest in extent to the following?
- https://github.com/cloudfoundry/community/blob/16a60c768a39c7079d0df0a7c5c6010a74cd5e74/toc/rfc/rfc-draft-cf-on-kind.md?plain=1#L17
- https://github.com/cloudfoundry/community/blob/16a60c768a39c7079d0df0a7c5c6010a74cd5e74/toc/rfc/rfc-draft-cf-on-kind.md?plain=1#L19

Best regards
Jan

mauricebrinkmann:
I like this approach very much as I was asked mostly from young devs several times after giving CF talks "how and where can I try it out" and this approach helps closing this gap in my eyes, since cf on your laptop's hypervisor from Pivotal (https://github.com/cloudfoundry-attic/cfdev) and from Suse with their own Stemcells (https://github.com/SUSE/scf) back then are discontinued. But I also understand CF on KinD as a lightweight test environment for platform ops to test out new component combinations, too. Both use cases are strong and appreciated. Is there a repo for early adopters..? ;-D

beyhan (approved):

stephanme (approved):

PR #1391, opened 2025-12-16
Add mrosecrance as a FI Approver

mrosecrance:
I was just removed as part of https://github.com/cloudfoundry/community/pull/1382 and would like to get re-added. I use the permissions given as part of this at least once a month to check concourse and to schedule things like CI maintenence and stemcell cutting.


aramprice (dismissed):
Pending the update requested.

beyhan (approved):

PR #1437, opened 2026-02-13
RFC: Full mTLS for Diego Container-to-Container Traffic

rkoster:
## Summary

This RFC proposes implementing full mutual TLS (mTLS) for container-to-container (C2C) traffic in Diego, enabling applications to both authenticate themselves and verify the identity of connecting applications.

**[View the full RFC](https://github.com/cloudfoundry/community/blob/rfc-diego-c2c-mtls/toc/rfc/rfc-draft-diego-c2c-mtls.md)**

The approach introduces:
- **Port 62443**: New HTTP-based C2C mTLS listener with XFCC header forwarding (server-side)
- **Port 61445**: Egress proxy for automatic instance identity certificate injection (client-side)

## Key Points

- Backwards compatible: existing port 61443 remains unchanged
- Dual opt-in model: both operators (BOSH properties) and app authors must opt-in
- Phased implementation allowing gradual adoption
- Server apps can extract caller identity from `X-Forwarded-Client-Cert` header

## Implementation Phases

1. **Phase 1**: Server-side HTTP-based C2C mTLS port (62443)
2. **Phase 2**: Client-side egress proxy (61445) with selective cert injection
3. **Phase 3**: Full integration

cc @cloudfoundry/toc @cloudfoundry/wg-app-runtime-platform

### DIFF
--- a/toc/rfc/rfc-draft-diego-c2c-mtls.md
+++ b/toc/rfc/rfc-draft-diego-c2c-mtls.md
@@ -1,0 +1,280 @@
+# Meta
+[meta]: #meta
+- Name: Full mTLS for Diego Container-to-Container Traffic
+- Start Date: 2026-02-13
+- Author(s): @rkoster
+- Status: Draft
+- RFC Pull Request: (fill in with PR link after you submit it)
+
+
+## Summary
+
+This RFC proposes implementing full mutual TLS (mTLS) for container-to-container (C2C) traffic in Diego, enabling applications to both authenticate themselves and verify the identity of connecting applications.
+
+The approach introduces a **new HTTP-based listener on port 62443** that runs alongside the existing TCP-based C2C port (61443), providing a dual opt-in model for both operators and application authors. This enables zero-trust networking where all app-to-app traffic can be authenticated and authorization decisions can be made based on app/space/org identity.
+
+
+## Problem
+
+Cloud Foundry's Diego runtime currently runs Envoy as a per-container sidecar for handling **inbound traffic only**. While this provides TLS termination and route integrity, several limitations prevent true zero-trust app-to-app communication:
+
+### No Client-Side (Egress) Proxy
+
+Envoy is not configured as an egress proxy. Applications cannot automatically inject their instance identity certificate when making outbound connections. The current workaround requires applications to manually use `CF_INSTANCE_CERT` and `CF_INSTANCE_KEY` environment variables to present their identity.
+
+### No Identity Extraction on Server Side
+
+Even when mTLS is enabled, the receiving application has no way to know who connected. Envoy does not forward client certificate information via headers (no XFCC header configuration).
+
+### C2C Port (61443) is TCP-Based
+
+The C2C listener (port 61443) is a TCP proxy, not an HTTP proxy. This means:
+- No HTTP connection manager
+- No XFCC header support
+- No mTLS client certificate validation
+- No HTTP/2 support
+
+### Existing mTLS Only Affects Route Integrity
+
+The existing BOSH property `containers.proxy.require_and_verify_client_certificates` only affects route integrity ports (Gorouter → App), not C2C traffic.
+
+### Current Architecture
+
+```mermaid
+flowchart LR
+    subgraph Container
+        Envoy["Envoy\n(port 61001)"]
+        App["App\n(port 8080)"]
+        Envoy -->|plain TCP| App
+    end
+    Gorouter -->|TLS| Envoy
+    
+    Note["Envoy only handles INBOUND traffic\n(DownstreamTlsContext)"]
+```
+
+### Current Certificate Setup
+
+Diego generates two separate certificates per container:
+
+| Certificate | SAN Contents | Purpose | mTLS Support | HTTP/2 |
+|------------|--------------|---------|--------------|--------|
+| Instance Identity | Container IP address | Route integrity (Gorouter → App) | Yes (optional) | Yes |
+| C2C Certificate | Internal route hostnames (e.g., `app.apps.internal`) | Container-to-container TLS | No | No |
+
+Both certificates include app identity in the `OrganizationalUnit` field with claims like `app:some-app-guid`, `space:some-space-guid`, and `organization:some-org-guid`.
+
+
+## Proposal
+
+### Design Principles
+
+1. **Backwards compatibility**: Keep port 61443 unchanged for existing applications
+2. **Dual opt-in**: Both operator and application author must opt-in
+3. **New port for new behavior**: Port 62443 for HTTP-based mTLS with XFCC header support
+
+### Port Allocation
+
+| Port | Type | Purpose | mTLS | XFCC | Status |
+|------|------|---------|------|------|--------|
+| 61001+ | HTTP | Route integrity | Optional | No | Existing |
+| 61443 | TCP | Legacy C2C TLS | No | No | Existing (unchanged) |
+| **62443** | **HTTP** | **C2C mTLS with identity** | **Required** | **Yes** | **New** |
+| 61445 | HTTP | Egress proxy (client-side) | N/A | N/A | New |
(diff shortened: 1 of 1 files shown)


toc/rfc/rfc-draft-diego-c2c-mtls.md
```diff
@@ -0,0 +1,271 @@
...
+```
+CF_INSTANCE_MTLS_PROXY=http://127.0.0.1:61445
+```
+
+**Enforcing Proxy Usage**: Operators can make the egress proxy mandatory for all applications by setting `HTTP_PROXY` and `HTTPS_PROXY` to `http://127.0.0.1:61445` using [running environment variable groups](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#evgroups). This ensures all HTTP/HTTPS traffic from applications is routed through the egress proxy, with instance identity certificates injected only for C2C mTLS backends.
+
+**BOSH Properties**:
+
+```yaml
+containers.proxy.enable_egress_proxy:
```
plowin: Hey @rkoster, thx a lot for raising this RFC! I was wondering if the egress-part could also be enabled independent of the c2c-parts in this RFC. So, when only enabling `enable_egress_proxy` and an app-developer adding the Env `CF_INSTANCE_MTLS_PROXY=http://127.0.0.1:61445` without the properties from step 1 and 3, would the egress proxy add the instance-cert in the mTLS connection towards gorouter (or a TLS-terminating component in front of gorouter)?

In our setup, we do not enable c2c as `network.write` is not exposed to the app developers. Yet, there is demand for platform features to support authentication from AppA to AppB.
We might consider implementing a [per-route feature as per rfc-0027](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0027-generic-per-route-features.md) in gorouter to even parse the client-cert data on platform-side, without additional logic within each app.

toc/rfc/rfc-draft-diego-c2c-mtls.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### Result
+
+When an operator enables `containers.proxy.enable_c2c_mtls_listener`:
+- Port 62443 becomes available on all containers
+- Clients connecting to 62443 MUST present a valid instance identity certificate
+- The receiving app gets an `X-Forwarded-Client-Cert` header with caller identity:
+
+```
+X-Forwarded-Client-Cert: Hash=abc123;Subject="CN=instance-guid,OU=app:client-app-guid,OU=space:space-guid"
```
maxmoehl: The haproxy-boshrelease already includes code to handle [XFFC headers](https://github.com/cloudfoundry/haproxy-boshrelease/blob/51a0e09d55da05e4c571a48be01cb47416729673/jobs/haproxy/templates/haproxy.config.erb#L631-L649), for foundations that already run with HAProxy we should consider how these should interact. Having both use the same header names but using different formats seems like a bad idea. My first thought was to add another header that is set by envoy to indicate that this is app2app traffic, like `cf-app-to-app: 1`. This is then used to select the right context in which to evaluate the provided metadata.

toc/rfc/rfc-draft-diego-c2c-mtls.md
```diff
@@ -0,0 +1,271 @@
...
+  example: ["apps.internal.", "my.apps.internal."]
+  default: []
+```
+
+This follows the same naming and format as the [`internal_domains` property in `bosh-dns-adapter`](https://github.com/cloudfoundry/cf-networking-release/blob/develop/jobs/bosh-dns-adapter/spec).
+
+#### Considerations
+
+- **SNI Handling**: Envoy needs to extract the target hostname for proper TLS handshake
+- **NO_PROXY**: Applications should configure `NO_PROXY` for traffic that should not go through the proxy
```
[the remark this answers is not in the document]
rkoster: Say you need to do app to app, but also call an external service, then the external service could be in NO_PROXY. There is nothing done for this functionality in this RFC. This is just standard behaviour most http client libraries have: https://wikitech.wikimedia.org/wiki/HTTP_proxy.

toc/rfc/rfc-draft-diego-c2c-mtls.md
```diff
@@ -0,0 +1,271 @@
...
+
+**UpstreamTlsContext**: Configure the egress proxy to attach the instance identity certificate only when connecting to C2C mTLS backends, identified by destination port (62443) and hostname pattern matching the configured `internal_domains`.
+
+**Environment Variable**: Inject a CF-specific environment variable for applications to discover the egress proxy:
+
+```
+CF_INSTANCE_MTLS_PROXY=http://127.0.0.1:61445
+```
+
+**Enforcing Proxy Usage**: Operators can make the egress proxy mandatory for all applications by setting `HTTP_PROXY` and `HTTPS_PROXY` to `http://127.0.0.1:61445` using [running environment variable groups](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#evgroups). This ensures all HTTP/HTTPS traffic from applications is routed through the egress proxy, with instance identity certificates injected only for C2C mTLS backends.
```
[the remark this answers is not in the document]
rkoster: There is no enforcement, using the proxy is just and easier way of setting things up for a whole range of applications which already support proxy related variables. But apps can also configure their client themselves using the client cert. The enforcement of mtls, would be on the network policy front. Given that an app could be configured to only allow container to container traffic on the static mtls port.

toc/rfc/rfc-draft-diego-c2c-mtls.md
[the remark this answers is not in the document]
maxmoehl: Ah, I didn't know about that :)

toc/rfc/rfc-draft-diego-c2c-mtls.md
chombium: Hi @rkoster, I really like this idea. I only wonder if we implement this, how can we do another iteration to support bring your own certs for apps for c2c networking, where the developers specify ca, cert and key for both communicating parties. Have you maybe done some thoughts about that?

toc/rfc/rfc-draft-diego-c2c-mtls.md
```diff
@@ -0,0 +1,271 @@
...
+```
+
+#### Changes Required
+
+**New Port Constant**: Add `C2CMTLSPort = 62443` alongside the existing `C2CTLSPort = 61443`.
+
+**New HTTP-based Listener**: Create a new Envoy listener with:
+- HTTP connection manager (not TCP proxy)
+- `DownstreamTlsContext` with `RequireClientCertificate: true`
+- Validation context trusting the instance identity CA
```
mariash: what certificate will be presented by the server? same as C2C? maybe should be stated in RFC.

toc/rfc/rfc-draft-diego-c2c-mtls.md
```diff
@@ -0,0 +1,271 @@
...
+- Author(s): @rkoster
+- Status: Draft
+- RFC Pull Request: [community#1437](https://github.com/cloudfoundry/community/pull/1437)
+
+
+## Summary
+
+This RFC proposes implementing full mutual TLS (mTLS) for container-to-container (C2C) traffic in Diego, enabling applications to both authenticate themselves and verify the identity of connecting applications.
+
+The approach introduces a **new HTTP-based listener on port 62443** that runs alongside the existing TCP-based C2C port (61443), providing a dual opt-in model for both operators and application authors. This enables zero-trust networking where all app-to-app traffic can be authenticated and authorization decisions can be made based on app/space/org identity.
```
mariash: This RFC enables zero-trust networking only for HTTP traffic and not TCP, so not all app-to-app traffic.

mariash (changes requested):

beyhan:
@rkoster do you plan to close this RFC as the  #1481 has been accepted?

PR #1453, opened 2026-03-16
RFC: BOSH-Provided Dynamic Disks via Volume Services

rkoster:
This PR adds the RFC "BOSH-Provided Dynamic Disks via Volume Services".

For easier viewing, you can see [the full RFC](https://github.com/rkoster/community/blob/rfc-bosh-dynamic-disks-volume-services/toc/rfc/rfc-draft-bosh-dynamic-disks-via-volume-services.md) as preview.

## Summary

This RFC proposes a mechanism for BOSH to provide IaaS-managed persistent disks to Diego containers through the existing volume services architecture.

- **BOSH Director** gains a `permissions` model controlling which instance groups may create, attach, detach, and delete disks
- **BOSH Agent** gains subcommands that relay disk requests to the Director over NATS
- **Volume driver** implements Docker Volume Plugin v1.12, translating volume operations into Agent disk commands

Diego's volman discovers this driver automatically — **zero changes to Diego, CAPI, or the CF CLI**.

This is a **foundation technology** enabling Cloud Foundry to run stateful single-container workloads — agentic coding sessions, cloud-based developer environments, and long-running AI agent processes.

## Key Design Points

- No Director credentials in workloads — Agent uses existing NATS mTLS
- Director-enforced permissions declared in deployment manifest
- Disk CIDs flow through Cloud Controller service bindings (same as NFS/SMB)
- VM-level locking allows dynamic disk ops during `bosh deploy`

### DIFF
--- a/toc/rfc/rfc-draft-bosh-dynamic-disks-via-volume-services.md
+++ b/toc/rfc/rfc-draft-bosh-dynamic-disks-via-volume-services.md
@@ -9,39 +9,53 @@
 
 ## Summary
 
-Enable Diego containers to use IaaS-managed persistent disks through the existing volume services architecture:
+Enable Diego containers to use IaaS-managed persistent disks through a **secure-by-default** architecture that leverages BOSH's existing per-instance identity:
 
-- **BOSH Director** gains a `permissions` model controlling which instance groups may create, attach, detach, and delete disks
-- **BOSH Agent** gains subcommands that relay disk requests to the Director over NATS
-- **Volume driver** implements Docker Volume Plugin v1.12, translating volume operations into Agent disk commands
+- **No credential distribution** — The volume driver calls the local BOSH Agent binary, which relays requests to the Director over its existing mTLS NATS connection. No UAA tokens, no API credentials on cells.
+- **Per-instance blast radius** — Each Agent has a unique certificate. A compromised cell can only perform operations on itself (`disk.self.attach`), not on other cells. No lateral movement.
+- **Manifest-driven permissions** — Operators declare capabilities in the deployment manifest. The Director enforces them at runtime. No external credential management.
 
-Diego's volman discovers this driver automatically — **zero changes to Diego, CAPI, or the CF CLI**.
+The architecture has four layers:
 
-This is a **foundation technology** enabling Cloud Foundry to run stateful single-container workloads — agentic coding sessions, cloud-based developer environments, and long-running AI agent processes — by giving Diego containers access to dedicated IaaS block storage managed by BOSH.
+1. **BOSH Director** — `permissions` model controlling disk lifecycle operations per instance group
+2. **BOSH Agent** — subcommands relaying disk requests to the Director
+3. **Volume driver** — Docker Volume Plugin v1.12, collocated on Diego cells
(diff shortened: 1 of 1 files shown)


rkoster:
Thanks @mariash — since rfc-0053 merged I've built a proof-of-concept to validate the centralized Disk Provisioner architecture. Repository: https://github.com/rkoster/cf-bosh-disk-broker-release

The POC implements the three CF components from rfc-0053 (Disk Provisioner, Service Broker, Volume Driver) as a single BOSH release. End-to-end it works: `cf create-service` → `cf bind-service` → app start → IaaS disk created and attached → filesystem formatted → bind-mounted into the container. Data persists across restarts and redeployments.

The most important thing the POC clarifies is architectural, directly addressing this concern:

> Could you please specify why NATS communication is preferred over an existing Director HTTP API model? The main concern I have is security, it opens a back door on Diego cells to issue IaaS commands to Director.

You're right that disk orchestration should not originate from Diego cells, and the POC makes the reason concrete. Disk orchestration needs to live in a component with a whole-picture view of LRP state. When an ActualLRP transitions to `CLAIMED`, the BBS event carries the `cell_id` of the cell where Diego placed it. The centralized Disk Provisioner uses that to direct the Director: attach this disk to *that* cell. The volume driver on the cell has no equivalent information — it only sees the `Mount` call in front of it; it does not know which cell it's on relative to the rest of the cluster, whether another instance of the same app holds the disk, or what the global scheduling state is. This is a fundamental constraint of the Docker Volume Plugin protocol. Disk orchestration in the driver is not just a security problem; it's an information problem.

**On auth model:** The POC uses UAA credentials for the Disk Provisioner → Director calls, following rfc-0053's current auth model. This was a deliberate scope decision — the goal was to validate the orchestration architecture, not to implement a new auth mechanism. UAA credentials still require operators to create and manage a client, distribute the secret, and rotate it. The mTLS path (rfc-0053's noted future improvement, and the subject of the agreed follow-on RFC) eliminates that entirely: the Disk Provisioner authenticates to the Director using its BOSH-issued instance certificate, and the Director authorizes against manifest-declared permissions — no credential management, same deployment workflow operators already use for everything else.

On your suggestion:

> I suggest having a separate RFC for the Director permissions model you proposed in this RFC.

That's exactly the plan. I'll update this PR to focus on: (1) the mTLS client-cert auth on the Director HTTP API with the manifest permissions model, and (2) the two Diego changes the POC shows are still missing:

- **`InstanceIndexedVolumeIds`** — volman has no support for the instance-indexed suffix. Without it, all instances of a scaled app share one disk, which only works when they land on the same cell.
- **Stop-first evacuation** — Rep calls `requestAuction()` before the evacuating container stops. The replacement instance tries to attach the disk before the evacuating cell has released it.

These gaps exist regardless of auth model and affect any block device integration with Diego.

On @nicolasbender's March 27 question about certificate identification: the BOSH Agent's NATS client certificate is issued per-instance at deploy time and its Subject/SAN encodes deployment name, instance group, and instance ID. For the Director HTTP API auth path, the same certificate is presented on the HTTPS connection — the Director extracts the identity from the TLS client cert, resolves the instance group in the manifest, and checks the declared permissions. No new PKI infrastructure needed. This is the model the follow-on RFC will specify.

PR #1480, opened 2026-04-22
Fix RFC numbers

beyhan:
I manually assigned numbers to some PRs because I didn't generate merge commits, which our automation requires. In doing so, I missed adding leading zeros to the numbers. When the last RFC number was generated by the automation, it didn't recognize the numbers with missing zeros and assigned an incorrect number. I'm correcting this with this PR.


rkoster (approved):

beyhan:
As this is about approved RFCs and already approved numbers and small number fix I'm going to merge this.

PR #1497, opened 2026-05-08
[RFC] Deprecate Windows support in Binary Buildpack

beyhan:
[Better View](https://github.com/cloudfoundry/community/blob/rfc-deprecate-windows-support-binary-buildpack/toc/rfc/rfc-draft-deprecate-and-remove-windows-support-in-binary-buildpack.md)

### DIFF
--- a/toc/rfc/rfc-draft-deprecate-and-remove-windows-support-in-binary-buildpack.md
+++ b/toc/rfc/rfc-draft-deprecate-and-remove-windows-support-in-binary-buildpack.md
@@ -1,0 +1,59 @@
+# Meta 
+[meta]: #meta 
+- Name: Deprecate and Remove Windows Support in Binary Buildpack 
+- Start Date: 2025-01-15 
+- Author(s): @tnikolova82, @ivanovac, @beyhan 
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (fill in with PR link after you submit it)
+- Related RFCs: [RFC-0050 Java Buildpack Migration to Golang](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0050-java-buildpack-migration-to-golang.md) 
+ 
+## Summary 
+ 
+Deprecate and immediately remove Windows support from the Cloud Foundry Binary Buildpack. [`v1.1.21`](https://github.com/cloudfoundry/binary-buildpack-release/releases/tag/v1.1.21) will be the last release shipping Windows support. The next release will be Linux-only. 
+ 
+## Problem 
+ 
+- The Windows support in the Binary Buildpack is effectively unmaintained following the departure of its core maintainer from the CF community. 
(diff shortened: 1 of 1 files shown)


beyhan:
This RFC was discussed during the TOC meeting on May 19th. The following points were raised:

the [hwc-buildpack](https://github.com/cloudfoundry/hwc-buildpack) faces the same maintenance challenges as the Binary Buildpack and should be included in this RFC. Without maintained Windows buildpacks, Cloud Foundry’s Windows support cannot be properly validated and is at risk of deprecation.


## Proposed Solution

There is consensus around maintaining one buildpack for validation purposes of Windows support. The Binary Buildpack is the preferred option as it requires less effort compared to the hwc-buildpack. This solution is dependent on community involvement and commitment.

## Questions for the Community

1. Is there genuine interest and need for Windows support in Cloud Foundry?
2. If yes: Are you willing to actively maintain the [binary-buildpack](https://github.com/cloudfoundry/binary-buildpack)?

## Next Steps

- **If community interest exists**: The TOC will help establish the maintenance structure.
- **If no community involvement materializes**: We will proceed with the deprecation of all Windows-based buildpacks and initiate an RFC for deprecating Windows support in Cloud Foundry itself.

Gerg:
Given Beyhan's above comment, it seems like this RFC is a de-facto "Deprecate Windows support for Cloud Foundry". It doesn't really make sense to maintain the Windows runtime interface, if we don't have a way to exercise the interface via buildpacks.

My impression is that the Binary Buildpack is very simple, even for the Windows distribution. The difficulties come from validating the Windows distribution.

We do currently have a fair bit of Windows infrastructure and knowledge in the Bosh and Diego working group areas, so maybe there is an opportunity to leverage that.

CF on Windows is a differentiating feature and, at least as of a couple years ago, there were a number of different consumers in the community. It seems a shame to throw it away, especially since it more-or-less "just works".

I propose that instead of deprecating Windows support for the binary buildpack, we instead establish a new "cross-platform buildpacks" working group area with a combination of current buildpack maintainers and Windows specialists from other working groups. There may be potential for sharing infrastructures with existing teams for buildpack validation.

Gerg:
This is outside the current scope of this RFC, but for the HWC buildpack, we could move that to a "windows builpacks" working group area, or deprecate the buildpack.

On one hand, the Windows binary buildpack should be sufficient to exercise the Windows runtime interface. On the other hand, there may still be active community consumers of the HWC buildpack who will miss it. As a data point, we don't currently use the HWC buildpack within Broadcom.

beyhan:
Let's continue the discussion on this during the next TOC meeting scheduled for 2nd of June. It will be great if we could have an understanding about what is required to validate the Windows support from an application developer perspective. If the binary-buildpack is "too" simple for that and hwc-buildpack is not used I'm wondering what could be a validation option in this case.

wayneeseguin:
These two buildpacks on Windows are critical for one of our customers so we need to discuss the path forward and what it would look like for us to help maintain the buildpacks and pipelines.

beyhan:
After discussions [RFC to continue](https://github.com/cloudfoundry/community/pull/1515) the support of the Windows based buildpacks has been opened and approved. We are closing this in favor of the accepted RFC.

PR #1501, opened 2026-05-19
RFC: AWS-LC as TLS Backend for HAProxy

hoffmaen:
Link to the document for quick review: [rfc-draft-haproxy-awslc.md](https://github.com/sap-contributions/cf-community/blob/rfc-haproxy-awslc/toc/rfc/rfc-draft-haproxy-awslc.md)

### Summary

This PR adds the RFC for [AWS-LC](https://github.com/aws/aws-lc) as TLS Backend for the [haproxy-boshrelease](https://github.com/cloudfoundry/haproxy-boshrelease).

AWS-LC is independent of the version provided via the BOSH stemcell and can be kept up to date interdependently. The OpenSSL that is currently used via the stemcell is on an outdated version that has severe performance impacts in high-load scenarios. These would be improved in newer OpenSSL versions, but not resolved. HAProxy explicitely recommends against OpenSSL in production ([haproxy/haproxy#3086](https://github.com/haproxy/haproxy/issues/3086)).

AWS-LC alleviates these issues and is proposed to be added as an optional release. For operators nothing changes, unless they want to opt into using AWS-LC. 

AWS-LC is [FIPS 140-3 compliant](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4816).

### DIFF
--- a/toc/rfc/rfc-draft-haproxy-awslc.md
+++ b/toc/rfc/rfc-draft-haproxy-awslc.md
@@ -9,13 +9,13 @@
 
 ## Summary
 
-This proposal applies to HAProxy as deployed by haproxy-boshrelease, which Cloud Foundry uses as the HTTPS ingress in front of the Gorouter. It is **not** about the TCP Router shipped with cf-deployment, which is a separate component with a different role.
+**Note:** This proposal applies to HAProxy as deployed by haproxy-boshrelease, which Cloud Foundry uses as the HTTPS ingress in front of the Gorouter. It is **not** about the TCP Router shipped with cf-deployment, which is a separate component with a different role.
 
 Today, HAProxy is linked against the OpenSSL version provided by the BOSH stemcell. In practice that currently means OpenSSL 3.0.2 — significant TLS performance improvements have landed in OpenSSL 3.5, but adopting them requires a stemcell that ships OpenSSL 3.5, which is outside this release's control.
 
 Extend HAProxy with [AWS-LC](https://github.com/aws/aws-lc) as an optional TLS backend in Cloud Foundry's haproxy-boshrelease. AWS-LC is an open-source (Apache 2.0) cryptographic library maintained by AWS, based on BoringSSL, with a [FIPS 140-3 validated module](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4816).
 
-Load testing demonstrates that AWS-LC delivers **43% lower CPU usage** and **65% higher peak TLS connection rate** compared to OpenSSL 3.0 for TLS handshakes. HAProxy upstream explicitly recommends against OpenSSL in production ([haproxy#3086](https://github.com/haproxy/haproxy/issues/3086)).
+Load testing demonstrates that AWS-LC delivers **43% lower CPU usage** and **65% higher peak TLS connection rate** compared to OpenSSL 3.0 for TLS handshakes. Under FIPS, the gap widens further to **39% lower CPU** and **109% higher peak rate** (OpenSSL FIPS vs AWS-LC FIPS). HAProxy upstream explicitly recommends against OpenSSL in production ([haproxy#3086](https://github.com/haproxy/haproxy/issues/3086)).
 
 The change is non-breaking: OpenSSL remains the default, and AWS-LC is offered as an opt-in variant alongside existing releases.
 
@@ -23,33 +23,23 @@ The change is non-breaking: OpenSSL remains the default, and AWS-LC is offered a
 
 HAProxy terminates all inbound TLS connections for Cloud Foundry. The TLS handshake is the most CPU-intensive operation on the ingress data path. Under connection-heavy workloads — rolling deployments, autoscaling events, short-lived connections — OpenSSL 3.0 becomes the bottleneck.
 
-### Measured Performance Gap
-
-Load testing on representative CF infrastructure (HAProxy 3.2.16, TLS 1.3, `oha --disable-keepalive` isolating handshake performance):
-
-| Metric | OpenSSL 3.0 | AWS-LC | Improvement |
-|--------|-------------|--------|-------------|
-| Peak new TLS connections/s (CPU saturated) | ~5,000 | ~8,100 | **+65%** |
-| CPU at 5,000 conn/s (c=20) | 86% | 49% | **-43%** |
-| CPU at 4,000 conn/s (c=40) | 62% | 44% | **-29%** |
-| p95 latency at 5,000 conn/s | 5.1ms | 3.5ms | **-32%** |
-| p95 latency at saturation (10k target) | 23.6ms | 11.2ms | **-53%** |
-
-With keepalive (no handshake per request), throughput is identical (~45k req/s). The difference is entirely in the **TLS handshake path**.
-
 ### Production Impact
 
-In real Cloud Foundry deployments the OpenSSL bottleneck appears at much lower connection rates than the synthetic numbers above suggest. The HAProxy maintainers attribute this to OpenSSL's poor performance under realistic ingress workloads — particularly when HAProxy also opens encrypted connections to backends, which doubles the TLS work per request ([haproxy#3086](https://github.com/haproxy/haproxy/issues/3086)). With AWS-LC we expect HAProxy to handle load peaks substantially better than what the synthetic benchmarks alone would predict.
+In real Cloud Foundry deployments the OpenSSL bottleneck appears at lower connection rates than synthetic benchmarks alone would suggest. The HAProxy maintainers attribute this to OpenSSL's poor performance under realistic ingress workloads — particularly when HAProxy also opens encrypted connections to backends, which doubles the TLS work per request ([haproxy#3086](https://github.com/haproxy/haproxy/issues/3086)). With AWS-LC we expect HAProxy to handle load peaks substantially better than what the synthetic benchmarks alone would predict.
 
-- **Connection storms**: During rolling deployments, all clients reconnect simultaneously.
-- **Tail latency**: Under load, OpenSSL's p95 grows to 23.6ms while AWS-LC stays at 11.2ms.
-- **Scaling costs**: The 43% CPU reduction at typical operating rates eliminates unnecessary horizontal scaling.
+- **Connection storms**: During rolling deployments and client reconnection waves, large numbers of new TLS handshakes arrive simultaneously and saturate the ingress.
+- **Tail latency**: Under load, OpenSSL's p95 handshake latency grows steeply while AWS-LC remains stable.
+- **Scaling costs**: The CPU reduction at typical operating rates eliminates unnecessary horizontal scaling and frees headroom for traffic spikes.
 
-### Upstream Recommendation
+### HAProxy Upstream Recommendation
 
-HAProxy maintainers explicitly recommend against OpenSSL, citing performance issues and architectural incompatibilities with OpenSSL 3.x's provider dispatch model ([haproxy#3086](https://github.com/haproxy/haproxy/issues/3086)). HAProxy ships optimized code paths for AWS-LC via the `USE_OPENSSL_AWSLC` build flag.
+HAProxy maintainers explicitly recommend against OpenSSL, citing performance issues and architectural incompatibilities with OpenSSL 3.x's provider dispatch model ([haproxy#3086](https://github.com/haproxy/haproxy/issues/3086)). HAProxy ships optimized code paths for AWS-LC via the `USE_OPENSSL_AWSLC=1` build flag, and HAProxy Technologies publishes a [State of SSL Stacks](https://www.haproxy.com/blog/state-of-ssl-stacks) review that recommends AWS-LC and ships [official Community Performance Packages compiled with AWS-LC](https://www.haproxy.com/company/news/haproxy-technologies-announces-availability-of-haproxy-community-performance-packages-compiled-with-aws-lc).
+
+Other TLS libraries supported by HAProxy were considered and rejected: WolfSSL requires a commercial production license, while upstream BoringSSL has no stable release cadence and no FIPS certification. AWS-LC is the only option that combines an open-source license, stable releases, FIPS 140-3 validation, and an HAProxy-optimized integration path.
 
 ## Proposal
+
+Extend the haproxy-boshrelease build matrix: keep the existing OpenSSL-based release as the default, and additionally ship slim single-variant releases for AWS-LC and AWS-LC FIPS — each available with and without custom HAProxy patches. In addition, ship one multi-variant release that bundles all six HAProxy binaries together and lets operators switch between TLS backends at deploy time via a BOSH property, without recompiling.
 
 ### Release Variants
 
@@ -67,13 +57,13 @@ The naming schema treats OpenSSL as the standard: the existing release keeps its
 | `awslc-fips-patched` | AWS-LC FIPS | FIPS with custom HAProxy patches |
 | `multi` | All of the above | Runtime-switchable, for migration |
 
-### Single-Variant Releases
+#### Single-Variant Releases
 
 Each single-variant release contains one monolithic `haproxy` package with only the blobs required for that specific TLS backend. The packaging script auto-detects which library to build based on the included blobs.
 
 These releases are **slim and deploy quickly**: only the necessary dependencies are compiled, and the release tarball contains no unused components. A typical OpenSSL deployment compiles in ~15 minutes; AWS-LC in ~25 minutes (including cmake).
 
-### Multi-Variant Release
+#### Multi-Variant Release
 
 The multi release bundles all six HAProxy binaries into a single deployment. A BOSH property selects the active variant at runtime:
 
@@ -116,6 +106,32 @@ The job wrapper resolves the binary path at startup:
 
 This ensures backward compatibility: existing deployments that use the single-variant `openssl` release continue to work without manifest changes.
 
+### Measured Performance Gap
+
+The tables below quantify the production scenarios described in the Problem section. Load testing on representative CF infrastructure (HAProxy 3.2.16, TLS 1.3, `oha --disable-keepalive` isolating handshake performance):
+
+| Metric | OpenSSL 3.0 | AWS-LC | Improvement |
+|--------|-------------|--------|-------------|
+| Peak new TLS connections/s (CPU saturated) | ~5,000 | ~8,100 | **+65%** |
+| CPU at 5,000 conn/s (c=20) | 86% | 49% | **-43%** |
+| CPU at 4,000 conn/s (c=40) | 62% | 44% | **-29%** |
+| p95 latency at 5,000 conn/s | 5.1ms | 3.5ms | **-32%** |
+| p95 latency at saturation (10k target) | 23.6ms | 11.2ms | **-53%** |
+
+With keepalive (no handshake per request), throughput is identical (~45k req/s). The difference is entirely in the **TLS handshake path**.
+
+### FIPS Performance Gap
+
+The same benchmark was repeated with FIPS-validated builds (OpenSSL 3.0.2 with the FIPS provider vs AWS-LC FIPS 3.3.0). Under FIPS, the gap is roughly twice as large as in the non-FIPS case:
+
+| Metric | OpenSSL FIPS | AWS-LC FIPS | Improvement |
+|--------|--------------|-------------|-------------|
+| Peak new TLS connections/s (CPU saturated) | ~3,500 | ~7,500 | **+109%** |
+| CPU at 2,000 conn/s | 49% | 30% | **-39%** |
+| p95 latency at 5,000 conn/s (c=20) | 11.5ms | 3.0ms | **-73%** |
+
+OpenSSL FIPS saturates at less than half the connection rate of AWS-LC FIPS. This is consistent with OpenSSL 3.0's provider dispatch model: every cryptographic operation pays an indirection cost that the FIPS provider compounds with additional self-test bookkeeping. AWS-LC's FIPS module is integrated directly into its TLS implementation without that overhead, so customers requiring FIPS compliance pay the *largest* performance penalty under OpenSSL — and gain the most from migrating to AWS-LC.
+
 ### Cipher and Curve Differences
 
 AWS-LC (both FIPS and non-FIPS) differs from OpenSSL in TLS cipher availability:
@@ -131,15 +147,6 @@ In practice this has no client impact: DHE is rarely required by modern clients,
 **Compatibility note**: A client would only break on AWS-LC if it supports **DHE but not ECDHE**. Such stacks predate ~2010 and also predate TLS 1.2. Under a TLS 1.2-minimum policy (already enforced in this deployment), they are filtered out at the protocol layer regardless of cipher selection — every client that can negotiate TLS 1.2 also supports ECDHE. Clients that currently happen to negotiate DHE will transparently fall back to ECDHE with no visible impact.
 
 Recommended curve configuration for AWS-LC: `X25519:P-256:P-384:P-521`
-
-### FIPS Verification
-
-The AWS-LC FIPS module is built from a frozen, certified source tree using a fixed compiler toolchain. To produce a FIPS-compliant binary, AWS-LC's build process runs a `delocate`/`inject_hash` step that rewrites assembly and embeds an integrity hash of the module into the resulting library. This step is implemented in Go, so a Go toolchain (~200MB) must be available at compile time. Go is only a build-time dependency — it is cleaned up after the FIPS library is built and is not part of the runtime.
-
-FIPS mode can be verified at runtime:
-
-- `haproxy -vv` reports "AWS-LC FIPS 3.3.0"
-- Stats socket: `SslFipsModeRuntime: 1`
 
 ### Version Management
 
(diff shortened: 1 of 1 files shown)


beyhan:
Hi @hoffmaen,

thank you for this pr. Is it intentionally in draft status?

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,187 @@
...
+- **Connection storms**: During rolling deployments, all clients reconnect simultaneously.
+- **Tail latency**: Under load, OpenSSL's p95 grows to 23.6ms while AWS-LC stays at 11.2ms.
+- **Scaling costs**: The 43% CPU reduction at typical operating rates eliminates unnecessary horizontal scaling.
+
+### Upstream Recommendation
+
+HAProxy maintainers explicitly recommend against OpenSSL, citing performance issues and architectural incompatibilities with OpenSSL 3.x's provider dispatch model ([haproxy#3086](https://github.com/haproxy/haproxy/issues/3086)). HAProxy ships optimized code paths for AWS-LC via the `USE_OPENSSL_AWSLC` build flag.
+
+## Proposal
+
```
Soha-Albaghdady, line 61: I would suggest to add 2 lines to describe what we propose "support multiple variants, choose between them using config property" before going into the variants

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,187 @@
...
+- Name: AWS-LC as TLS Backend for HAProxy
+- Start Date: 2026-05-06
+- Author(s): @hoffmaen
+- Status: Draft
+- RFC Pull Request:
+- Affected Component(s): haproxy-boshrelease
+
+## Summary
+
+This proposal applies to HAProxy as deployed by haproxy-boshrelease, which Cloud Foundry uses as the HTTPS ingress in front of the Gorouter. It is **not** about the TCP Router shipped with cf-deployment, which is a separate component with a different role.
```
Soha-Albaghdady: May be add a "Note:" suffix

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Production Impact
+
+In real Cloud Foundry deployments the OpenSSL bottleneck appears at much lower connection rates than the synthetic numbers above suggest. The HAProxy maintainers attribute this to OpenSSL's poor performance under realistic ingress workloads — particularly when HAProxy also opens encrypted connections to backends, which doubles the TLS work per request ([haproxy#3086](https://github.com/haproxy/haproxy/issues/3086)). With AWS-LC we expect HAProxy to handle load peaks substantially better than what the synthetic benchmarks alone would predict.
+
+- **Connection storms**: During rolling deployments, all clients reconnect simultaneously.
+- **Tail latency**: Under load, OpenSSL's p95 grows to 23.6ms while AWS-LC stays at 11.2ms.
+- **Scaling costs**: The 43% CPU reduction at typical operating rates eliminates unnecessary horizontal scaling.
+
+### Upstream Recommendation
```
Soha-Albaghdady: -> HAproxy Upstream recommendation

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,187 @@
...
+|---------|-------------|-------------|
+| `openssl` | System OpenSSL | Default, backward-compatible (no change for existing users) |
+| `openssl-patched` | System OpenSSL | With custom HAProxy patches |
+| `awslc` | AWS-LC | High-performance, non-FIPS |
+| `awslc-patched` | AWS-LC | With custom HAProxy patches |
+| `awslc-fips` | AWS-LC FIPS 3.3.0 | FIPS 140-3 validated |
+| `awslc-fips-patched` | AWS-LC FIPS | FIPS with custom HAProxy patches |
+| `multi` | All of the above | Runtime-switchable, for migration |
+
+### Single-Variant Releases
```
Soha-Albaghdady: The section needs to be indented to the right (under release variants)

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,187 @@
...
+
+Each single-variant release contains one monolithic `haproxy` package with only the blobs required for that specific TLS backend. The packaging script auto-detects which library to build based on the included blobs.
+
+These releases are **slim and deploy quickly**: only the necessary dependencies are compiled, and the release tarball contains no unused components. A typical OpenSSL deployment compiles in ~15 minutes; AWS-LC in ~25 minutes (including cmake).
+
+### Multi-Variant Release
```
Soha-Albaghdady: Same as above

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,187 @@
...
+
+**Compatibility note**: A client would only break on AWS-LC if it supports **DHE but not ECDHE**. Such stacks predate ~2010 and also predate TLS 1.2. Under a TLS 1.2-minimum policy (already enforced in this deployment), they are filtered out at the protocol layer regardless of cipher selection — every client that can negotiate TLS 1.2 also supports ECDHE. Clients that currently happen to negotiate DHE will transparently fall back to ECDHE with no visible impact.
+
+Recommended curve configuration for AWS-LC: `X25519:P-256:P-384:P-521`
+
+### FIPS Verification
+
+The AWS-LC FIPS module is built from a frozen, certified source tree using a fixed compiler toolchain. To produce a FIPS-compliant binary, AWS-LC's build process runs a `delocate`/`inject_hash` step that rewrites assembly and embeds an integrity hash of the module into the resulting library. This step is implemented in Go, so a Go toolchain (~200MB) must be available at compile time. Go is only a build-time dependency — it is cleaned up after the FIPS library is built and is not part of the runtime.
+
+FIPS mode can be verified at runtime:
```
Soha-Albaghdady: Too detailed, can be skipped

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,187 @@
...
+- **Direct route (recommended once confident)**: deploy the slim `awslc` release. No `ssl_variant` property, no multi-binary overhead, no compilation of unused variants.
+- **Safety-net route (recommended for risk-averse migrations)**: deploy the multi release first, set `ssl_variant: awslc`, validate under production traffic, and once confident, switch to the slim `awslc` release. If problems surface during the multi-release phase, flipping `ssl_variant` back to `openssl` and redeploying takes ~2 minutes with no recompilation.
+
+Phased rollout:
+
+1. **Phase 1** (this RFC): ship AWS-LC variants alongside OpenSSL. No change for existing deployments.
+2. **Phase 2** (optional): operators choose the safety-net route via the multi release, **or** jump directly to the slim single-variant `awslc` release.
+3. **Phase 3**: switch the default release shipped by haproxy-boshrelease (the no-suffix slim release) from OpenSSL to AWS-LC. The multi release continues to exist for migrations but is not the long-term target — most production deployments end up on a slim single-variant release.
+
+## Alternatives Considered
```
Soha-Albaghdady, line 238: I would propose to add this in the problem statement to show what option we considered and why we chose AWS-LC. Proposing accordingly to move the performance comparison table to the proposal section.

Soha-Albaghdady (commented):
LGTM, proposing some minor formatting and restructuring

beyhan:
@hoffmaen could you please ask in your WG (meeting) about feedback.

b1tamara (approved):
LGTM

rkoster:
# Code Review

**PR #1501**: RFC: AWS-LC as TLS Backend for HAProxy

## Summary

This RFC proposes adding AWS-LC as an optional TLS backend for the haproxy-boshrelease to address OpenSSL 3.0.2 TLS-handshake performance and the lack of post-quantum key exchange. The problem statement is strong: it is well-evidenced (benchmarks, HAProxy upstream recommendation in haproxy#3086, FIPS 140-3 validation) and the non-breaking, opt-in framing with OpenSSL-as-default is the right compatibility posture. The core AWS-LC value proposition is convincing. The main gap is the packaging design: the RFC introduces a 7-tarball matrix (openssl, openssl-patched, awslc, awslc-patched, awslc-fips, awslc-fips-patched, multi) but only justifies 4 of the 7. The three '-patched' variants are never explained, and there is no analysis of the CI/compilation cost this matrix imposes despite it roughly doubling the build surface. There are also smaller correctness/process issues: the implementation link 404s (the branch lives on a fork as PR #925), the RFC metadata omits fields required by the template, and the document still carries Draft status with an empty RFC Pull Request field.

## Critical Issues - Must Fix

1. **The 7-variant release matrix is under-justified: the three '-patched' variants have no stated rationale**
   - File: `toc/rfc/rfc-draft-haproxy-awslc.md`
   - The Release Variants table defines seven tarballs: openssl, openssl-patched, awslc, awslc-patched, awslc-fips, awslc-fips-patched, and multi. The RFC justifies openssl (backward compat), awslc (performance), awslc-fips (FIPS compliance), and multi (migration safety net). It does NOT justify the '-patched' axis at all. 'custom HAProxy patches' appears in the table and Build Architecture sections but the RFC never states what these patches are, why they are needed, which users need them, or why they must ship as separate first-class release variants rather than being folded into the base builds or omitted. This single unexplained axis doubles the variant count from 3 to 6 and doubles the number of HAProxy binaries in the multi release from 3 to 6. Reviewers cannot evaluate the proposal's scope without knowing what problem the patched variants solve.
   - Fix: Add a subsection explaining the custom HAProxy patches: what they change, why they cannot be upstreamed or made default, and who requires them. Then justify (or drop) each patched variant explicitly. If the patches are not required for the initial rollout, consider deferring all three '-patched' variants to a follow-up and shipping only openssl / awslc / awslc-fips / multi.

2. **No analysis of CI and compilation cost for the expanded build matrix**
   - File: `toc/rfc/rfc-draft-haproxy-awslc.md`
   - The proposal significantly expands what haproxy-boshrelease must compile and test: the multi release alone builds 6 HAProxy binaries plus aws-lc, aws-lc-fips, cmake, and a Go toolchain (confirmed against the implementation's packages-multi directory). Six additional single-variant tarballs are also produced. The RFC gives only two isolated build-time figures (~15 min OpenSSL, ~25 min AWS-LC) and no accounting of the total CI impact: how many release tarballs are built and tested per pipeline run, how compilation time and cost scale across the full matrix, what the storage/artifact footprint is, or how the test matrix (BOSH deployment tests per variant) grows. For a shared community release this recurring cost is a first-order concern, not an implementation detail.
   - Fix: Add a 'Costs' or 'CI / Build Impact' section quantifying the added compilation time, pipeline duration, artifact count/size, and test-matrix growth for the full set of variants. Use this to justify the matrix size, and explicitly weigh it against dropping the patched variants and/or the multi release.

## Important Issues - Should Fix

1. **Implementation branch link is broken (404)**
   - File: `toc/rfc/rfc-draft-haproxy-awslc.md`
   - The References section links to https://github.com/cloudfoundry/haproxy-boshrelease/tree/aws-lc-multi as the implementation branch. That branch does not exist in cloudfoundry/haproxy-boshrelease and the URL returns 404. The implementation actually lives on the fork sap-contributions/haproxy-boshrelease (branch aws-lc-multi), proposed upstream as PR #925 (https://github.com/cloudfoundry/haproxy-boshrelease/pull/925). Reviewers following the link cannot see the code.
   - Suggestion: Update the link to the fork branch (https://github.com/sap-contributions/haproxy-boshrelease/tree/aws-lc-multi) and/or reference the upstream PR cloudfoundry/haproxy-boshrelease#925.

2. **RFC metadata omits required template fields**
   - File: `toc/rfc/rfc-draft-haproxy-awslc.md`
   - The Meta block is missing the 'Related RFCs' field present in rfc-template.md, and the 'RFC Pull Request' field is left empty. rfc-0023-add-cf-supports-for-fips-stemcell may be relevant to the FIPS discussion and worth cross-linking under Related RFCs.
   - Suggestion: Add the 'Related RFCs' field (linking rfc-0023 if applicable, or leaving empty per template) and fill in 'RFC Pull Request' with this PR's URL, as the RFC README process step requires.

3. **Runtime binary selection via filesystem-path probing is described but its failure modes are not**
   - File: `toc/rfc/rfc-draft-haproxy-awslc.md`
   - Runtime Selection resolves the binary by checking whether /var/vcap/packages/haproxy-<ssl_variant> exists, else falling back to /var/vcap/packages/haproxy. The RFC does not specify behaviour when ssl_variant is set to a valid value on a single-variant (non-multi) release where that package is absent: does it silently fall back to the wrong backend, or fail fast? Silent fallback could mask a misconfiguration where an operator believes they are running AWS-LC FIPS but are actually on OpenSSL.
   - Suggestion: Specify the validation/error behaviour when ssl_variant names a variant not present in the deployed release. Prefer failing fast with a clear error over silent fallback, and state this in the RFC.

4. **No testing strategy for the variants, and the reference implementation adds zero test coverage**
   - File: `toc/rfc/rfc-draft-haproxy-awslc.md`
   - The RFC defines 7 variants with materially different TLS behaviour (no DHE, no X448, no FFDH in AWS-LC; FIPS-mode claims; PQ named groups) and a runtime ssl_variant selector, but says nothing about how any of this is verified. haproxy-boshrelease already has good coverage - rspec template specs and a Go/Ginkgo acceptance suite (https_frontend_test.go, mtls_frontend_test.go, strict_sni_test.go, tcp_frontend_test.go). The linked implementation branch (aws-lc-multi, 36 files changed) modifies NONE of these: zero *_spec.rb and zero *_test.go changes; the only acceptance-tests edits are README.md and run-local.sh. So nothing verifies that awslc negotiates the expected ciphers, that the documented DHE->ECDHE fallback actually works, that SNI/mTLS still pass on the new binaries, that the ssl_variant switch selects the right binary, or that FIPS is genuinely active. These are the exact claims the RFC relies on.
   - Suggestion: Add a Testing section stating which variants receive automated acceptance coverage and what it asserts (cipher/curve negotiation, DHE->ECDHE fallback, SNI + mTLS regression, ssl_variant selection resolves the correct binary, FIPS-active assertion for the FIPS variants), and how that test matrix scales in the pipeline. The implementation PR should extend the existing rspec/acceptance suites accordingly.

5. **cf-deployment integration and testing expectation is undefined**
   - File: `toc/rfc/rfc-draft-haproxy-awslc.md`
   - The only cf-deployment reference is the line-18 disclaimer that this is not about the TCP Router. haproxy-boshrelease is not in the cf-deployment base manifest; it is opt-in via operations/use-haproxy.yml, which pins a single haproxy release version and adds one haproxy instance group. The RFC does not state whether the AWS-LC variants are expected to be exercised at the cf-deployment level - e.g. via a companion ops-file to use-haproxy.yml or an ssl_variant wiring - nor who owns that integration. Without this, it is unclear how an operator using the standard cf-deployment ops-file path would adopt or validate AWS-LC end to end.
   - Suggestion: State the cf-deployment expectation explicitly: either add/describe a companion ops-file (e.g. use-haproxy-awslc.yml) naming which variant it wires and who maintains it, or declare cf-deployment integration testing out of scope with a rationale and point operators at the haproxy-boshrelease-level path instead.

## Minor Issues

1. Filename still uses the rfc-draft- prefix in `toc/rfc/rfc-draft-haproxy-awslc.md`
2. Status is Draft while the PR is out of draft and under active review in `toc/rfc/rfc-draft-haproxy-awslc.md`

## Positive Notes

- Strong, well-evidenced problem statement: benchmark tables for both non-FIPS and FIPS, the HAProxy upstream recommendation (haproxy#3086), and FIPS 140-3 validation together make a compelling case.
- Correct compatibility posture: OpenSSL remains the default and AWS-LC is strictly opt-in, so existing deployments see no change.
- The cipher/curve differences section (no DHE, no X448, no FFDH) is thorough and honestly addresses the main client-compatibility risk, correctly noting TLS 1.2-minimum policy filters out affected legacy clients.
- Including a concrete, phased, reversible migration path with realistic rollback timing (~2 min property switch) shows good operational thinking.
- Alternatives Considered addresses both other TLS libraries (WolfSSL, BoringSSL) and other proxies (Envoy, nginx) with clear reasons for rejection.

## Recommendation

Request changes - critical issues need to be addressed before merging.

rkoster (changes requested):
Requesting changes - see the detailed review comment above. Main blockers: the -patched variant axis is unjustified and there is no CI/build-cost analysis for the expanded matrix. Important: broken implementation link, missing testing strategy (impl branch changes no tests), and undefined cf-deployment integration.

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,201 @@
...
+
+## References
+
+- [HAProxy issue #3086: Recommend against OpenSSL](https://github.com/haproxy/haproxy/issues/3086)
+- [HAProxy Technologies — State of SSL Stacks](https://www.haproxy.com/blog/state-of-ssl-stacks)
+- [HAProxy Technologies — Community Performance Packages compiled with AWS-LC](https://www.haproxy.com/company/news/haproxy-technologies-announces-availability-of-haproxy-community-performance-packages-compiled-with-aws-lc)
+- [AWS-LC GitHub](https://github.com/aws/aws-lc)
+- [AWS-LC FIPS 140-3 certificate #4816](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4816)
+- [oha load testing tool](https://github.com/hatoo/oha)
+- [Implementation branch](https://github.com/cloudfoundry/haproxy-boshrelease/tree/aws-lc-multi)
```
rkoster: This implementation link 404s - the aws-lc-multi branch does not exist in cloudfoundry/haproxy-boshrelease. The code lives on the fork sap-contributions/haproxy-boshrelease (branch aws-lc-multi) and is proposed upstream as PR #925. Please update this link so reviewers can reach the implementation.

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,201 @@
...
+The naming schema treats OpenSSL as the standard: the existing release keeps its current name with no `-openssl` suffix and continues to link against the system-provided OpenSSL from the stemcell. AWS-LC variants carry an explicit `awslc` (and `awslc-fips`) suffix, so existing deployments see no naming change.
+
+| Variant | TLS Backend | Description |
+|---------|-------------|-------------|
+| `openssl` | System OpenSSL | Default, backward-compatible (no change for existing users) |
+| `openssl-patched` | System OpenSSL | With custom HAProxy patches |
+| `awslc` | AWS-LC | High-performance, non-FIPS |
+| `awslc-patched` | AWS-LC | With custom HAProxy patches |
+| `awslc-fips` | AWS-LC FIPS 3.3.0 | FIPS 140-3 validated |
+| `awslc-fips-patched` | AWS-LC FIPS | FIPS with custom HAProxy patches |
```
rkoster, line 77: The '-patched' axis (openssl-patched, awslc-patched, awslc-fips-patched) is introduced here but never explained anywhere in the RFC. What are the 'custom HAProxy patches', why are they needed, and who needs them? This axis doubles the variant count from 3 to 6 (and doubles the multi release's HAProxy binaries), which is the main driver of the added build/CI cost. Please justify each patched variant explicitly, or defer them to a follow-up if they are not required for the initial rollout.

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,201 @@
...
+The multi release avoids redundant compilation through shared packages:
+
+```
+haproxy-deps       → Lua, PCRE2, socat, hatop (compiled once, shared by all 6 variants)
+cmake              → CMake (compiled once, shared by aws-lc and aws-lc-fips)
+aws-lc             → AWS-LC library (non-FIPS)
+aws-lc-fips        → AWS-LC FIPS library (requires Go toolchain for delocate)
+haproxy-openssl    → HAProxy binary linked to system OpenSSL
+haproxy-awslc      → HAProxy binary linked to AWS-LC
+haproxy-awslc-fips → HAProxy binary linked to AWS-LC FIPS
```
rkoster, line 125: This build architecture makes the CI cost concrete: the multi release compiles 6 HAProxy binaries plus aws-lc, aws-lc-fips, cmake, and a Go toolchain, and six additional single-variant tarballs are produced on top. The RFC gives only ~15 min / ~25 min per-variant figures but never totals the pipeline time, artifact count/size, or per-variant BOSH test-matrix growth. For a shared community release this recurring cost is a first-order design concern. Please add a Costs / CI-impact section quantifying it and use it to justify the matrix size.

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+HAProxy is compiled with `USE_OPENSSL_AWSLC=1` for both FIPS and non-FIPS. FIPS is a library property — HAProxy itself does not need separate FIPS logic.
+
+### Runtime Selection
+
+The job wrapper resolves the binary path at startup:
+
+1. If `/var/vcap/packages/haproxy-<ssl_variant>` exists → use variant binary (multi release)
+2. Otherwise → fall back to `/var/vcap/packages/haproxy` (single-variant release)
```
rkoster: What happens on a single-variant release if an operator sets ha_proxy.ssl_variant to a valid-but-absent variant? As written, the fallback to /var/vcap/packages/haproxy could silently run a different TLS backend than the operator intended (e.g. they believe they are on awslc-fips but are actually on openssl). Recommend failing fast with a clear error instead of silent fallback, and stating that behaviour here.

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,201 @@
...
+- **No X448 curve**: Only X25519, P-256, P-384, P-521 are supported for key exchange.
+- **No FFDH groups**: `ffdhe2048`, `ffdhe3072`, etc. are not available as named groups in TLS.
+- **All TLS 1.3 ciphersuites work**: AES-128-GCM, AES-256-GCM, CHACHA20-POLY1305.
+- **ECDHE ciphers for TLS 1.2 work**: All `ECDHE_RSA_*` and `ECDHE_ECDSA_*` suites are available.
+
+In practice this has no client impact: DHE is rarely required by modern clients, and ECDHE provides equivalent or better security with superior performance. Existing cipher configurations that include both DHE and ECDHE ciphers continue to work — HAProxy silently skips DHE entries that AWS-LC cannot fulfill, and clients negotiate ECDHE instead.
+
+**Compatibility note**: A client would only break on AWS-LC if it supports **DHE but not ECDHE**. Such stacks predate ~2010 and also predate TLS 1.2. Under a TLS 1.2-minimum policy (already enforced in this deployment), they are filtered out at the protocol layer regardless of cipher selection — every client that can negotiate TLS 1.2 also supports ECDHE. Clients that currently happen to negotiate DHE will transparently fall back to ECDHE with no visible impact.
```
rkoster: These behavioural claims (DHE absent, clients transparently fall back to ECDHE, existing mixed cipher configs keep working) are exactly the kind of thing that needs an automated test - but the implementation branch (aws-lc-multi) changes no acceptance or rspec tests at all (the existing suite has https_frontend_test.go, mtls_frontend_test.go, strict_sni_test.go). Please add a Testing section covering which variants get automated coverage and what it asserts: cipher/curve negotiation, DHE->ECDHE fallback, SNI + mTLS regression, ssl_variant selecting the correct binary, and a FIPS-active assertion for the FIPS variants. Separately, please clarify the cf-deployment expectation - haproxy is opt-in there via operations/use-haproxy.yml (one pinned release), so state whether a companion ops-file / ssl_variant wiring is planned or whether cf-deployment integration is out of scope.

toc/rfc/rfc-draft-haproxy-awslc.md
```diff
@@ -0,0 +1,201 @@
+# Meta
+[meta]: #meta
+- Name: AWS-LC as TLS Backend for HAProxy
+- Start Date: 2026-05-06
+- Author(s): @hoffmaen
+- Status: Draft
+- RFC Pull Request:
+- Affected Component(s): haproxy-boshrelease
```
rkoster, line 9: Two metadata gaps vs rfc-template.md: the 'Related RFCs' field is missing, and 'RFC Pull Request' is empty. The RFC README process asks you to fill in the PR link. Consider cross-linking rfc-0023-add-cf-supports-for-fips-stemcell under Related RFCs given the FIPS overlap.

b1tamara: Added PR link and `Related RFCs` to the metadata.

toc/rfc/rfc-draft-haproxy-awslc.md
b1tamara: Replaced by the link to Draft PR, as the feature branch on the fork is usually deleted after merge (in case it will be merged)

toc/rfc/rfc-draft-haproxy-awslc.md
hoffmaen: Thanks for the detailed review.
The RFC now has a "Custom HAProxy Patches (the -patched variants)" subsection under Release Variants. It documents what the patch does (relaxes RFC 6455 Sec-WebSocket-Key enforcement in the H1 mux), why it's required in some scenarios (legacy clients opening long-lived WebSocket connections without a spec-compliant key), why it can't be the default (it makes HAProxy intentionally non-compliant with RFC 6455), and why it ships as a separate variant rather than a runtime toggle (BOSH properties are deploy-time, not compile-time, and BOSH won't recompile on a property change, so a compile-time patch toggle isn't expressible as a property). The patch is a long-standing carried requirement, not new to any HAProxy version.

toc/rfc/rfc-draft-haproxy-awslc.md
hoffmaen: Added a "Costs and CI Impact" section. The key correction: a BOSH release tarball only ships source + blobs. The actual HAProxy/AWS-LC compilation happens on the target VM at deploy time, and for the single-variant releases each operator compiles only the one variant they deploy. So extra variants impose no compile cost on operators who don't use them (the multi release is the only thing that compiles everything, and it's explicitly a migration/testing artifact). The incremental cost is confined to the Concourse release-build pipeline: all the extra variants together add a few minutes of pipeline wall-clock plus the disk to store the extra tarballs. Given we cut 9 releases in H1 2026, that recurring cost is negligible. The acceptance-test matrix is per-backend (awslc, awslc-fips), so it doesn't grow with the -patched axis.

toc/rfc/rfc-draft-haproxy-awslc.md
hoffmaen: Regarding cf-deployment integration: Added a short cf-deployment Integration subsection stating this is out of scope. haproxy-boshrelease is opt-in in cf-deployment via use-haproxy.yml (auto-bumped release version); operators adopting AWS-LC manage their own release pinning outside that flow, so no cf-deployment change is required here. The RFC notes that first-class support could later be added via a variant selector in use-haproxy.yml that still tracks automatic version bumps, but that is a cf-deployment-side change and out of scope for this RFC.

hoffmaen: Added a Testing section. The core approach is to run the existing acceptance suite (HTTPS over TLS 1.2/1.3, ALPN, mTLS, strict SNI, WebSocket, + ~20 feature tests) against the AWS-LC-linked binary - since those already drive the full TLS data path, they're the real regression guarantee that swapping the TLS library preserves behaviour, and they directly cover the SNI + mTLS points. The one new test is an rspec spec for the ssl_variant binary selector.
     
We deliberately diverged on a few of the specific asks:
  
* DHE→ECDHE fallback: removed, because the original "fallback" claim was wrong — there is no fallback. Under the shipped default cipher order (ECDHE ahead of DHE, via SSL_OP_CIPHER_SERVER_PREFERENCE), an ECDHE-capable client already negotiates ECDHE, so a client still on DHE has no ECDHE alternative and simply fails on AWS-LC. The RFC now states this accurately, so there's no fallback behaviour to test.
* Per-cipher/curve negotiation tests: not added — the suite's style is functional (drive a real client, assert success), not primitive-level assertions; the differences are documented and bounded by the recommended config instead.
* FIPS-active acceptance run: not added — each acceptance run recompiles the release (~20–30 min for an AWS-LC variant), so a second full run for awslc-fips roughly doubles cost for little signal given FIPS is a library property; FIPS activation is verified out-of-band during release validation. 
  
Note the implementation branch is still draft and will follow the RFC, so the test changes land on the implementation PR rather than being present now.

toc/rfc/rfc-draft-haproxy-awslc.md
hoffmaen: Good catch - the silent fallback is now documented as a configuration error that must fail fast.

hoffmaen:
Regarding draft status, it's under review but not yet approved, also there's no RFC number yet - what should I enter? @rkoster

rkoster (approved):

stephanme:
We start the final comment period with the goal to accept this RFC in the next TOC meeting.

beyhan (approved):

stephanme (approved):

Gerg (approved):

cweibel (approved):

PR #1530, opened 2026-06-24
RFC: ARM64 (aarch64) Architecture Support for Cloud Foundry

vighes:
This RFC proposes adding ARM64 architecture support to Cloud Foundry,
enabling deployments on ARM64-based infrastructure (AWS Graviton, Ampere
Altra, etc.) with 20-40% better price/performance.

A proof-of-concept validated all major CF components (Diego, Garden-runC,
Gorouter, Loggregator, Cloud Controller, UAA, BOSH) compile and execute
on ARM64 with zero source code changes.

Looking forward to community feedback on the proposed workstreams and
open questions.

[Preview link](https://github.com/vighes/community/blob/rfc-arm64-support/toc/rfc/rfc-draft-arm64-support.md)

cc @beyhan

### DIFF
--- a/toc/rfc/rfc-draft-arm64-support.md
+++ b/toc/rfc/rfc-draft-arm64-support.md
@@ -5,7 +5,7 @@
 - **Author(s):** Sachin Vighe
 - **Status:** Draft
 - **Related RFCs:** [rfc-0026-noble-os](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0026-noble-os.md)
-- **Affected Component(s):** bosh-linux-stemcell-builder, Diego, Garden-runC, Gorouter, Loggregator, Cloud Controller, UAA, cf-deployment, buildpacks
+- **Affected Component(s):** bosh-linux-stemcell-builder, bosh-agent, bosh-dns, Diego, Garden-runC, Gorouter, Loggregator, Cloud Controller, UAA, cf-deployment, buildpacks, BPM, NATS
 
 # Summary
 
@@ -53,16 +53,16 @@ Enable ARM64 as a supported architecture through a phased approach, delivering t
 - BOSH Acceptance Tests (BATS) pass on ARM64 stemcells
 - CF Acceptance Tests (CATs) pass on ARM64 deployment
 - Applications can be deployed via standard buildpacks on ARM64 cells
-- Performance ≥ 100% of x86_64 baseline for equivalent instance classes
+- Performance is comparable to x86_64 baseline for equivalent instance classes (validated via benchmarking during Phase 2)
 - Zero impact on existing x86_64 deployments
 
 ## Non-Goals
 
-1. **Not breaking x86_64 support** — existing deployments continue unchanged
-2. **Not requiring CF operators to migrate** — ARM64 is optional, operators opt in
-3. **Not supporting 32-bit ARM** — only 64-bit ARM (aarch64)
-4. **Not requiring homogeneous deployments** — mixed-architecture deployments (ARM64 Diego cells alongside x86_64 control plane) are in scope following the existing Windows cell model with architecture-aware placement
-5. **Not creating provider-specific forks** — all work contributed upstream
+1. **Windows ARM64** — this RFC covers Linux ARM64 only. Windows ARM64 support may be considered separately in the future.
+2. **32-bit ARM** — only 64-bit ARM (aarch64) is in scope
+3. **Requiring operators to migrate** — ARM64 is strictly opt-in. Operators choose to deploy on ARM64 by selecting an ARM64 stemcell; existing x86_64 deployments are unaffected.
+4. **Requiring homogeneous deployments** — mixed-architecture deployments (ARM64 Diego cells alongside x86_64 control plane) are supported, following the existing Windows cell model with architecture-aware placement
+5. **Provider-specific forks** — all work is contributed upstream to CF Foundation repositories
 
 ## Phases
 
@@ -143,11 +143,11 @@ The implementation requires work across multiple CF working groups. The dependen
 ```
 Stemcells (bosh-linux-stemcell-builder)
     ↓
-BOSH Agent (bosh-agent)
+BOSH Agent (bosh-agent) + BOSH DNS
     ↓
 BOSH Director (bosh)
     ↓
-Diego (diego-release) + Garden (garden-runc-release) ← critical path
+Diego (diego-release) + Garden (garden-runc-release)
     ↓
 Cloud Controller (capi-release) + Routing + Logging
     ↓
@@ -155,13 +155,15 @@ cflinuxfs5 + Buildpacks
     ↓
 Applications
 ```
+Each layer depends on the one above it — the entire dependency chain is effectively the critical path, since it either all works or it doesn't. Garden-runC carries the highest *implementation risk* within the chain due to C code and architecture-specific seccomp policies, but every layer requires validation.
 
 ## Foundational Infrastructure WG
 
 The Foundational Infrastructure Working Group owns `bosh-linux-stemcell-builder` and the BOSH agent. Work includes:
 
-- Extend `bosh-linux-stemcell-builder` to produce ARM64 stemcells (Ubuntu Noble initially, Noble/Resolute Raccoon when applicable)
+- Extend `bosh-linux-stemcell-builder` to produce ARM64 stemcells (Ubuntu Noble initially, Resolute Raccoon when applicable)
 - Validate the BOSH agent on ARM64 (POC has demonstrated this works)
+- Validate BOSH DNS on ARM64
 - Set up CI pipelines producing ARM64 stemcell builds (ARM64 CI workers will be provided)
 - Validate the BOSH Director (Ruby) operates correctly on ARM64 (POC validated 134 gems)
 - For mixed AMD64/ARM64 environments: address BOSH Director's current single-architecture compilation VM limitation (e.g., support multiple compilation VM types or architecture-aware compilation routing)
@@ -203,7 +205,7 @@ The App Runtime Deployments Working Group owns `cf-deployment`. Work includes:
 The ARI Working Group owns the classical buildpacks which have higher adoption in CF today. Note: ARM64 support for Cloud Native Buildpacks (CNBs) is already being addressed by the Paketo WG separately (see [ARM64 Paketo Buildpacks](https://www.cloudfoundry.org/blog/arm64-paketo-buildpacks/)). Work for classical buildpacks includes:
 
 - Produce ARM64 variants of classical buildpack dependencies (JDK, Node.js, Python, Ruby, Go, .NET, PHP, nginx runtimes)
-- Produce an ARM64 variant of the `cflinuxfs4` root filesystem
+- Produce an ARM64 variant of the `cflinuxfs5` (Noble) root filesystem
 - Validate classical buildpack `detect` and `compile` phases on ARM64 cells
 - Update classical buildpack CI to produce and test ARM64 artifacts
 
@@ -216,6 +218,14 @@ The ARI Working Group owns the classical buildpacks which have higher adoption i
 ## CI/CD Infrastructure
 
 ARM64 build workers will be provided for Concourse pipelines to enable native ARM64 compilation and testing. The initial CI infrastructure will be contributed alongside the implementation work. Long-term CI/CD ownership should be discussed as part of this RFC.
+
+## Proposed New Areas
+
+Following the TOC discussion, this RFC proposes creating two new areas to lower the entry barrier and establish clear ownership:
+
+1. **ARM64 Stemcell Area** — responsible for ARM64 stemcell builds, BOSH agent ARM64 validation, and ARM64 CI worker infrastructure. This area has a clear boundary: produce and maintain ARM64 stemcells that pass BATS.
+
+2. **ARM64 CF Stack & Buildpacks Area** — responsible for `cflinuxfs5` ARM64 root filesystem and classical buildpack ARM64 dependency compilation. This area has a clear boundary: produce ARM64 variants of the stack and buildpack dependencies that enable `cf push` on ARM64 cells.
 
 # Open Questions
 
--- a/toc/rfc/rfc-draft-arm64-support.md
+++ b/toc/rfc/rfc-draft-arm64-support.md
@@ -1,0 +1,248 @@
+# Meta
+
+- **Name:** ARM64 (aarch64) Architecture Support for Cloud Foundry
+- **Start Date:** 2026-06-24
+- **Author(s):** Sachin Vighe
+- **Status:** Draft
+- **Related RFCs:** [rfc-0026-noble-os](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0026-noble-os.md)
+- **Affected Component(s):** bosh-linux-stemcell-builder, Diego, Garden-runC, Gorouter, Loggregator, Cloud Controller, UAA, cf-deployment, buildpacks
+
+# Summary
+
+This RFC proposes adding ARM64 (aarch64) architecture support to the Cloud Foundry ecosystem, enabling CF deployments on AWS Graviton and other ARM64 processors. This work is designed to offer potential cost-to-performance savings for CF operators while maintaining full compatibility with existing x86_64 deployments. A proof-of-concept has demonstrated that all major CF components compile and execute on ARM64 with zero source code changes. This RFC proposes a phased approach to deliver ARM64 stemcells, validate platform components, and produce ARM64-compatible buildpacks and stacks.
+
+# Problem
+
+Cloud Foundry currently supports only the x86_64 architecture. ARM64 processors are the fastest-growing architecture in cloud infrastructure, with all major cloud providers offering ARM64 instances. ARM64 instances offer 20–40% better price/performance compared to equivalent x86_64 instances, along with up to 60% better energy efficiency (sustainability benefits). CF operators cannot leverage these savings today because:
+
+1. No ARM64 BOSH stemcells are published
+2. CF component BOSH releases do not produce ARM64 artifacts
+3. The `cflinuxfs4` root filesystem and buildpacks only ship x86_64 binaries
+4. No ARM64 testing or CI/CD infrastructure exists
+
+### Previous Work
+
+In 2018, the Cloud Foundry community [demonstrated CF running on a Raspberry Pi](https://www.cloudfoundry.org/blog/baking-clouds-experiment-raspberry-pi-bosh-cloud-foundry/), identifying seccomp policy issues and syscall differences between ARM and x86_64 as key challenges. Since then, the ARM64 ecosystem has matured significantly — Go has had excellent ARM64 support since Go 1.5 (2015), the OCI runc project fully supports ARM64 with proper seccomp, and all major language runtimes ship ARM64 builds.
+
+### Existing ARM64 Support
+
+Some CF ecosystem tools already have ARM64 builds:
+- **BOSH CLI** — already publishes ARM64 binaries
+- **CF CLI** — already publishes ARM64 binaries
+- **ARM64 JDKs** — production-ready ARM64 JDKs are available
+
+# Proposal
+
+Enable ARM64 as a supported architecture through a phased approach, delivering the work across multiple working groups.
+
+## Goals
+
+1. **Enable ARM64 deployments** of Cloud Foundry across cloud providers
+2. **Maintain x86_64 compatibility** — no breaking changes to existing deployments
+3. **Support multi-architecture builds** in CI/CD pipelines
+4. **Upstream all changes** to CF Foundation repositories (no provider-specific forks)
+5. **Provide production-ready** ARM64 support (not experimental/preview)
+
+## Success Criteria
+
+- BOSH Acceptance Tests (BATS) pass on ARM64 stemcells
+- CF Acceptance Tests (CATs) pass on ARM64 deployment
+- Applications can be deployed via standard buildpacks on ARM64 cells
+- Performance ≥ 100% of x86_64 baseline for equivalent instance classes
+- Zero impact on existing x86_64 deployments
+
+## Non-Goals
+
+1. **Not breaking x86_64 support** — existing deployments continue unchanged
+2. **Not requiring CF operators to migrate** — ARM64 is optional, operators opt in
+3. **Not supporting 32-bit ARM** — only 64-bit ARM (aarch64)
+4. **Not supporting mixed-architecture Diego cells initially** — cells within a deployment are homogeneous per availability zone
+5. **Not creating provider-specific forks** — all work contributed upstream
+
+## Phases
+
+### Phase 1: ARM64 Stemcell
+
+Publish an ARM64 BOSH stemcell based on the current CF Linux operating system (Ubuntu Jammy, and subsequently Noble/Resolute). The stemcell MUST include:
+
+- ARM64 kernel and userspace
+- BOSH agent compiled for ARM64
+- EFI boot support for cloud environments
+- Identical package set to the x86_64 stemcell
+
+### Phase 2: Platform Component Validation
+
+All BOSH releases in `cf-deployment` MUST produce ARM64 binaries alongside x86_64. Go-based components SHOULD cross-compile via CI with `GOARCH=arm64`. Components with C dependencies (e.g., Garden-runC, runC) MUST be validated with native ARM64 compilation.
+
+### Phase 3: ARM64 Stack and Buildpacks
+
+An ARM64 variant of the `cflinuxfs4` root filesystem and ARM64-compiled buildpack dependencies MUST be produced to enable `cf push` application staging on ARM64 cells.
+
+### Phase 4: Multi-Architecture Deployment Support
+
+`cf-deployment` SHOULD provide ops-files or native support for ARM64 deployments, allowing operators to deploy CF on ARM64 infrastructure.
+
+# Proof of Concept Results
+
+A proof-of-concept was conducted to validate ARM64 feasibility before proposing this RFC. The POC compiled and validated 30 ARM64 binaries from 7 major CF releases on real ARM64 hardware (4 vCPU, 16 GB RAM).
+
+## Components Validated
+
+| Component Group | Binaries | Result |
+|---|---|---|
+| **Diego** | rep, auctioneer, bbs, locket | ✅ Cross-compiles and executes |
+| **Garden-runC** | gdn, grootfs, dadoo, dontpanic, thresholder, socket2me, execas, maximus, newuidmap, newgidmap | ✅ Cross-compiles and executes |
+| **runC** | OCI container runtime | ✅ Builds natively with seccomp support |
+| **Gorouter** | gorouter, routing-api, cf-tcp-router, route-registrar | ✅ Cross-compiles and executes |
+| **Loggregator-agent** | loggregator-agent, forwarder-agent, syslog-agent, syslog-binding-cache, prom-scraper, udp-forwarder | ✅ Cross-compiles and executes |
+| **Cloud Controller** | cc-uploader, tps-watcher (Go); 229 Ruby gems | ✅ Go cross-compiles; Ruby gems build natively |
+| **UAA** | cloudfoundry-identity-server WAR | ✅ Architecture-independent (Java) |
+| **BOSH** | bosh-agent, bosh-cli, bosh-agent-pipe; 134 Ruby gems | ✅ Go cross-compiles; Ruby gems build natively |
+
+## Key Findings
+
+1. **Zero source code changes required.** All Go components cross-compile cleanly with `GOOS=linux GOARCH=arm64 CGO_ENABLED=0 go build`. No patches or architecture-specific workarounds were needed.
+
+2. **runC builds natively with full seccomp support.** The container runtime — the highest-risk component due to C code and architecture-specific syscall tables — compiles and functions correctly on ARM64. This resolves the key concern identified in the 2018 Raspberry Pi experiment.
+
+3. **Ruby native extensions work.** All C-extension gems required by Cloud Controller (229 gems including nokogiri, puma, mysql2, oj, pg) and BOSH Director (134 gems) compile successfully on `aarch64-linux`.
+
+4. **ARM64 stemcell boots successfully.** A prototype ARM64 stemcell (Ubuntu Jammy + BOSH agent, 784 MB) was built using `debootstrap --arch=arm64` with EFI boot support and validated on ARM64 hardware.
+
+## POC Test Details
+
+The integration test suite validates binary correctness on real ARM64 hardware rather than running existing component-level integration tests (e.g., Diego's inigo tests or Gorouter's integration suite). The tests verify:
+
+- All 30 compiled binaries are valid ARM64 ELF executables (`file` command confirms `ELF 64-bit LSB executable, ARM aarch64`)
+- Binaries execute correctly and respond to CLI flags (e.g., `--help`, `--version`)
+- Ruby native extensions load correctly on `aarch64-linux` (require/load without errors)
+- runC creates and runs containers on ARM64 with seccomp filtering
+- A multi-container deployment simulation (router + apps + Diego components + BOSH agent + Garden) functions end-to-end
+
+Component-specific integration test suites (e.g., CATs, BATS, Diego BBS tests) running on ARM64 would be part of the validation work in the proposed workstreams below.
+
+## Performance
+
+The POC validated that ARM64 delivers comparable or better performance than x86_64 for CF workloads, including Go compilation, application startup, and memory usage. Detailed benchmarking on production-representative infrastructure should be part of later validation phases within the workstreams.
+
+# Workstreams
+
+The implementation requires work across multiple CF working groups. The dependency chain is:
+
+```
+Stemcells (bosh-linux-stemcell-builder)
+    ↓
+BOSH Agent (bosh-agent)
+    ↓
+BOSH Director (bosh)
+    ↓
+Diego (diego-release) + Garden (garden-runc-release) ← critical path
+    ↓
+Cloud Controller (capi-release) + Routing + Logging
+    ↓
+cflinuxfs4 + Buildpacks
+    ↓
+Applications
+```
+
+## Foundational Infrastructure WG
+
+The Foundational Infrastructure Working Group owns `bosh-linux-stemcell-builder` and the BOSH agent. Work includes:
+
+- Extend `bosh-linux-stemcell-builder` to produce ARM64 stemcells (Ubuntu Jammy initially, Noble/Resolute when applicable)
+- Validate the BOSH agent on ARM64 (POC has demonstrated this works)
+- Set up CI pipelines producing ARM64 stemcell builds (ARM64 CI workers will be provided)
+- Validate the BOSH Director (Ruby) operates correctly on ARM64 (POC validated 134 gems)
+- Run BOSH Acceptance Tests (BATS) on ARM64 stemcells
+- Publish 0.x beta ARM64 stemcells on bosh.io for community validation
+- After successful community validation, publish 1.x GA ARM64 stemcells
+
+## App Runtime Platform WG
+
+The App Runtime Platform Working Group owns Diego, Garden-runC, Gorouter, Loggregator, and other runtime components. Work includes:
+
+- Add ARM64 cross-compilation targets to CI pipelines for all Go-based components
+- Validate Garden-runC (including runC with seccomp) on ARM64 Diego cells — **this is the critical path**
+- Run component-level integration tests on ARM64 (Diego inigo tests, Gorouter integration tests)
+- Validate the Loggregator pipeline on ARM64
+- Validate container networking (Silk CNI, cf-networking) on ARM64
+- Validate NATS and BPM on ARM64
+
+## App Runtime Interfaces WG
+
+The App Runtime Interfaces Working Group owns Cloud Controller (CAPI) and UAA. Work includes:
+
+- Validate Cloud Controller Ruby application runs on ARM64 stemcells (POC validated gem compilation)
+- Validate UAA Java WAR deploys and runs on ARM64 JVM (Eclipse Temurin, SAP Machine, or equivalent ARM64 JDK)
+- Ensure the Go helper binaries (cc-uploader, tps-watcher) are built for ARM64 in CI
+- Run Cloud Controller acceptance tests (CATS) on ARM64
+
+## App Runtime Deployments WG
+
+The App Runtime Deployments Working Group owns `cf-deployment`. Work includes:
+
+- Set up cf-deployment validation against the ARM64 stemcell (similar to existing Jammy/Noble validation)
+- Distribute issues discovered during validation to the responsible component Working Groups
+- Provide an ops-file enabling operators to deploy CF on ARM64
+- Validate the full CF deployment lifecycle (deploy, upgrade, scale) on ARM64
+
+## Paketo WG
+
+The Paketo Working Group owns buildpacks. Work includes:
+
+- Produce ARM64 variants of all CF buildpack dependencies (JDK, Node.js, Python, Ruby, Go, .NET, PHP, nginx, R runtimes)
+- Produce an ARM64 variant of the `cflinuxfs4` root filesystem
+- Validate buildpack `detect` and `compile` phases on ARM64 cells
+- Update buildpack CI to produce and test ARM64 artifacts
+
+## All CF Release Authors
+
+- Add `GOARCH=arm64` cross-compilation to release CI pipelines
+- Validate that BOSH release packaging scripts work on ARM64 stemcells
+- Address any architecture-specific assumptions in job scripts or compilation scripts
+
+## CI/CD Infrastructure
+
+ARM64 build workers will be provided for Concourse pipelines to enable native ARM64 compilation and testing. The initial CI infrastructure will be contributed alongside the implementation work. Long-term CI/CD ownership should be discussed as part of this RFC.
+
+# Open Questions
+
+1. **Stemcell naming convention:** How should ARM64 stemcells be identified? Proposed: `ubuntu-jammy-arm64` alongside existing `ubuntu-jammy` (implicitly x86_64). Open to community input on naming and metadata approach.
+
+2. **CI/CD long-term ownership:** ARM64 CI workers will be provided initially to unblock development. Should long-term ownership transfer to the CF Foundation, or remain as a community-contributed resource?
+
+3. **Stemcell release cadence:** Should ARM64 stemcells follow the same release cadence as x86_64 from day one, or start with a separate beta cadence until stability is proven?
+
+4. **Buildpack priority order:** Which buildpacks should be prioritized for ARM64 support? Proposed: Java → Go → Node.js → Python → Ruby → others. Open to community input based on operator demand.
+
+# Affected Repositories
+
+The full scope covers 60+ repositories across the CF ecosystem. The critical path (P0) repositories are:
+
+**BOSH Infrastructure (5):**
+- `bosh` — Core orchestration (Go, Ruby)
+- `bosh-agent` — VM agent (Go)
+- `bosh-linux-stemcell-builder` — Stemcell creation (Shell, Ruby)
+- Cloud Provider Interfaces (CPIs) for target platforms (e.g., `bosh-aws-cpi-release`, `bosh-google-cpi-release`, `bosh-vsphere-cpi-release`) and corresponding light stemcell builders
+
+**Container Runtime (4):**
+- `diego-release` — Container orchestration (Go)
+- `garden-runc-release` — Container runtime (Go + C) **[highest risk]**
+- `grootfs` — Container filesystem (Go)
+- `guardian` — Garden backend (Go)
+
+**Core Services (5):**
+- `capi-release` — Cloud Controller API (Ruby)
+- `uaa-release` — Authentication (Java)
+- `routing-release` — HTTP routing (Go)
+- `loggregator-release` — Logging (Go)
+- `nats-release` — Messaging (Go)
+
+**Application Layer (2):**
+- `cflinuxfs4-release` — Root filesystem (Shell)
+- `java-buildpack` — Java applications (priority buildpack)
+
+**Testing (1):**
+- `cf-acceptance-tests` — Full platform validation (Go)
+
+Remaining repositories (P1/P2) are primarily Go-based and are expected to cross-compile with minimal effort. A detailed repository inventory will be shared alongside this RFC.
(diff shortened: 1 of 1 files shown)


toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,248 @@
...
+2. **Not requiring CF operators to migrate** — ARM64 is optional, operators opt in
+3. **Not supporting 32-bit ARM** — only 64-bit ARM (aarch64)
+4. **Not supporting mixed-architecture Diego cells initially** — cells within a deployment are homogeneous per availability zone
+5. **Not creating provider-specific forks** — all work contributed upstream
+
+## Phases
+
+### Phase 1: ARM64 Stemcell
+
+Publish an ARM64 BOSH stemcell based on the current CF Linux operating system (Ubuntu Jammy, and subsequently Noble/Resolute). The stemcell MUST include:
```
beyhan: According to the [Ubuntu release plan](https://ubuntu.com/about/release-cycle) Jammy EOL is scheduled for May 2027. Giving that and the fact that we introduced many stemcell changes from Jammy to Noble (like switching from cgroup v1 to cgroup v2) I would suggest to take Jammy out of scope and focus on Noble and Resolute Raccoon.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Paketo WG
+
+The Paketo Working Group owns buildpacks. Work includes:
+
+- Produce ARM64 variants of all CF buildpack dependencies (JDK, Node.js, Python, Ruby, Go, .NET, PHP, nginx, R runtimes)
+- Produce an ARM64 variant of the `cflinuxfs4` root filesystem
+- Validate buildpack `detect` and `compile` phases on ARM64 cells
+- Update buildpack CI to produce and test ARM64 artifacts
+
```
beyhan: The Paketo WG is responsible for the [Cloud Native Buildpacks ](https://buildpacks.io/) which are known as the v3 buildpacks or the current generation of buildpacks. They are [supported by CF](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0028-cnb-lifecycle.md) and CNBs already worked on introducing ARM64 support. See https://buildpacks.io/docs/for-app-developers/how-to/special-cases/build-for-arm/ and https://www.cloudfoundry.org/blog/arm64-paketo-buildpacks/.

The [classical buildpacks](https://github.com/cloudfoundry/community/blob/main/toc/working-groups/app-runtime-interfaces.md?plain=1#L74-L166) have higher adoption in the CF world at the moment and they are owned by the [ARI WG](https://github.com/cloudfoundry/community/blob/main/toc/working-groups/app-runtime-interfaces.md). My impression is that the classical buildpacks are meant here.

toc/rfc/rfc-draft-arm64-support.md
vighes: Agreed. Updated to focus on Noble and Resolute Raccoon, removing Jammy given its May 2027 EOL and the cgroup v2 changes already introduced.

toc/rfc/rfc-draft-arm64-support.md
vighes: Good catch. Removed Paketo WG as a separate workstream since CNB ARM64 support is already in progress. Focused the RFC on classical buildpacks under ARI WG, with a note acknowledging the existing Paketo effort.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,248 @@
...
+- Applications can be deployed via standard buildpacks on ARM64 cells
+- Performance ≥ 100% of x86_64 baseline for equivalent instance classes
+- Zero impact on existing x86_64 deployments
+
+## Non-Goals
+
+1. **Not breaking x86_64 support** — existing deployments continue unchanged
+2. **Not requiring CF operators to migrate** — ARM64 is optional, operators opt in
+3. **Not supporting 32-bit ARM** — only 64-bit ARM (aarch64)
+4. **Not supporting mixed-architecture Diego cells initially** — cells within a deployment are homogeneous per availability zone
```
rkoster: why? wouldn't targetting just diego cells to run on ARM be a much smaller scope and more in line with what we have for windows?

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,248 @@
...
+Publish an ARM64 BOSH stemcell based on the current CF Linux operating system (Ubuntu Noble, and subsequently Resolute Raccoon as it becomes available). The stemcell MUST include:
+
+- ARM64 kernel and userspace
+- BOSH agent compiled for ARM64
+- EFI boot support for cloud environments
+- Identical package set to the x86_64 stemcell
+
+### Phase 2: Platform Component Validation
+
+All BOSH releases in `cf-deployment` MUST produce ARM64 binaries alongside x86_64. Go-based components SHOULD cross-compile via CI with `GOARCH=arm64`. Components with C dependencies (e.g., Garden-runC, runC) MUST be validated with native ARM64 compilation.
```
rkoster: how will components which currently ship binaries as blobs be handled? For example the [jdk shipped by the uaa-release](https://github.com/cloudfoundry/uaa-release/blob/develop/packages/uaa/packaging#L4-L12).

rkoster (changes requested):
Given the impact on the ecosystem I think it would only make sense to also include apple silicon support as in scope. Because this would bring a lot of value to the CF Core developers who have MacBooks and now way of easily validating component changes locally.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,248 @@
...
+
+## CI/CD Infrastructure
+
+ARM64 build workers will be provided for Concourse pipelines to enable native ARM64 compilation and testing. The initial CI infrastructure will be contributed alongside the implementation work. Long-term CI/CD ownership should be discussed as part of this RFC.
+
+# Open Questions
+
+1. **Stemcell naming convention:** How should ARM64 stemcells be identified? Proposed: `ubuntu-jammy-arm64` alongside existing `ubuntu-jammy` (implicitly x86_64). Open to community input on naming and metadata approach.
+
+2. **CI/CD long-term ownership:** ARM64 CI workers will be provided initially to unblock development. Should long-term ownership transfer to the CF Foundation, or remain as a community-contributed resource?
```
rkoster, line 235: Ideally eventually we end up with CF Foundation ownership and get some AWS credits to help with offsetting the increase in CI spend. 

Also related, these workers are these concourse workers running on ARM?

vighes:
> Given the impact on the ecosystem I think it would only make sense to also include apple silicon support as in scope. Because this would bring a lot of value to the CF Core developers who have MacBooks and now way of easily validating component changes locally.

Our POC cross-compiled all Go components from Apple Silicon (M4 Pro Mac), However full local integration (runC with seccomp, container creation) still requires Linux ARM64 either via Docker/Lima on Mac or a remote ARM64 instance. The Linux ARM64 work directly benefits Apple Silicon developers for compilation and unit testing, though not for full local container deployment. I'll add a note acknowledging this developer workflow benefit.

toc/rfc/rfc-draft-arm64-support.md
vighes: Fair point, the Windows cell model is a good precedent for architecture-aware placement. I initially marked mixed cells as out of scope to reduce first-implementation complexity but you are right that targeting ARM64 Diego cells alongside an x86_64 control plane is a smaller, more achievable scope and mirrors the existing Windows approach. I will revise this to acknowledge mixed-architecture cells as a valid deployment model rather than a non-goal.

toc/rfc/rfc-draft-arm64-support.md
vighes: For components shipping binary blobs (like the JDK in uaa-release), the approach would be to add ARM64 variants of those blobs alongside the existing x86_64 ones. BOSH packaging scripts would select the correct blob based on the stemcell architecture. For the JDK specifically, ARM64 builds are readily available. I will add a note in Phase 2 clarifying how binary blob components are handled including architecture-conditional blob selection in packaging scripts.

toc/rfc/rfc-draft-arm64-support.md
vighes: Yes, these would be Concourse workers running on ARM64. For Go cross-compilation tasks, existing x86_64 workers can produce ARM64 binaries via GOARCH=arm64. But for native compilation (runC with seccomp, Ruby gem native extensions, stemcell building) and integration testing, ARM64 workers are required.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,252 @@
...
+- EFI boot support for cloud environments
+- Identical package set to the x86_64 stemcell
+
+### Phase 2: Platform Component Validation
+
+All BOSH releases in `cf-deployment` MUST produce ARM64 binaries alongside x86_64. Go-based components SHOULD cross-compile via CI with `GOARCH=arm64`. Components with C dependencies (e.g., Garden-runC, runC) MUST be validated with native ARM64 compilation. Components shipping pre-compiled binary blobs (e.g., JDK in uaa-release) MUST include ARM64 variants of those blobs, with packaging scripts selecting the correct variant based on stemcell architecture (`uname -m`).
+
+### Phase 3: ARM64 Stack and Buildpacks
+
+An ARM64 variant of the `cflinuxfs4` root filesystem and ARM64-compiled buildpack dependencies MUST be produced to enable `cf push` application staging on ARM64 cells.
```
stephanme: Similar to the stemcell: I would skip cflinuxfs4 (Jammy) and focus on cflinuxfs5 (Noble).

cflinuxfs5 + buildpacks are meanwhile integrated into cf-deployment (opt-in by an experimental ops file): https://github.com/cloudfoundry/cf-deployment/issues/1323

toc/rfc/rfc-draft-arm64-support.md
beyhan: @taylorsilva to my knowledge Concourse is supporting ARM64 and having concourse workers running on ARM64 won't be an issue. Is this correct?

norman-abramovitz:
I totally agree with @rkoster on its value for CF developers.   It would be much easier to have a local CF where I can generate test CF environments rather than using Docker to create local test environments.  It would also help promote Korifi as a development environment, where I am currently trying to get Stratos to run in a Korifi environment.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,252 @@
...
+- Validate Cloud Controller Ruby application runs on ARM64 stemcells (POC validated gem compilation)
+- Validate UAA Java WAR deploys and runs on ARM64 JVM (Eclipse Temurin, SAP Machine, or equivalent ARM64 JDK)
+- Ensure the Go helper binaries (cc-uploader, tps-watcher) are built for ARM64 in CI
+- Run Cloud Controller acceptance tests (CATS) on ARM64
+
+## App Runtime Deployments WG
+
+The App Runtime Deployments Working Group owns `cf-deployment`. Work includes:
+
+- Set up cf-deployment validation against the ARM64 stemcell (similar to existing Jammy/Noble validation)
```
jochenehret, line 199: Setting up another validation pipeline for ARM64 is not a problem. But we should clarify the cost increase in advance.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,252 @@
...
+- Distribute issues discovered during validation to the responsible component Working Groups
+- Provide an ops-file enabling operators to deploy CF on ARM64
```
jochenehret, line 201: For cf-deployment, we would need to switch to an ARM64 stemcell. The ARM64 VM types will be managed in the BOSH cloud-config. Not sure if more changes are required. Do you still have the ops files from the PoC?

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -31,6 +31,10 @@ Some CF ecosystem tools already have ARM64 builds:
 - **CF CLI** — already publishes ARM64 binaries
 - **ARM64 JDKs** — production-ready ARM64 JDKs are available
 
+### Developer Workflow Benefit
+
+ARM64 Linux support also benefits CF core developers on Apple Silicon MacBooks (M1/M2/M3/M4). All Go-based CF components can be cross-compiled locally using `GOOS=linux GOARCH=arm64`, enabling faster development and unit test iteration without needing remote infrastructure.
```
rkoster: I was more thinking along the lines of: https://github.com/cloudfoundry/bosh-deployment/issues/497. Which is close to working but running into rosetta issues. So the aim would be to have apple silicon running ARM linux docker VM (arm linux kernel) then use the bosh docker cpi with a ARM stemcell. This should just work, but it would be great to have it be part of the testing scope. This probably would also directly benefit / speed up the review process of all the other parts of this PR, because it would give many a way to verify the proposed changes locally.

toc/rfc/rfc-draft-arm64-support.md
oliver-heinrich: +1: As Stephan mentioned we should only support ARM for Noble/cflinuxfs5.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,252 @@
...
+### Previous Work
+
+In 2018, the Cloud Foundry community [demonstrated CF running on a Raspberry Pi](https://www.cloudfoundry.org/blog/baking-clouds-experiment-raspberry-pi-bosh-cloud-foundry/), identifying seccomp policy issues and syscall differences between ARM and x86_64 as key challenges. Since then, the ARM64 ecosystem has matured significantly — Go has had excellent ARM64 support since Go 1.5 (2015), the OCI runc project fully supports ARM64 with proper seccomp, and all major language runtimes ship ARM64 builds.
+
+### Existing ARM64 Support
+
+Some CF ecosystem tools already have ARM64 builds:
+- **BOSH CLI** — already publishes ARM64 binaries
+- **CF CLI** — already publishes ARM64 binaries
+- **ARM64 JDKs** — production-ready ARM64 JDKs are available
```
dmikusa, line 32: Paketo Cloud Native Buildpacks are also ARM64 compatible.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,252 @@
...
+- BOSH agent compiled for ARM64
+- EFI boot support for cloud environments
+- Identical package set to the x86_64 stemcell
+
+### Phase 2: Platform Component Validation
+
+All BOSH releases in `cf-deployment` MUST produce ARM64 binaries alongside x86_64. Go-based components SHOULD cross-compile via CI with `GOARCH=arm64`. Components with C dependencies (e.g., Garden-runC, runC) MUST be validated with native ARM64 compilation. Components shipping pre-compiled binary blobs (e.g., JDK in uaa-release) MUST include ARM64 variants of those blobs, with packaging scripts selecting the correct variant based on stemcell architecture (`uname -m`).
+
+### Phase 3: ARM64 Stack and Buildpacks
+
```
dmikusa, line 83: When we did this for Paketo Buildpacks, most of the work and challenges were around automation and release processes. How do you store/access the binaries since there are two copies of everything ow? How do you monitor for updates since that process is now arch specific? How do you release all this throughout pipelines? Lot of assumptions in pipelines that have to be sorted out.

On the CF side, size of Buildpacks that bundle binaries with the buildpack is going to be an issue. Assuming that’s still being done, haven’t looked in a bit. Expect Buildpacks to be twice as large or need to be spilt up some way so they’re arch specific.

toc/rfc/rfc-draft-arm64-support.md
vighes: Agreed, updated in latest commit to target cflinuxfs5 (Noble) only.

toc/rfc/rfc-draft-arm64-support.md
vighes: Agreed, We can provide more details on expected compute hours once we scope the pipeline configurations.

toc/rfc/rfc-draft-arm64-support.md
vighes: We didn't modify the actual cf-deployment repo during the POC but we did create ops files as part of the prep work. The minimal change for cf-deployment is a stemcell architecture switch:

```
# operations/use-arm64-stemcell.yml
- type: replace
  path: /stemcells/alias=default/architecture?
  value: arm64
```
We believe the stemcell switch + cloud-config is sufficient for cf-deployment itself, but full validation would reveal if any additional changes. I also have a sample Graviton cloud-config ready. Happy to contribute these as a follow-up PR to cf-deployment once the RFC is accepted.

toc/rfc/rfc-draft-arm64-support.md
vighes: That's a great use case. I will add local BOSH-lite on Apple Silicon as a validation target in the testing scope. Once the ARM64 stemcell is available in Docker format, this should be straightforward to verify on M-series Macs.

toc/rfc/rfc-draft-arm64-support.md
vighes: Thank you, I will add Paketo Cloud Native Buildpacks to the "Existing ARM64 Support" section alongside BOSH CLI and CF CLI.

toc/rfc/rfc-draft-arm64-support.md
vighes: A few thoughts:
Storage/access: Architecture-specific blobs stored separately, selected at staging time based on cell architecture. Similar to how Paketo handles it today.
Monitoring for updates: The update-checking process would need to be arch-aware checking for both x86_64 and ARM64 releases of each dependency.
Buildpack size: Agree this is an issue if bundles contain both architectures. Architecture-specific buildpack variants (separate downloads per arch) seems cleaner than doubling the size of a single bundle.
Pipeline assumptions: This is likely the largest effort area for classical buildpacks. Would appreciate any lessons learned from the Paketo ARM64 work that we could apply.
I'll add a note in the RFC acknowledging that buildpack release automation is a significant workstream that benefits from Paketo's experience.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,259 @@
...
+```
+
+## Foundational Infrastructure WG
+
+The Foundational Infrastructure Working Group owns `bosh-linux-stemcell-builder` and the BOSH agent. Work includes:
+
+- Extend `bosh-linux-stemcell-builder` to produce ARM64 stemcells (Ubuntu Noble initially, Noble/Resolute Raccoon when applicable)
+- Validate the BOSH agent on ARM64 (POC has demonstrated this works)
+- Set up CI pipelines producing ARM64 stemcell builds (ARM64 CI workers will be provided)
+- Validate the BOSH Director (Ruby) operates correctly on ARM64 (POC validated 134 gems)
```
ystros, line 169: Do we imagine users operating environments that use both AMD64 and ARM64 stemcells? If so, there will likely be work necessary in the Director itself. e.g. currently, the cloud config can only define a single compilation VM type, which will only work for compiling against one architecture.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,259 @@
...
+- Validate that BOSH release packaging scripts work on ARM64 stemcells
+- Address any architecture-specific assumptions in job scripts or compilation scripts
+
+## CI/CD Infrastructure
+
+ARM64 build workers will be provided for Concourse pipelines to enable native ARM64 compilation and testing. The initial CI infrastructure will be contributed alongside the implementation work. Long-term CI/CD ownership should be discussed as part of this RFC.
+
+# Open Questions
+
+1. **Stemcell naming convention:** How should ARM64 stemcells be identified? Proposed: `ubuntu-jammy-arm64` alongside existing `ubuntu-jammy` (implicitly x86_64). Open to community input on naming and metadata approach.
```
ystros: One option would be to follow the convention set by FIPS (which the pipelines build but do not publish to `bosh.io`) and embed the ARM64 into the name field (e.g `bosh-aws-xen-hvm-ubuntu-noble-arm64` instead of just `bosh-aws-xen-hvm-ubuntu-noble`) while keeping the `os` field just `ubuntu-<line name>`. It would depend on whether we want users to opt-in simply by their choice of uploaded stemcells OR if we want them to be very explicit about wanting ARM64 in the manifest itself.

Regardless of naming approach, it would make sense to me to add the architecture as an additional metadata field as well.

toc/rfc/rfc-draft-arm64-support.md
wayneeseguin: at 
https://github.com/concourse/concourse/releases/tag/v8.2.4 
I see https://github.com/concourse/concourse/releases/download/v8.2.4/concourse-8.2.4-linux-arm64.tgz 
under releases section so looks like the answer is yes.

wayneeseguin:
> > Given the impact on the ecosystem I think it would only make sense to also include apple silicon support as in scope. Because this would bring a lot of value to the CF Core developers who have MacBooks and now way of easily validating component changes locally.
> 
> Our POC cross-compiled all Go components from Apple Silicon (M4 Pro Mac), However full local integration (runC with seccomp, container creation) still requires Linux ARM64 either via Docker/Lima on Mac or a remote ARM64 instance. The Linux ARM64 work directly benefits Apple Silicon developers for compilation and unit testing, though not for full local container deployment. I'll add a note acknowledging this developer workflow benefit.

This comes to mind as a new possibility for running locally on Mac: https://github.com/apple/container

toc/rfc/rfc-draft-arm64-support.md
vighes: I believe for environments using both AMD64 and ARM64 stemcells, the BOSH Director would need to support architecture-aware compilation either multiple compilation VM types or the ability to route compilation to the correct architecture. I will add it as an open question / consideration for the mixed-architecture deployment model. For the initial implementation (all-ARM64 deployment), this isn't a blocker since compilation VMs would all be ARM64.

toc/rfc/rfc-draft-arm64-support.md
vighes: Thanks for the FIPS naming precedent, that's a useful reference. Embedding architecture in the name while keeping the os field as ubuntu-noble makes sense. Adding architecture as an explicit metadata field is also a good idea regardless of naming approach it makes the stemcell self-describing. I will update the open question to reference the FIPS approach as a proposed option.

toc/rfc/rfc-draft-arm64-support.md
vighes: Thanks for confirming Concourse ARM64 support.

toc/rfc/rfc-draft-arm64-support.md
aramprice: Using `variant` for the architecture will run into problems if we ever need to support FIPS on ARM64.

toc/rfc/rfc-draft-arm64-support.md
beyhan: This has been addressed with https://github.com/cloudfoundry/community/pull/1530/changes/a939f6c6873acbc0a3d756ae0f766ff958bd6339

toc/rfc/rfc-draft-arm64-support.md
beyhan: This has been addressed with https://github.com/cloudfoundry/community/pull/1530/changes/bf6fcf5d46106469860f369b956126248d46e836

toc/rfc/rfc-draft-arm64-support.md
beyhan: This has been addressed with https://github.com/cloudfoundry/community/pull/1530/changes/a939f6c6873acbc0a3d756ae0f766ff958bd6339

toc/rfc/rfc-draft-arm64-support.md
beyhan: This was addressed with https://github.com/cloudfoundry/community/pull/1530/changes/8da0208113099eae209bdc1df067b19f4626b3a0. Additionally, we touched on this during the FI WG meeting yesterday see following option which could work but it has a limitation mentioned below. 

If you upload the different arch stemcells under different names e.g.:
  - ubuntu-jammy/1 (AMD64)
  - ubuntu-jammy-arm64/1 (ARM64)

 And instance groups reference them accordingly, then BOSH will compile on the correct architecture. It creates separate compilation VMs: one booted from the AMD64 stemcell
image, one from the ARM64 image.

There is a limitation with this approach. The single compilation config vm_type is applied to both compilation VMs. This means:

  - You cannot specify instance_type: m7g.large (ARM-native on AWS) for ARM64 compilation and instance_type: m7i.large for AMD64 compilation.
  - If the CPI's instance type is architecture-specific (as it is on AWS, GCP, Azure), you would be creating an ARM64 stemcell VM on an AMD64 instance type (or vice versa), which the cloud provider would likely reject or which would run incorrectly.
 
This is the concrete gap the workaround does not solve. The compilation config would need to support either:
  - A map of stemcell → vm_type, or
  - Architecture-aware compilation blocks

  to actually work correctly in practice on most real cloud environments. Adding this here just FYI.

toc/rfc/rfc-draft-arm64-support.md
beyhan: Additionally, we discussed during the FI WG meeting yesterday that this could be clarify during the implementation and if required RFC could be updated.

beyhan:
I went through the comments and resolved the ones that appeared to be clarified. To summarize, there are two "major" questions that still need clarification before we can proceed with the RFC:

1. How are we going to cover the increased CI/CD costs?
2. What should the long-term CI/CD ownership model be?

I'll raise these questions during the TOC meeting to work out how we can resolve them.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,260 @@
...
+- Validate the Loggregator pipeline on ARM64
+- Validate container networking (Silk CNI, cf-networking) on ARM64
+- Validate NATS and BPM on ARM64
+
+## App Runtime Interfaces WG
+
+The App Runtime Interfaces Working Group owns Cloud Controller (CAPI) and UAA. Work includes:
+
+- Validate Cloud Controller Ruby application runs on ARM64 stemcells (POC validated gem compilation)
+- Validate UAA Java WAR deploys and runs on ARM64 JVM (Eclipse Temurin, SAP Machine, or equivalent ARM64 JDK)
```
stephanme: UAA is owned by Foundational Infrastructure WG.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,260 @@
...
+- Distribute issues discovered during validation to the responsible component Working Groups
+- Provide an ops-file enabling operators to deploy CF on ARM64
+- Validate the full CF deployment lifecycle (deploy, upgrade, scale) on ARM64
+
+## App Runtime Interfaces WG (Classical Buildpacks)
+
+The ARI Working Group owns the classical buildpacks which have higher adoption in CF today. Note: ARM64 support for Cloud Native Buildpacks (CNBs) is already being addressed by the Paketo WG separately (see [ARM64 Paketo Buildpacks](https://www.cloudfoundry.org/blog/arm64-paketo-buildpacks/)). Work for classical buildpacks includes:
+
+- Produce ARM64 variants of classical buildpack dependencies (JDK, Node.js, Python, Ruby, Go, .NET, PHP, nginx runtimes)
+- Produce an ARM64 variant of the `cflinuxfs4` root filesystem
```
stephanme: cflinuxfs5 (not 4).

beyhan:
> I went through the comments and resolved the ones that appeared to be clarified. To summarize, there are two "major" questions that still need clarification before we can proceed with the RFC:
> 
> 1. How are we going to cover the increased CI/CD costs?
> 2. What should the long-term CI/CD ownership model be?
> 
> I'll raise these questions during the TOC meeting to work out how we can resolve them.

We discussed the two points from above and current decision is:
1. Ask the WG leads to provide rough cost estimation.
2. Suggest creation of two new areas one for the new ARM64 stemcell and one for the CF stack and buildpacks. By this we can lower entry barrier for the contribution and have clear responsibilities for the beginning.

toc/rfc/rfc-draft-arm64-support.md
vighes: Thanks Stephan, Corrected — moved UAA validation to the Foundational Infrastructure WG section. Updated in latest commit.

toc/rfc/rfc-draft-arm64-support.md
vighes: Fixed — updated to cflinuxfs5 in the buildpacks section. Thanks for catching the remaining references.

vighes:
> > I went through the comments and resolved the ones that appeared to be clarified. To summarize, there are two "major" questions that still need clarification before we can proceed with the RFC:
> > 
> > 1. How are we going to cover the increased CI/CD costs?
> > 2. What should the long-term CI/CD ownership model be?
> > 
> > I'll raise these questions during the TOC meeting to work out how we can resolve them.
> 
> We discussed the two points from above and current decision is:
> 
>     1. Ask the WG leads to provide rough cost estimation.
> 
>     2. Suggest creation of two new areas one for the new ARM64 stemcell and one for the CF stack and buildpacks. By this we can lower entry barrier for the contribution and have clear responsibilities for the beginning.

Thanks for the update. The two-area approach (ARM64 stemcell area + CF stack/buildpacks area) makes sense for clear ownership boundaries.

beyhan:
> > > I went through the comments and resolved the ones that appeared to be clarified. To summarize, there are two "major" questions that still need clarification before we can proceed with the RFC:
> > > 
> > > 1. How are we going to cover the increased CI/CD costs?
> > > 2. What should the long-term CI/CD ownership model be?
> > > 
> > > I'll raise these questions during the TOC meeting to work out how we can resolve them.
> > 
> > 
> > We discussed the two points from above and current decision is:
> > ```
> > 1. Ask the WG leads to provide rough cost estimation.
> > 
> > 2. Suggest creation of two new areas one for the new ARM64 stemcell and one for the CF stack and buildpacks. By this we can lower entry barrier for the contribution and have clear responsibilities for the beginning.
> > ```
> 
> Thanks for the update. The two-area approach (ARM64 stemcell area + CF stack/buildpacks area) makes sense for clear ownership boundaries.



I asked about cost estimation in the [TOC Slack channel](https://cloudfoundry.slack.com/archives/C026XJGNC2U/p1784095513968399). @vighes it would be great if you could incorporate the second point about the new WG areas into the RFC.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,260 @@
+# Meta
+
+- **Name:** ARM64 (aarch64) Architecture Support for Cloud Foundry
+- **Start Date:** 2026-06-24
+- **Author(s):** Sachin Vighe
+- **Status:** Draft
+- **Related RFCs:** [rfc-0026-noble-os](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0026-noble-os.md)
+- **Affected Component(s):** bosh-linux-stemcell-builder, Diego, Garden-runC, Gorouter, Loggregator, Cloud Controller, UAA, cf-deployment, buildpacks
```
wayneeseguin: Can't put my finger on it at the moment but I feel there are some more affected components. bosh-dns? others?

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,260 @@
...
+
+Some CF ecosystem tools already have ARM64 builds:
+- **BOSH CLI** — already publishes ARM64 binaries
+- **CF CLI** — already publishes ARM64 binaries
+- **ARM64 JDKs** — production-ready ARM64 JDKs are available
+- **Paketo Cloud Native Buildpacks** — already ARM64 compatible (see [ARM64 Paketo Buildpacks](https://www.cloudfoundry.org/blog/arm64-paketo-buildpacks/))
+
+### Developer Workflow Benefit
+
+ARM64 Linux support also benefits CF core developers on Apple Silicon MacBooks (M1/M2/M3/M4). All Go-based CF components can be cross-compiled locally using `GOOS=linux GOARCH=arm64`, enabling faster development and unit test iteration without needing remote infrastructure. Additionally, an ARM64 stemcell would enable running BOSH with the Docker CPI natively on Apple Silicon (see [bosh-deployment#497](https://github.com/cloudfoundry/bosh-deployment/issues/497)), giving developers a fully local CF environment without Rosetta emulation issues. Validating this local BOSH-lite workflow on Apple Silicon is included in the testing scope.
```
wayneeseguin, line 37: This would be a fantastic outcome indeed. We would be able to get a 'cf local' up and running again most likely, especially with apple releasing their containers project.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,260 @@
...
+3. **Support multi-architecture builds** in CI/CD pipelines
+4. **Upstream all changes** to CF Foundation repositories (no provider-specific forks)
+5. **Provide production-ready** ARM64 support (not experimental/preview)
+
+## Success Criteria
+
+- BOSH Acceptance Tests (BATS) pass on ARM64 stemcells
+- CF Acceptance Tests (CATs) pass on ARM64 deployment
+- Applications can be deployed via standard buildpacks on ARM64 cells
+- Performance ≥ 100% of x86_64 baseline for equivalent instance classes
```
wayneeseguin: Is this always guaranteed? 🤔

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,260 @@
...
+- Zero impact on existing x86_64 deployments
+
+## Non-Goals
+
+1. **Not breaking x86_64 support** — existing deployments continue unchanged
```
wayneeseguin: A bit confusing, wouldn't it be better to state this as a direct goal vs non-goal, e.g. it is a goal to have no impact on existing?

wayneeseguin: I mean maybe it's fine, I find the logic of this section a bit confusing on first read.

wayneeseguin: But I still feel that this section should be rewritten to directly state goals of not breaking things with opt-in support etc...

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,260 @@
...
+The implementation requires work across multiple CF working groups. The dependency chain is:
+
+```
+Stemcells (bosh-linux-stemcell-builder)
+    ↓
+BOSH Agent (bosh-agent)
+    ↓
+BOSH Director (bosh)
+    ↓
+Diego (diego-release) + Garden (garden-runc-release) ← critical path
```
wayneeseguin: Isn't the whole tree really the 'critical path' ? it either all works or it doesn't.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Foundational Infrastructure WG
+
+The Foundational Infrastructure Working Group owns `bosh-linux-stemcell-builder` and the BOSH agent. Work includes:
+
+- Extend `bosh-linux-stemcell-builder` to produce ARM64 stemcells (Ubuntu Noble initially, Noble/Resolute Raccoon when applicable)
+- Validate the BOSH agent on ARM64 (POC has demonstrated this works)
+- Set up CI pipelines producing ARM64 stemcell builds (ARM64 CI workers will be provided)
+- Validate the BOSH Director (Ruby) operates correctly on ARM64 (POC validated 134 gems)
+- For mixed AMD64/ARM64 environments: address BOSH Director's current single-architecture compilation VM limitation (e.g., support multiple compilation VM types or architecture-aware compilation routing)
```
wayneeseguin, line 170: This is key, for example being able to run Diego Cells in isolation segments on say split intel vs arm would be quite valuable.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Success Criteria
+
+- BOSH Acceptance Tests (BATS) pass on ARM64 stemcells
+- CF Acceptance Tests (CATs) pass on ARM64 deployment
+- Applications can be deployed via standard buildpacks on ARM64 cells
+- Performance ≥ 100% of x86_64 baseline for equivalent instance classes
+- Zero impact on existing x86_64 deployments
+
+## Non-Goals
```
wayneeseguin, line 59: Perhaps should declare Windows ARM64 support is a non-goal for this effort?

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,260 @@
...
+- `nats-release` — Messaging (Go)
+
+**Application Layer (2):**
+- `cflinuxfs5-release` — Root filesystem (Shell)
+- `java-buildpack` — Java applications (priority buildpack)
+
+**Testing (1):**
+- `cf-acceptance-tests` — Full platform validation (Go)
+
+Remaining repositories (P1/P2) are primarily Go-based and are expected to cross-compile with minimal effort. A detailed repository inventory will be shared alongside this RFC.
```
wayneeseguin, line 271: This might address my initial comments.

wayneeseguin (changes requested):
Added comments, suggest adding the Windows ARM64 as a non-goal explicitly at this time (could always be added later).

vighes:
> > > > I went through the comments and resolved the ones that appeared to be clarified. To summarize, there are two "major" questions that still need clarification before we can proceed with the RFC:
> > > > 
> > > > 1. How are we going to cover the increased CI/CD costs?
> > > > 2. What should the long-term CI/CD ownership model be?
> > > > 
> > > > I'll raise these questions during the TOC meeting to work out how we can resolve them.
> > > 
> > > 
> > > We discussed the two points from above and current decision is:
> > > ```
> > > 1. Ask the WG leads to provide rough cost estimation.
> > > 
> > > 2. Suggest creation of two new areas one for the new ARM64 stemcell and one for the CF stack and buildpacks. By this we can lower entry barrier for the contribution and have clear responsibilities for the beginning.
> > > ```
> > 
> > 
> > Thanks for the update. The two-area approach (ARM64 stemcell area + CF stack/buildpacks area) makes sense for clear ownership boundaries.
> 
> I asked about cost estimation in the [TOC Slack channel](https://cloudfoundry.slack.com/archives/C026XJGNC2U/p1784095513968399). @vighes it would be great if you could incorporate the second point about the new WG areas into the RFC.

Hi @beyhan Done — see the "Proposed New Areas" section which proposes two new areas with clear boundaries and scope.

vighes:
> Added comments, suggest adding the Windows ARM64 as a non-goal explicitly at this time (could always be added later).

Added as explicit non-goal #1. There's no Windows Server ARM64 offering from major cloud providers today, so this is firmly out of scope.

toc/rfc/rfc-draft-arm64-support.md
vighes: Added bosh-agent, bosh-dns, BPM and NATS to the Affected Components. Also added their validation to the relevant workstreams. The Affected Repositories section at the bottom provides the full breakdown.

toc/rfc/rfc-draft-arm64-support.md
vighes: Fair point, not guaranteed across all workload types. Reworded to "comparable" with benchmarking deferred to Phase 2. POC data and public Graviton benchmarks suggest parity or better for Go/Java/Ruby workloads.

toc/rfc/rfc-draft-arm64-support.md
vighes: Restructured, Non-Goals section now focuses purely on what's out of scope. Thanks for flagging.

toc/rfc/rfc-draft-arm64-support.md
vighes: that's right, the whole chain is the critical path. Removed the misleading annotation from the diagram and clarified while the entire chain must work, Garden-runC carries the highest implementation risk specifically due to C code and seccomp.

toc/rfc/rfc-draft-arm64-support.md
vighes: Agreed! Once ARM64 stemcells exist, BOSH-lite on Docker CPI becomes native on Apple Silicon no more Rosetta issues. Validating this workflow is included in the testing scope.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,270 @@
...
+## App Runtime Platform WG
+
+The App Runtime Platform Working Group owns Diego, Garden-runC, Gorouter, Loggregator, and other runtime components. Work includes:
+
+- Add ARM64 cross-compilation targets to CI pipelines for all Go-based components
+- Validate Garden-runC (including runC with seccomp) on ARM64 Diego cells — **this is the critical path**
+- Run component-level integration tests on ARM64 (Diego inigo tests, Gorouter integration tests)
+- Validate the Loggregator pipeline on ARM64
+- Validate container networking (Silk CNI, cf-networking) on ARM64
+- Validate NATS and BPM on ARM64
```
ameowlia: BPM is not part of the ARP working group

vighes: Good catch — moved BPM to the Foundational Infrastructure WG section. Thanks!

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,271 @@
...
+
+# Affected Repositories
+
+The full scope covers 60+ repositories across the CF ecosystem. The critical path (P0) repositories are:
+
+**BOSH Infrastructure (5):**
+- `bosh` — Core orchestration (Go, Ruby)
+- `bosh-agent` — VM agent (Go)
+- `bosh-linux-stemcell-builder` — Stemcell creation (Shell, Ruby)
+- Cloud Provider Interfaces (CPIs) for target platforms (e.g., `bosh-aws-cpi-release`, `bosh-google-cpi-release`, `bosh-vsphere-cpi-release`) and corresponding light stemcell builders
```
s4heid, line 249:
```suggestion
- Cloud Provider Interfaces (CPIs) for target platforms (e.g., `bosh-aws-cpi-release`, `bosh-google-cpi-release`, `bosh-vsphere-cpi-release`, `bosh-azure-cpi-release`) and corresponding light stemcell builders
```

Supporting ARM64 on Azure requires more than producing an ARM64 stemcell; the Azure CPI must create an Azure Compute Gallery image with the correct architecture and Gen2 metadata, and it must be paired with an ARM64-compatible Azure VM size. I created the following issues for tracking the Azure-CPI prerequistes:

- https://github.com/cloudfoundry/bosh-azure-cpi-release/issues/748
- https://github.com/cloudfoundry/bosh-azure-cpi-release/issues/749
- https://github.com/cloudfoundry/bosh-azure-cpi-release/issues/750

beyhan:
> > > > I went through the comments and resolved the ones that appeared to be clarified. To summarize, there are two "major" questions that still need clarification before we can proceed with the RFC:
> > > > 
> > > > 1. How are we going to cover the increased CI/CD costs?
> > > > 2. What should the long-term CI/CD ownership model be?
> > > > 
> > > > I'll raise these questions during the TOC meeting to work out how we can resolve them.
> > > 
> > > 
> > > We discussed the two points from above and current decision is:
> > > ```
> > > 1. Ask the WG leads to provide rough cost estimation.
> > > 
> > > 2. Suggest creation of two new areas one for the new ARM64 stemcell and one for the CF stack and buildpacks. By this we can lower entry barrier for the contribution and have clear responsibilities for the beginning.
> > > ```
> > 
> > 
> > Thanks for the update. The two-area approach (ARM64 stemcell area + CF stack/buildpacks area) makes sense for clear ownership boundaries.
> 
> I asked about cost estimation in the [TOC Slack channel](https://cloudfoundry.slack.com/archives/C026XJGNC2U/p1784095513968399). @vighes it would be great if you could incorporate the second point about the new WG areas into the RFC.

**Regarding: Rough estimation of the expected costs**
We have now a rough estimation about the costs based on the current GCP setup we have. This could be different on AWS. The estimations are:
* best case ~4K $ per month
* worst case ~7K $ per month

More details are available in this [Slack thread](https://cloudfoundry.slack.com/archives/C026XJGNC2U/p1784095513968399).

**Regarding: Short term & long-term maintenance & CI/CD ownership**
We have discussed how to organize the initial contribution in its own WG areas. We need to discuss options for the short term & long-term maintenance & CI/CD ownership.

We discussed during the TOC meeting yesterday that it makes sense to have a call to discuss. @christopherclark will check with @vighes for having a call.

toc/rfc/rfc-draft-arm64-support.md
wayneeseguin: Can we also do openstack stemcell, we use it quite a bit?

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,271 @@
...
+## Phases
+
+### Phase 1: ARM64 Stemcell
+
+Publish an ARM64 BOSH stemcell based on the current CF Linux operating system (Ubuntu Noble, and subsequently Resolute Raccoon as it becomes available). The stemcell MUST include:
+
+- ARM64 kernel and userspace
+- BOSH agent compiled for ARM64
+- EFI boot support for cloud environments
+- Identical package set to the x86_64 stemcell
```
aramprice, line 76: Nit: these will likely not be 100% identical as there can be slight variances in package naming, and presence based on platofrm.

toc/rfc/rfc-draft-arm64-support.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Phase 2: Platform Component Validation
+
+All BOSH releases in `cf-deployment` MUST produce ARM64 binaries alongside x86_64. Go-based components SHOULD cross-compile via CI with `GOARCH=arm64`. Components with C dependencies (e.g., Garden-runC, runC) MUST be validated with native ARM64 compilation. Components shipping pre-compiled binary blobs (e.g., JDK in uaa-release) MUST include ARM64 variants of those blobs, with packaging scripts selecting the correct variant based on stemcell architecture (`uname -m`).
```
aramprice, line 80: Explore: there may be value in investigating whether a release can be assembled for *one* or *multiple* (both) supported architectures. There is a significant cost to shuffling around an ARM64 JVM which might not be used in a pure x86_64 environment (vice versa for a pure ARM64 env).

Similarly it would be valuable if compiled releases could be created for one or many architectures, rather than aways including every architecture (both in this case).

beyhan:
> Short term & long-term maintenance & CI/CD ownership

As I will be going on vacation and have been driving this topic, I reached out in the meantime to maintainers from the different working groups to gather their perspectives on the second question: “Short-term and long-term maintenance and CI/CD ownership.”

Overall, the feedback has been largely positive regarding the adoption and future maintenance of this feature. Based on these discussions, I believe we have a good basis to consider the maintenance question addressed.

One recommendation I received was to engage with the community early during the PR implementation phase. This would allow us to provide early feedback and guidance throughout the implementation, while also helping to avoid potential pitfalls that could create challenges later.

If we can identify a suitable approach to manage the validation costs, I believe this RFC should be ready to move into the final comment period.

vighes:
Following a sync call with @christopherclark  and @beyhan, we have identified a short-term resolution for CI/CD cost to unblock ARM64 development.

beyhan:
We discussed this during the TOC meeting on 15th of September and decided to start the FCP.

norman-abramovitz:
Stratos has already supported ARM64 since we use it for testing.  The cf-mgmt will just be adding another delivered artifact.

PR #1457, opened 2026-03-23
Fix yq commit message parsing in contributions script

a-hassanin:
f631f9255859feb0eac3f08ce3b504c6ee1f5a17 (Group by PR)
Refactors toc/working-groups/contributions-for-user.sh to replace flat commit listing with find_commits_grouped_by_pr_for_repo.

“Code contributions” now:
- finds merged PRs authored by the user (search/issues ... author:<user> is:merged)
- prints each PR as a top-level bullet
- fetches commits per PR and prints commit bullets under the PR only when the PR has more than one commit
- falls back to recent PullRequestEvent merged/closed events only when search returns no merged PRs.
- Removes sort -u from the code-contributions section so PR/commit grouping is preserved.

76bad69c7358df12c5ceb4e28dd8282828d476b7 (Fix yq commit message parsing in contributions script)
- Fixes the yq expression in commit formatting from (.commit.message | split("\n")[0]) to ((.commit.message | split("\n"))[0]), avoiding the !!seq ... cannot be added to a !!str runtime error.
- Fixes gist filename generation to use ${WORKING_GROUP} instead of ${wg} in uri_safe_file, preventing incorrect/empty filename prefixes.

PR #1532, opened 2026-06-25
Improve contributions-for-user.sh: surface authored PRs, group commits, skip empty areas

Ivaylogi98:
## What

Reworks `toc/working-groups/contributions-for-user.sh` so the generated gists actually reflect a user's contribution footprint. The previous version omitted authored PRs entirely, generated empty gists for unrelated areas, and missed several categories of contribution relevant to RFC-0006 approver requirements.

## Why

Running the script as a candidate for approver returned gists that contained only review/comment activity — no authored PRs, no code contributions on others' PRs, and a flood of empty gists for areas the user has never touched. This made the output unusable for assembling evidence of substantial contributions.

### New sections in the gist
- **PRs Authored** — PRs where the user is the author, with their open/closed/merged state. Previously absent entirely.
- **Commits on others' PRs** — commits authored by the user that landed in a PR authored by someone else (i.e. pushed onto a collaborator's branch). Previously invisible.
- **Code contributions (merged PRs)** — replaces the flat commit listing. Merged PRs are now top-level bullets; commits are indented underneath only when a PR has more than one. Picks up the refactor proposed in #1457.

### Correctness fixes
- Gist filename was using the undefined `${wg}` instead of `${WORKING_GROUP}`, producing empty-prefixed filenames.
- `yq` commit-message parsing used `.commit.message | split("\n")[0]` which raised `!!seq cannot be added to !!str`. Fixed to `((.commit.message | split("\n"))[0])`.
- GitHub search index silently omits some PRs (`bosh-agent#393` is one example). Added a targeted events-based fallback that does single-PR lookups for any PR numbers visible in the user's events but missing from search results.
- The author/committer commit search used `repos/{repo}/commits?author=` which filters by email rather than login, missing most commits. Replaced with `search/commits?q=author:${user}+repo:${repo}`.
- Normalize `USERNAME` via `gh api users/${USERNAME}` so the GitHub search API (case-sensitive on `author:` / `reviewed-by:` qualifiers) matches reliably.

### Performance / robustness
- Skip gist creation for areas with zero contributions — eliminates a dozen empty gists per run.
- `--paginate` added to `search/commits` so >100 commits are not silently dropped.
- Events fallback filters to `payload.action == "opened"` so each PR triggers at most one API lookup, not one per recorded event.
- All deduplication moved from O(n²) array loops to O(1) associative-array sets (`local -A seen`).
- `emit_pr_with_commits` promoted from a nested closure to a top-level function with an explicit `repo` parameter, removing the implicit closure-by-name on the outer function's `${repo}`.
- Dropped the now-deprecated `cloak-preview` and `groot-preview` Accept headers — both endpoints have been stable for years.
- Shebang switched to `#!/usr/bin/env bash` with an explicit bash 4+ version check, since the script now relies on associative arrays. macOS users on the system bash 3.2 get a clear "install bash via Homebrew" message instead of cryptic `local -A` errors mid-run.

## Testing

Ran against `Foundational Infrastructure` for an approver-candidate user. Areas with no contributions are now skipped; areas with contributions produce gists with all five sections populated, including PRs the GitHub search index had been silently dropping.

Authored with Claude Code

### DIFF
--- a/toc/working-groups/contributions-for-user.sh
+++ b/toc/working-groups/contributions-for-user.sh
@@ -1,6 +1,14 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e -o pipefail
+
+# Requires bash 4+ for associative arrays. macOS ships bash 3.2 at /bin/bash —
+# install a modern bash (e.g. via Homebrew) so that `env bash` resolves to 4+.
+if (( BASH_VERSINFO[0] < 4 )); then
+  echo "This script requires bash 4 or newer (found ${BASH_VERSION})." >&2
+  echo "On macOS: brew install bash, then re-run." >&2
+  exit 1
+fi
 
 repo_root=$(dirname "$(cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd)")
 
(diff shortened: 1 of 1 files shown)


beyhan:
@Ivaylogi98 are you planing to work on this or should we close it? I would like to keep the number of open pr as small as possible.

Ivaylogi98:
@beyhan we can close it for now, I'll reopen when it's in a better state.

PR #756, opened 2024-01-18
Remove individual repo access via github action

ameowlia:
You can see a dry run of the github action [here](https://github.com/cloudfoundry/community/actions/runs/7576159899/job/20634338969). It prints out everyone who will have their access removed.

fixes #528

Issue #528, opened 2023-02-07
Cleanup the access permissions of github users not managed by the CFF ogranization automation 


beyhan:
The current version of the automation which we have in place maintains the access permissions for the CFF organization repositories on github teams level. It doesn't manage the access permissions of individual github users to github repositories. That is why there are still users with access permissions to the CFF organization repositories which shouldn't be there.  This issue is about finding a way for automatic and periodical cleanup of such user permissions. 

This is a result of the discussions during the [TOC meeting on 2023-02-07](https://docs.google.com/document/d/1ng8rAcnrCbTm5RVNAKZ_ptqXiwcBIPgC8rbYp6A8lUo/edit#heading=h.3xng24xf236).

Gerg:
For the [App Runtime Interfaces working group](https://github.com/cloudfoundry/community/blob/main/toc/working-groups/app-runtime-interfaces.md), the following users are expected to lose access to the respective repositories:

@KevinJCross -  cloudfoundry/app-autoscaler-cli-plugin
@stanimirovn -  cloudfoundry/docs-buildpacks
@hibell -  cloudfoundry/ibm-websphere-liberty-buildpack
@kevin-ortega -  cloudfoundry/ibm-websphere-liberty-buildpack
@sclevine -  cloudfoundry/dotnet-core-buildpack
@arjun024 -  cloudfoundry/buildpacks-ci
@brayanhenao -  cloudfoundry/buildpacks-ci
@GarimaSharma -  cloudfoundry/buildpacks-workstation
@mgibson1121 -  cloudfoundry/buildpacks-workstation
@arjun024 -  cloudfoundry/libbuildpack
@brayanhenao -  cloudfoundry/libbuildpack
@arjun024 -  cloudfoundry/switchblade
@brayanhenao -  cloudfoundry/switchblade
@matt-royal -  cloudfoundry/cloud_controller_ng
@soochow -  cloudfoundry/capi-team-checklists
@weymanf -  cloudfoundry/cc-uploader
@maximilien -  cloudfoundry/cli-i18n
@davidschachner -  cloudfoundry/cli-docs-scripts
@blgm -  cloudfoundry/jsonry
@xtreme-bozhidar-lenchov -  cloudfoundry/notifications-release

[...]

PR #1598, opened 2026-08-03
Remove individual access for concourse org repos

stephanme:
- same procedure as for cloudfoundry org


beyhan (approved):

stephanme:
@taylorsilva: I plan to merge this PR today if there is no veto.

It applies the same individual user access cleanup for the concourse org that is in place for the cloudfoundry org since 2 years. Introduced by #528 and #756.

taylorsilva (approved):

Issue #1602, opened 2026-08-08
Grant cloudfoundry/stratos workflows write access to the ghcr.io/cloudfoundry/stratos package


norman-abramovitz:
The Stratos release workflows build the all-in-one container image successfully but cannot push it: `denied: permission_denied: write_package` for `ghcr.io/cloudfoundry/stratos`. Login succeeds; authorization fails.

The package predates the current workflows (its only tag, `stable`, was pushed over 5 years ago), so it carries its own legacy permissions and the repository was never granted access to it. `GITHUB_TOKEN` with `packages: write` therefore can't push, and no credential in the repo can fix that. Full diagnosis in cloudfoundry/stratos#5763.

Since package access sits outside the org-management automation in this repo, this needs an org admin:

On https://github.com/orgs/cloudfoundry/packages/container/stratos/settings, under **Manage Actions access**, add the `cloudfoundry/stratos` repository with the **Write** role.

No new credential or secret is involved — once granted, the existing workflows publish as-is, and release runs stop reporting failure for an otherwise-successful release.

@stephanme @Gerg — filing here per the repo conventions since this isn't coverable by a PR to the org yaml; happy to adjust if there's a better route.

stephanme:
My proposal is to add a `temp-ari-stratos-admins` team to [orgs/orgs.yml](https://github.com/cloudfoundry/community/blob/main/orgs/orgs.yml) that grants repo admin rights to selected approvers of the area. For an example, search for `temp-ari-buildpacks-admins` in this file.

Alternatives have been discussed in #800. The mentioned option 1 (Manually configured temporary admin teams) is still in use.

beyhan:
We discussed this during the TOC meeting today and still the [comment](https://github.com/cloudfoundry/community/issues/1602#issuecomment-5254809622) made by Stephan is what we can offer at the moment. @norman-abramovitz  please follow that approach.

norman-abramovitz:
Following up now that #1613 is merged. The temp-ari-stratos-admins team is created and I have admin on cloudfoundry/stratos, but it doesn't reach the GHCR package:
opening the package settings for ghcr.io/cloudfoundry/stratos gives "You do not have permissions to administrate this package."

The package was published in November 2020 (a single stable tag, an openSUSE Leap 15.2 base image, 245 downloads total). It predates the current release workflow, so
The repo's GITHUB_TOKEN gets write_package denied on every release. That's cloudfoundry/stratos#5763.
 
[...]

PR #1613, opened 2026-09-01
Add temp-ari-stratos-admins team

norman-abramovitz:
Follows up #1602, which asked for workflows in `cloudfoundry/stratos` to be able to write the `ghcr.io/cloudfoundry/stratos` package.

Per @stephanme's proposal in https://github.com/cloudfoundry/community/issues/1602#issuecomment-5254809622, confirmed by @beyhan after the TOC meeting on 2026-09-01, this adds a temporary admin team for the area rather than granting package permissions directly. It follows the `temp-ari-buildpacks-admins` pattern already in this file.

Maintainers are approvers for the Stratos area.

### Why this is needed

The **Build All-in-One Image** job in the Stratos release workflow has failed with `denied: permission_denied: write_package` on every release since v5.1.0 — five consecutive releases, most recently v5.5.2 today. The release itself completes and its 9 assets publish correctly, so the only symptom is that each run's conclusion reads `failure` and the AIO image never reaches the registry. Repo admin lets the area configure the package's write permissions and clear that.

Tracked on the Stratos side as cloudfoundry/stratos#5763.

### DIFF
--- a/orgs/orgs.yml
+++ b/orgs/orgs.yml
@@ -2837,6 +2837,16 @@ orgs:
         privacy: closed
         repos:
           terraform-provider-cloudfoundry: admin
+      temp-ari-stratos-admins:
+        description: "Temporary team for managing Stratos secrets and packages until we have a permanent plan."
+        maintainers:
+          - norman-abramovitz
+          - krutten
+          - wayneeseguin
+        members: []
+        privacy: closed
+        repos:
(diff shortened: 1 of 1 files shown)


norman-abramovitz:
Does this PR require two reviewers to approve?

stephanme (approved):

PR #1615, opened 2026-09-01
Archive Korifi repositories

danail-branekov:
The CF-on-K8S WG is being archived, see [RFC-0060](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0060-archive-cf-on-k8s-wg.md)

As part of this activity, korifi repositories should be archived.

### DIFF
--- a/orgs/branchprotection.yml
+++ b/orgs/branchprotection.yml
@@ -274,27 +274,7 @@ branch-protection:
         cf-k8s-secrets:
           protect: false
         korifi:
-          protect: true
-          enforce_admins: true
-          allow_force_pushes: false
-          allow_deletions: false
-          include: ["^main$"]
-          required_status_checks:
-            contexts:
-              - "api-tests"
-              - "controllers-tests"
-              - "linter"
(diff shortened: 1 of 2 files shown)


orgs/branchprotection.yml
```diff
@@ -274,27 +274,7 @@ branch-protection:
         cf-k8s-secrets:
           protect: false
         korifi:
```
stephanme, line 276: The special config for the korifi repo should be removed.
Default branch protection from org level will be applied (which should be no protection for archived projects).

stephanme (changes requested):

orgs/branchprotection.yml
danail-branekov: done, branch protection config removed for repos `korifi-ci`, `cf-k8s-secrets`, `korifi`

stephanme (approved):

Gerg (approved):

beyhan (approved):

mkocher (approved):

PR #1624, opened 2026-09-10
[RFC] Provide CF Platform Logs over Otel 

beyhan:
[Preview](https://github.com/cloudfoundry/community/blob/otel-platform-logs/toc/rfc/rfc-draft-platform-logs-with-opentelemetry.md)


toc/rfc/rfc-draft-platform-logs-with-opentelemetry.md
```diff
@@ -0,0 +1,146 @@
...
+- Foundational Infrastructure
+- Application Runtime Platform
+
+### Implementation
+
+Cloud Foundry supports platform log collection on two major types of operating systems: Linux with the `syslog-release` and Windows with the [windows-syslog-release](https://github.com/cloudfoundry/windows-syslog-release). To offer a well-integrated opt-in option, both releases should be analyzed so the OpenTelemetry-based approach can complement rather than disrupt existing setups.
+
+#### Focus
+
+This implementation will focus on collection and processing of the platform component logs (BOSH job logs) stored in `/var/vcap/sys/log` directory. Supporting other paths where logs are stored is nice to have capability.
```
jorbaum:
```suggestion
This implementation will focus on collection and processing of the platform component logs (BOSH job logs) stored in `/var/vcap/sys/log` directory. Supporting logs stored in other paths is a nice-to-have capability.
```

jorbaum (commented):
I found a minor fix for one sentence.

Proof-read again and still gets my 👍

jorbaum:
There are two additional log paths for "platform logs" besides blackbox:

1. BOSH jobs → forwarder-agent → syslog-agent → registered as an aggregate drain → log-cache
2. BOSH jobs → forwarder-agent → loggregator-agent → doppler → RLP (legacy, removed in shared-nothing)

Those are logs produced by platform components, but also visible to users of the CF platform.

For 1) we might need to come up with a way to replace this path in this RFC? Log Cache has no OTLP ingress and the collector has no log-cache exporter.

```
forwarder-agent ─┬─ syslog-agent ──(RFC5424 syslog)──► log-cache-syslog-server ─► Log Cache ─► cf tail / Grafana
                 └─ otel-collector ──(OTLP)──► external backend        ✗ no route into Log Cache
```

For 2) OTLP egress with otel-collector is the modern equivalent. Even though [RFC 0018](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0018-aggregate-metric-egress-with-opentelemetry-collector.md) mainly targeted metrics only. The collector already supports those as it is able to forward everything coming from the forwarder-agent to otel endpoints (opt-in via `EMIT_OTEL_LOGS`).

I guess it depends on how we want to scope "platform logs" for this RFC.

PR #1625, opened 2026-09-10
Remove inactive users in FI WG

beyhan:
According to the rules for inactivity defined in [RFC-0025](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0025-define-criteria-and-removal-process-for-inactive-members.md), the following users should be removed from FI approvers/reviewers/bots:

@coolgang123
@xtreme-nitin-ravindran
@praveenkalluri18
@Jobsby

According to the [revocation policy in the RFC](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0025-define-criteria-and-removal-process-for-inactive-members.md#remove-the-membership-to-the-cloud-foundry-github-organization), users have two weeks to refute this revocation, if they wish by commenting on this pull-request and open a new pull-request to be re-added as approver after this one is merged.

Inactivity was detected by a github action, see https://github.com/cloudfoundry/community/pull/1622


Jobsby:
Hi everyone,

many thanks for organizing and maintaining the list of approvers.

I would like to exercise the two-week window provided for in the revocation policy to object to this revocation. My limited activity over the past few months was due to a seven-month fellowship, which has since concluded. I would like to continue serving as a reviewer/approver and gradually resume my activity.

I am committed to being actively involved again moving forward and would be happy to remain part of the team. However, if the preferred procedure under the policy entails being removed first and then submitting a new pull request for reinstatement, I am happy to follow that path, just let me know what works best.

Thanks again!

